PR #11, opened 2020-07-10
Report memory usage of OSDs

pponnuvel:
Reports the memory usage of Ceph OSDs


plugins/storage/01ceph
```diff
@@ -17,7 +17,8 @@ for svc in ${services[@]}; do
 
     ids="`get_ps| sed -r \"s/.*(${svc}[[:alnum:]\-]*)\s+.+--id\s+([[:digit:]]+)\s+.+/\2/g;t;d\"| tr -s '\n' ','| sort| sed -r -e 's/^\s+/  /g' -e 's/,$//g'`"
     for osd_id in `echo $ids| tr ',' ' '`;do
-        out_str="ceph-osd (id=$osd_id)"
+        osd_RSS=$(awk -v id="--id $osd_id" '/ceph-osd/ && $0 ~ id {print int($8/1024)}' sos_commands/process/ps_alxwww)
```
dosaboy, line 20: Can you make this conditional on ((VERBOSITY_LEVEL>=1)) i.e. -vv

I think this level of detail is better kept for a higher verbosity level

dosaboy: erm sorry that would be equivalent to -v

dosaboy:
lgtm

PR #13, opened 2020-07-15
[plugins/storage] report device types of OSDs (e.g. ssd or hdd)

pponnuvel:


plugins/storage/01ceph
```diff
@@ -26,6 +26,11 @@ for svc in ${services[@]}; do
                 out_str="$out_str (fsid=$osd_fsid) (device=$osd_device)"
             fi
         fi
+
```
dosaboy, line 29: i think this one should be under -vvv so VERBOSITY_LEVEL>=3

PR #14, opened 2020-07-15
Report OSDs' uptimes

pponnuvel:
Report how long the OSDs are running for.

The time reported is in DD:HH:MM:SS format (DD = days; HH = hours; MM = minutes; SS = seconds).

### DIFF
--- a/plugins/storage/01ceph
+++ b/plugins/storage/01ceph
@@ -11,6 +11,8 @@ services=(
 
 echo "ceph:"
 
+sos_time_secs=$(date --date="$(cat sos_commands/date/date)" +%s)
+
 for svc in ${services[@]}; do
     exists="`get_ps| sed -r \"s/.*(${svc}[[:alnum:]\-]*)\s+.+/\1/g;t;d\"| sort -u| sed -r 's/^\s+/  /g'`"
     [ -z "$exists" ] && continue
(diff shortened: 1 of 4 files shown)


pponnuvel:
Put this under `-v` as well (i.e. VERBOSITY_LEVEL >= 1).

plugins/storage/01ceph
```diff
@@ -11,13 +11,27 @@ services=(
 
 echo "ceph:"
 
+sos_time_secs=$(date --date="$(cat sos_commands/date/date)" +%s)
```
dosaboy, line 14: hotsos can be run against a live system or a sosreport and to make that possible, whenever you need to read from a location e.g. under sos_commands, if that location doesn't exist you can then try to run the actual command. The way this is currently done is to have warpper functions in https://github.com/dosaboy/hotsos/blob/master/helpers so e.g. we can replace this with a get_date () that will first check sos path then fallback to real. Also, you need to always preface sos paths with the $DATA_ROOT environment variable.

plugins/storage/01ceph
```diff
@@ -11,26 +11,37 @@ services=(
...
+            # OSD's uptime (time since it was started)
+            osd_start="$(get_ps_axo_flags | awk -v id="--id $osd_id" '/ceph-osd/ && $0 ~ id {print $13, $14, $15, $16, $17; exit}')"
+            if [[ $sos_time_secs -gt 0 && -n "$osd_start" ]]; then
+                osd_start_secs=$(date --date="${osd_start}" +%s)
+                osd_uptime_secs=$((sos_time_secs - osd_start_secs))
+                osd_uptime_str="$(seconds_to_date $osd_uptime_secs)"
+                out_str="$out_str (up=$osd_uptime_str)"
             fi
         fi
+
```
dosaboy, line 38: all looks good but can we keep fsid and device printed first before rss and uptime

plugins/storage/01ceph
```diff
@@ -11,26 +11,37 @@ services=(
...
-                osd_fsid=`tail -n+$offset sos_commands/ceph/ceph-volume_lvm_list| grep -m1 "osd fsid"| sed -r 's/.+\s+([[:alnum:]]+)/\1/g'`
-                osd_device=`tail -n+$offset sos_commands/ceph/ceph-volume_lvm_list| grep -m1 "devices"| sed -r 's/.+\s+([[:alnum:]\/]+)/\1/g'`
-                out_str="$out_str (fsid=$osd_fsid) (device=$osd_device)"
+            # OSD's uptime (time since it was started)
+            osd_start="$(get_ps_axo_flags | awk -v id="--id $osd_id" '/ceph-osd/ && $0 ~ id {print $13, $14, $15, $16, $17; exit}')"
+            if [[ $sos_time_secs -gt 0 && -n "$osd_start" ]]; then
+                osd_start_secs=$(date --date="${osd_start}" +%s)
+                osd_uptime_secs=$((sos_time_secs - osd_start_secs))
+                osd_uptime_str="$(seconds_to_date $osd_uptime_secs)"
+                out_str="$out_str (up=$osd_uptime_str)"
```
dosaboy, line 35: nit: "up" could be repalced with "etime" since thats the ps convention

plugins/storage/01ceph
```diff
@@ -11,26 +11,37 @@ services=(
...
 
+sos_time_secs=$(get_sosreport_time)
+
 for svc in ${services[@]}; do
     exists="`get_ps| sed -r \"s/.*(${svc}[[:alnum:]\-]*)\s+.+/\1/g;t;d\"| sort -u| sed -r 's/^\s+/  /g'`"
     [ -z "$exists" ] && continue
 
     ids="`get_ps| sed -r \"s/.*(${svc}[[:alnum:]\-]*)\s+.+--id\s+([[:digit:]]+)\s+.+/\2/g;t;d\"| tr -s '\n' ','| sort| sed -r -e 's/^\s+/  /g' -e 's/,$//g'`"
     for osd_id in `echo $ids| tr ',' ' '`;do
         out_str="ceph-osd (id=$osd_id)"
```
dosaboy, line 22: also while are here, can you change the above to:

out_str="ceph-osd.$osd_id"

PR #15, opened 2020-07-21
Add a bit of documentation

pponnuvel:


README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This tool is intended to be used to extract information out of
+sos reports that are commonly looked at when debugging problems.
+It has a number of *plugins* that deal with extracting application
+specific information.
+
+It's designed to be able to run on both live systems as well as
+sos reports.
+
+It has multiple *verbosity* levels (`-v` flag) to get more fine-grained
```
dosaboy, line 11: maybe make it clearer that the verbosity level is based on the number of "v" e.g. -v, -vv etc

dosaboy: actually now i see you have given an example of this below so maybe its ok

PR #19, opened 2020-07-24
[plugins/storage] report OSDs' uptime

pponnuvel:
Report the how long the Ceph OSDs are running for.
Useful to infer memory hogging/fragmentation, fd saturation, etc.

The necessary wrap functions for it to be able to run on both
live system as well as existing sos reports are added.

This also addresses the review comments from [previous PR](https://github.com/dosaboy/hotsos/pull/14) (which I closed & created this with cleanups):
 - change `ceph-osd (id=$osd_id)` to `ceph-osd.$osd_id`
 - Put device & fsid before RSS
 - Use `etime` instead of `up` (for osd uptimes)

### DIFF
--- a/helpers
+++ b/helpers
@@ -11,6 +11,18 @@ get_ps ()
 }
 export -f get_ps
 
+get_ps_axo_flags ()
+{
+    # Older sosrepot uses 'wchan' option while newer ones use 'wchan:20' - thus the glob is to cover both
+    local sos_path="${DATA_ROOT}/sos_commands/process/ps_axo_flags_state_uid_pid_ppid_pgid_sid_cls_pri_addr_sz_wchan*_lstart_tty_time_cmd"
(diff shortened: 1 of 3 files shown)


helpers
```diff
@@ -69,3 +81,37 @@ get_udevadm_info_dev ()
...
+    local sos_path="${DATA_ROOT}/sos_commands/date/date"
+    if [ -s "$sos_path" ]; then
+        date --date="$(cat $sos_path)" +%s
+    elif ! [ -d "${DATA_ROOT}sos_commands" ]; then
+        date +%s
+    fi
+}
+export -f get_sosreport_time
+
+get_ceph_volume ()
```
dosaboy, line 108: can we call this get_ceph_volume_lvm_list() just in case we want to add other ceph-volume commands

PR #20, opened 2020-08-05
[plugins/storage] Report OSD's fsid and device via lvm2

pponnuvel:
In some environments where OSDs may be deployed manually
via lvm2 (instead of juju/ceph-volume). So we can get OSD data
from 'lvs' command in such cases.

### DIFF
--- a/helpers
+++ b/helpers
@@ -115,3 +115,15 @@ get_ceph_volume_lvm_list ()
     fi
 }
 export -f get_ceph_volume_lvm_list
+
+get_lvm2_lvs ()
+{
+    local sos_path="${DATA_ROOT}/sos_commands/lvm2/lvs_-a_-o_lv_tags_devices*"
+    if [ -e $sos_path ]; then
+        cat $sos_path
+    elif ! [ -d "${DATA_ROOT}sos_commands" ] && which lvs >/dev/null; then
+        lvs -a -o lv_tags,devices
+    fi
+}
+export -f get_lvm2_lvs
+
--- a/plugins/storage/01ceph
+++ b/plugins/storage/01ceph
@@ -26,6 +26,11 @@ for svc in ${services[@]}; do
             osd_fsid=`get_ceph_volume_lvm_list | tail -n+$offset | grep -m1 "osd fsid" | sed -r 's/.+\s+([[:alnum:]]+)/\1/g'`
             osd_device=`get_ceph_volume_lvm_list | tail -n+$offset | grep -m1 "devices" | sed -r 's/.+\s+([[:alnum:]\/]+)/\1/g'`
             out_str="$out_str (fsid=$osd_fsid) (device=$osd_device)"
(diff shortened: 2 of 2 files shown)


plugins/storage/01ceph
```diff
@@ -26,6 +26,10 @@ for svc in ${services[@]}; do
             osd_fsid=`get_ceph_volume_lvm_list | tail -n+$offset | grep -m1 "osd fsid" | sed -r 's/.+\s+([[:alnum:]]+)/\1/g'`
             osd_device=`get_ceph_volume_lvm_list | tail -n+$offset | grep -m1 "devices" | sed -r 's/.+\s+([[:alnum:]\/]+)/\1/g'`
             out_str="$out_str (fsid=$osd_fsid) (device=$osd_device)"
+        else
+            declare -a arr=($(get_lvm2_lvs | awk -v id="osd_id=$osd_id" \
+                                             '/osd_fsid=/ && $0 ~ id {match($0, /osd_fsid=([^,]+)/,a); split($NF,b,"("); print a[1],b[1]}'))
```
dosaboy, line 31: Can we make the above line a little less hard to read? e.g.

declare -a arr=(get_lvm2_lvs|
                             awk -v id="osd_id=$osd_id"
                                 '/osd_fsid=/ && $0 ~ id {match($0, /osd_fsid=([^,]+)/,a); split($NF,b,"("); print a[1],b[1]}')
                         )

helpers
```diff
@@ -115,3 +115,15 @@ get_ceph_volume_lvm_list ()
...
 }
 export -f get_ceph_volume_lvm_list
+
+get_lvm2_lvs ()
+{
+    local sos_path="${DATA_ROOT}/sos_commands/lvm2/lvs_-a_-o_lv_tags_devices*"
+    if [ -e $sos_path ]; then
+        cat $sos_path
+    elif ! [ -d "${DATA_ROOT}sos_commands" ] && which lvs >/dev/null; then
+        lvs -a -o lv_tags,devices,lv_kernel_read_ahead,lv_read_ahead,stripes,stripesize
```
dosaboy, line 125: is this definitely going to produce equivalent output to the sosreport version?

pponnuvel: This is based on what's used in sos: https://github.com/sosreport/sos/blob/master/sos/report/plugins/lvm2.py#L68
I have previously tested this in sts bastion and that awk line (from 01ceph) retrieved the same info correctly.

pponnuvel: I'll test again in sts bastion, just to be sure.

plugins/storage/01ceph
pponnuvel: done.

helpers
pponnuvel: ok, I probably didn't check sts but few sosreports. Looks like all these options aren't needed - just the 2 are sufficient.

PR #28, opened 2020-09-01
[plugins/storage] report OSD's device type (e.g. ssd or hdd)

pponnuvel:
'ceph osd tree' shows the device 'class' which is part of all "Ceph" sosreports.


dosaboy:
lgtm. im really sorry but because i just merged a patch that i had forgotten to push a while a go now has a conflict so can you rebase it? really sorry...

PR #40, opened 2021-01-29
Fix buddyinfo / ps + ipaddr on k8s

niedbalski:
* Fixed buddyinfo for vms/containers
* Fixed no ps output process/ip addrs for k8s

Signed-off-by: Jorge Niedbalski <jorge.niedbalski@canonical.com>

### DIFF
--- a/plugins/kubernetes/01general
+++ b/plugins/kubernetes/01general
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 import re
 import os
+import sys
 
 import helpers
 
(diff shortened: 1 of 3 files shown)


plugins/kubernetes/01general
```diff
@@ -16,11 +17,17 @@ SERVICES = ["etcdctl",
             "containerd",
             "dockerd"]
 
+ps_output = helpers.get_ps()
+if not ps_output:
+    print("not found k8s processes running, skipping")
+    sys.exit(0)
+
```
dosaboy, line 24: We don't want to exit because it might still be possible to list pods/containers based in logs (see below). Also we want to always print "kubernetes:" so that we know the module ran. The idea i tried to use across plugins is that if there is no data it just prints empty (yaml) list.

niedbalski:
@dosaboy hey ed, I think I've addressed your request, let me know if that fits with what you're requesting, specially the part of printing the empty [] list.

PR #42, opened 2021-02-10
[plugin/ceph] Pythonise few leftovers in ceph plugin

pponnuvel:

### DIFF
--- a/plugins/storage/01ceph
+++ b/plugins/storage/01ceph
@@ -15,31 +15,48 @@ SERVICES = ["ceph-osd",
             "radosgw"]
 
 
-# TODO: finish converting these
 def ceph_pg_imbalance():
     """
-    ## upstream recommends 50-100 OSDs ideally.
-    #for line in helpers.get_ceph_osd_df_tree():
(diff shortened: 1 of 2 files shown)


plugins/storage/01ceph
```diff
@@ -15,31 +15,48 @@ SERVICES = ["ceph-osd",
...
+                pg = int(line.split()[-2])
+                if pg > 0 and (pg < 50 or pg > 200):
+                    bad_pgs += str(pg) + " "
+        except IndexError:
+            pass
+
+    if bad_pgs:
+        print("    pgs: WARNING! Found OSDs with pg counts {}."
+              .format(bad_pgs.strip()) +
+              " Recommended range is 50 - 200.")
```
dosaboy, line 39: Hi Pon, couple of things. Firstly can you please rebase this patch off master to get the large amount of changes landed. I would then like to request to additions/changes; some unit tests for these functions and also that instead of printing messages, you just allow the versions to be displayed in the yaml output e.g. in  ceph_versions_mismatch() just add vers to CEPH_INFO and it will be included in the output.

dosaboy (changes requested):

PR #43, opened 2021-02-19
[juju] Add 04known_bugs

freyes:
This is a new plugin meant to detect known bugs, this first version is
capable of detecting possible occurrences of LP: #1910958

Example output:

```
$ /home/freyes/Projects/hotsos/hotsos.sh --juju ./
hotsos:
  version: development
  repo-info: 50beab7
...
juju:
  machines:
    running:
      - 31 (version=unknown)
  charm-versions:
    - filebeat-33
    ...
    stopped:
      - rabbitmq-server-3
  - bug: 'LP: #1910958'
    description: Unit fails to start complaining there are members in the relation.
    files_with_matches:
      - ./var/log/juju/unit-rabbitmq-server-3.log
      - ./var/log/juju/unit-rabbitmq-server-2.log

INFO: see --help for more display options
```

### DIFF
--- a/plugins/juju/04known_bugs
+++ b/plugins/juju/04known_bugs
@@ -1,0 +1,39 @@
+#!/usr/bin/python3
+from common import searchtools
+from common.known_bugs_utils import add_known_bug
+from juju_common import (
+    JUJU_LOG_PATH
+)
+
+# maximum number of chars to print when a matching text is detected.
+MAX_MATCH_CHARS = 200
+
+
+def detect_known_bugs():
+    """Unit fails to start complaining there are members in the relation."""
+    known_bugs = {
+        1910958: {
+            "description": ("Unit fails to start complaining there are "
+                            "members in the relation."),
+            "pattern": (
+                r'.* manifold worker returned unexpected error: failed to '
+                r'initialize uniter for "[A-Za-z0-9-]+": cannot create '
+                r'relation state tracker: cannot remove persisted state, '
+                r'relation \d+ has members')
+            }
+        }
+
+    s = searchtools.FileSearcher()
+    for bug in known_bugs:
+        s.add_search_term(known_bugs[bug]["pattern"], [],
+                          f"{JUJU_LOG_PATH}/*", tag=bug)
+
+    results = s.search()
+
+    for bug in known_bugs:
+        if results.find_by_tag(bug):
+            add_known_bug(bug, known_bugs.get("description"))
+
+
+if __name__ == "__main__":
+    detect_known_bugs()
--- a/common/helpers.sh
+++ b/common/helpers.sh
@@ -145,7 +145,7 @@ get_ceph_versions ()
         cat $sos_path
     elif ! [ -d "${DATA_ROOT}sos_commands" ] && which ceph >/dev/null; then
         ceph versions
-    fi  
+    fi
 }
 export -f get_ceph_versions
 
--- a/common/known_bugs_utils.py
+++ b/common/known_bugs_utils.py
@@ -47,7 +47,8 @@ def add_known_bug(bug_id, description=None, type=LAUNCHPAD):
 
     current = _get_known_bugs()
     if current and current.get(MASTER_YAML_KNOWN_BUGS_KEY):
-        current[MASTER_YAML_KNOWN_BUGS_KEY].append(entry)
+        if entry not in current.get(MASTER_YAML_KNOWN_BUGS_KEY):
+            current[MASTER_YAML_KNOWN_BUGS_KEY].append(entry)
     else:
         current = {MASTER_YAML_KNOWN_BUGS_KEY: [entry]}
 
--- a/common/searchtools.py
+++ b/common/searchtools.py
@@ -138,9 +138,7 @@ class FileSearcher(object):
 
     def _search_task(self, term_key, fd, path):
         results = []
-        for ln, line in enumerate(fd):
-            # line numbers are not zero-indexed
-            ln += 1
+        for ln, line in enumerate(fd, start=1):
             for s_term in self.paths[term_key]:
                 if type(line) == bytes:
                     line = line.decode("utf-8")
@@ -153,7 +151,12 @@ class FileSearcher(object):
                 ret = s_term["key"].match(line)
                 if ret:
                     r = SearchResult(ln, path, s_term.get("tag"))
+                    # ensure we always add this
+                    r.add(0, ret[0])
                     for i in s_term["indices"]:
+                        if i == 0:
+                            continue
+
                         r.add(i, ret[i])
 
                     results.append(r)
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -3,3 +3,4 @@ mock
 nose
 pylint
 testtools
+rpdb
--- a/tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-2.log
+++ b/tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-2.log
@@ -1,0 +1,2 @@
+2021-02-17 08:18:44 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker returned unexpected error: failed to initialize uniter for "unit-rabbitmq-server-2": cannot create relation state tracker: cannot remove persisted state, relation 236 has members
+2021-02-17 08:20:34 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker returned unexpected error: failed to initialize uniter for "unit-rabbitmq-server-2": cannot create relation state tracker: cannot remove persisted state, relation 236 has members
--- a/tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-3.log
+++ b/tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-3.log
@@ -1,0 +1,2 @@
+2021-02-17 08:18:44 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker returned unexpected error: failed to initialize uniter for "unit-rabbitmq-server-2": cannot create relation state tracker: cannot remove persisted state, relation 236 has members
+2021-02-17 08:20:34 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker returned unexpected error: failed to initialize uniter for "unit-rabbitmq-server-2": cannot create relation state tracker: cannot remove persisted state, relation 236 has members
--- a/tests/unit/test_juju.py
+++ b/tests/unit/test_juju.py
@@ -1,13 +1,19 @@
 import mock
+import os
+import shutil
+import tempfile
 
 from importlib.util import spec_from_loader, module_from_spec
 from importlib.machinery import SourceFileLoader
 
 import utils
 
+from common import known_bugs_utils
+
+
 # need this for non-standard import
 specs = {}
-for plugin in ["01juju", "02charms", "03units"]:
+for plugin in ["01juju", "02charms", "03units", "04known_bugs"]:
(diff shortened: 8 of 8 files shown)


plugins/juju/04known_bugs
```diff
@@ -0,0 +1,59 @@
...
+              }
+    pattern = (
+        'manifold worker returned unexpected error: failed to initialize '
+        'uniter for "[A-Za-z0-9\\-]+": cannot create relation state tracker: '
+        'cannot remove persisted state, relation \\d+ has members'
+    )
+
+    cmd = ["grep", "-r", "-P", "-H", "--include=*.log", "--files-with-match",
+           pattern, JUJU_LOG_PATH]
+    try:
```
niedbalski, line 25: 1 cent on this, why not making this a util?

freyes:
On Fri, 2021-02-19 at 09:28 -0800, Jorge Niedbalski wrote:
> @niedbalski commented on this pull request.
> 
> In plugins/juju/04known_bugs:
> 
> > +
> +def detect_bug1910958():
> +    """Unit fails to start complaining there are members in the
> relation."""
> +    result = {"bug": "LP: #1910958",
> +              "description": ("Unit fails to start complaining there
> are "
> +                              "members in the relation."),
> +              }
> +    pattern = (
> +        'manifold worker returned unexpected error: failed to
> initialize '
> +        'uniter for "[A-Za-z0-9\\-]+": cannot create relation state
> tracker: '
> +        'cannot remove persisted state, relation \\d+ has members'
> +    )
> +
> +    cmd = ["grep", "-r", "-P", "-H", "--include=*.log", "--files-
> with-match",
> +           pattern, JUJU_LOG_PATH]
> +    try:
> 1 cent on this, why not making this a util?

a util as in "put this code in a generic module to be able to reuse
it"?

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,59 @@
...
+
+    return result
+
+
+KNOWN_ISSUES = [detect_bug1910958]
+
+
+if __name__ == "__main__":
+
+    issues_found = []
```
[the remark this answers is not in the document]
dosaboy: e.g. https://pastebin.ubuntu.com/p/Ffws7BWr7D/

common/constants.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+# HOTSOS GLOBALS
+VERBOSITY_LEVEL = int(os.environ.get('VERBOSITY_LEVEL', 0))
+DATA_ROOT = os.environ.get('DATA_ROOT', '/')
+
+JUJU_LOG_PATH = os.path.join(DATA_ROOT, "var/log/juju")
```
dosaboy, line 7: I like this but think it needs to be done for all plugins. I will fix this separately so that you can pull it into your patch.

dosaboy (changes requested):

common/constants.py
dosaboy: See https://github.com/dosaboy/hotsos/commit/da484b58d975331518993935f6e9448569178650

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,59 @@
...
+              "description": ("Unit fails to start complaining there are "
+                              "members in the relation."),
+              }
+    pattern = (
+        'manifold worker returned unexpected error: failed to initialize '
+        'uniter for "[A-Za-z0-9\\-]+": cannot create relation state tracker: '
+        'cannot remove persisted state, relation \\d+ has members'
+    )
+
+    cmd = ["grep", "-r", "-P", "-H", "--include=*.log", "--files-with-match",
```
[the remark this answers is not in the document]
freyes: any attempt to implement `grep` in python would fail miserably[0]. I understand the desire to not be using subprocess around, I'm in that boat too, but I truly believe that hotsos is better served relying on gnu grep

[0] https://lists.freebsd.org/pipermail/freebsd-current/2010-August/019310.html

common/constants.py
freyes: done.

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,59 @@
...
+           pattern, JUJU_LOG_PATH]
+    try:
+        output = subprocess.check_output(cmd,
+                                         universal_newlines=True)
+        log.debug("cmd: %s, exit code: 0", cmd)
```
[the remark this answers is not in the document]
freyes: done.

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,59 @@
...
+KNOWN_ISSUES = [detect_bug1910958]
+
+
+if __name__ == "__main__":
+
+    issues_found = []
+    for func in KNOWN_ISSUES:
+        result = func()
+        if result:
+            issues_found.append(result)
```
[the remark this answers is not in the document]
freyes: done.

plugins/juju/04known_bugs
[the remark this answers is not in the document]
freyes: I moved the code that calls grep to common.grep(). If this really needs to be reimplemented in pure python I can do that, I could do it too.

plugins/juju/04known_bugs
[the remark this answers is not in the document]
freyes: done. This is how it will look like in the revision of the patch:

```
./hotsos.sh  --juju tests/unit/fake_data_root/
...
juju:
  machines: {}
  charm-versions: []
  units:
    stopped:
      - rabbitmq-server-2
      - rabbitmq-server-3
    lxd:
      - ceph-osd-no-fixed-wal-7
      - nova-compute-0
  launchpad:
    - 1910958:
        description: Unit fails to start complaining there are members in the relation.
        match: 'tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-3.log:2021-02-17
          08:18:44 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker
          returned unexpected error: failed to initializ...
  
          tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-3.log:2021-02-17
          08:20:34 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker
          returned unexpected error: failed to initializ...
  
          tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-2.log:2021-02-17
          08:18:44 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker
          returned unexpected error: failed to initializ...
  
          tests/unit/fake_data_root/var/log/juju/unit-rabbitmq-server-2.log:2021-02-17
          08:20:34 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker
          returned unexpected error: failed to initializ...'

INFO: see --help for more display options
```

plugins/juju/04known_bugs
freyes: made this a function that wraps `grep` calls.

plugins/juju/04known_bugs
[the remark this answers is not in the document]
dosaboy: We can use this for now if you want but what we really need is a solution that scales. Imagine a scenario where we want to search a set of files using a set of keys. Running grep over each file for each key is going to take a very long time (in my experience grep regex is really slow). With a little thought I'm sure we could come up with something quite powerful that is flexible enough to search for anything we need and can do it must faster than grep. For example as a starting point I quite like the solutions proposed in https://www.blopig.com/blog/2016/08/processing-large-files-using-python/. That combined with a search classifier would be great.

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,70 @@
...
+import logging
+import subprocess
+
+from common import helpers, constants
+from juju_common import (
+    JUJU_LOG_PATH
+)
+
+
+log = logging.getLogger("juju.04known-bugs")
```
dosaboy, line 11: we dont need this do we?

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,70 @@
...
+        issues_found = {"juju": issues_found}
+        indent = 0
+    else:
+        indent = 2
+
+    return (issues_found, indent)
+
+
+if __name__ == "__main__":
+    (issues_found, indent) = main()
```
dosaboy, line 69: this will still print something if issues_found is empty. also, if we want to be able to report issues from > launchpad we need a top-level key like "known-bugs"

dosaboy (changes requested):

plugins/juju/04known_bugs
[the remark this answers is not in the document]
dosaboy: I've had a go at implementing some search tools, please take a look and let me know what you think - https://github.com/dosaboy/hotsos/pull/50

plugins/juju/04known_bugs
freyes: done. removed the logging

freyes:
New version of the patch, I dropped the use of grep, I believe there are no more outstanding issues to be addressed.

common/helpers.py
```diff
@@ -355,3 +356,23 @@ def get_hostname():
         return open(path, 'r').readlines()
 
     return []
+
+
+def grep(pattern, files):
```
dosaboy, line 361: Can we use common.searchtools? e.g. https://github.com/dosaboy/hotsos/blob/master/plugins/openstack/09agent_errors since it supports (a) parallel searches (b) gzipped files and (c) tagged results to make it much easier to identify positive hits.

freyes: done.

common/helpers.py
```diff
@@ -338,3 +339,23 @@ def get_uname():
         return open(path, 'r').read()
 
     return ""
+
+
+def grep(pattern, files):
+    """Run grep passing *args as arguments
```
dosaboy, line 345: Can we get rid of this since it isn't used and hopefully not needed.

common/constants.py
```diff
@@ -14,3 +14,6 @@
     USE_ALL_LOGS = True
 else:
     USE_ALL_LOGS = False
+
+# maximum number of chars to print when a matching text is detected.
+MAX_MATCH_CHARS = 200
```
dosaboy, line 19: I don't think this should be super global. Maybe better to make it Juju specific for now.

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,75 @@
...
+        'cannot remove persisted state, relation \\d+ has members'
+    )
+
+    try:
+        s = searchtools.FileSearcher()
+        s.add_search_term(pattern, [0], f"{JUJU_LOG_PATH}/*.log")
+
+        results = s.search()
+
+        disp_text = []
```
dosaboy, line 32: You can simplify this by tagging the search term with the bug id it corresponds to e.g. https://github.com/dosaboy/hotsos/blob/62d2fc468cd48d6bdc677e520db8d2a41e2f9d4f/plugins/openstack/08agent_checks#L444. That way you can identify hits by tag and only add the known_bug if you get a hit.

plugins/juju/04known_bugs
```diff
@@ -0,0 +1,75 @@
...
+        return None
+
+    return result
+
+
+KNOWN_ISSUES = [detect_bug1910958]
+
+
+def main():
+    issues_found = {"launchpad": []}
```
dosaboy, line 54: This code can be simplified now since the indentation handling is all done in the plugin_yaml code. See other plugins for an example but basically all you need to do is something like:

if issues_found:
        helpers.HOTSOSYaml.dump(issues_found)

and ignore all the indent checking stuff.

dosaboy (changes requested):
Thanks Felipe, looks better. Few comments but this is definitely getting close!

freyes:
Rebased the patch based on Ed's feedback.

dosaboy (approved):

dosaboy (approved):

PR #46, opened 2021-02-22
[plugin/storage] pythonise two funcs and add unit tests

pponnuvel:


common/helpers.py
```diff
@@ -181,14 +192,18 @@ def get_ceph_osd_tree():
...
-def get_ceph_versions():
+def get_ceph_versions(json_fmt=False):
     if DATA_ROOT == '/':
         output = subprocess.check_output(['ceph', 'versions'])
         return output.decode('UTF-8').splitlines()
 
     path = os.path.join(DATA_ROOT, "sos_commands/ceph/ceph_versions")
     if os.path.exists(path):
-        return open(path, 'r').readlines()
+        if json_fmt:
```
dosaboy, line 202: Since we need this function to return the same whether it is called from file or real i think its best to the json bit in the caller. You'll also need to add simplejson to requirements.txt.

tests/unit/test_storage.py
```diff
@@ -36,6 +36,34 @@ def test_get_service_info(self):
...
+                   'osd.15': 49,
+                   'osd.16': 316,
+                   'osd.17': 370,
+                   'osd.34': 392,
+                   'osd.37': 406,
+                   'osd.56': 209,
+                   'osd.72': 206}}
+        storage_01ceph.get_ceph_pg_imbalance()
+        print(result)
+        print(storage_01ceph.CEPH_INFO)
```
dosaboy, line 64: can you remove the print statements from these tests? (fyi i think you only see their output if they go to stderr anyway)

dosaboy (changes requested):
hey pon looks good, couple of small requests for you.

dosaboy (approved):

PR #52, opened 2021-03-15
specify UTF-8 encoding when reading from paths

### LINKED ISSUE #51: hotsos snap encoding issue
Running hotsos (snap) against a sosreport with cdk installed happens to trigger this when parsing snap names:

```
  File "/snap/hotsos/160/plugins/kubernetes/01general", line 100, in <module>
    get_snap_info()
  File "/snap/hotsos/160/plugins/kubernetes/01general", line 56, in get_snap_info
    snap_list_all = helpers.get_snap_list_all()
  File "/snap/hotsos/160/plugins/kubernetes/common/helpers.py", line 64, in catch_exception_inner2
    return f(*args, **kwargs)
  File "/snap/hotsos/160/plugins/kubernetes/common/helpers.py", line 218, in get_snap_list_all
    return open(path, 'r').readlines()
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1604: ordinal not in range(128)
```

brettmilford:
Fixes #51

### DIFF
--- a/common/helpers.py
+++ b/common/helpers.py
@@ -77,7 +77,7 @@ def get_ip_addr():
 
     path = os.path.join(DATA_ROOT, "sos_commands/networking/ip_-d_address")
     if os.path.exists(path):
-        return open(path, 'r').readlines()
+        return open(path, 'r', encoding='utf8').readlines()
 
     return []
 
@@ -90,7 +90,7 @@ def get_ip_link_show():
 
     path = os.path.join(DATA_ROOT, "sos_commands/networking/ip_-s_-d_link")
     if os.path.exists(path):
-        return open(path, 'r').readlines()
+        return open(path, 'r', encoding='utf8').readlines()
 
     return []
 
@@ -118,7 +118,7 @@ def get_ps():
 
     path = os.path.join(DATA_ROOT, "ps")
     if os.path.exists(path):
-        return open(path, 'r').readlines()
+        return open(path, 'r', encoding='utf8').readlines()
 
     return []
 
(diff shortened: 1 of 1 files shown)


dosaboy (commented):
Hi Brett, are you sure this fixes your issue? The default encoding in py3 is already utf-8. Some time back I hit a similar issue and actually had to fix it with commit fc8de80 which allows the file to be read anyway and just skips unreadable characters, something which is probably not appropriate in all cases.

UPDATE: the original fix was in commit 9014613

brettmilford:
Hey @dosaboy,
It does fix the issue for me; however using `errors="surrogateescape"` would work as well - I'm happy to try this if you think it would be a better approach. I decided on specifying UTF-8 as this is the encoding used for the `subprocess` commands. 

Text encoding used by open() is dependent on the platform and what is returned by locale.getpreferredencoding()
https://docs.python.org/3.7/library/functions.html#open

Which for my locally installed python is UTF-8
```
Python 3.8.6 (default, Jan 27 2021, 15:42:20)
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
'UTF-8'
```
But for some reason is ANSI for my snap confined python.
```
~/prjs/hotsos$ sudo snap run --shell hotsos
root@orpheus:/home/brett/prjs/hotsos# python3
Python 3.6.9 (default, Oct  8 2020, 12:12:24)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import locale
>>> locale.getpreferredencoding()
'ANSI_X3.4-1968'
```

dosaboy:
hi @brettmilford thats a really good point, i think it would be best to fix the locale so that everybody has utf-8 by default. i'll have a look into this.

dosaboy:
@brettmilford default locale is now hopefully fixed by commit bb11584

brettmilford:
Tested with a snap build from master/tip - it now works for me.
Thanks!

PR #54, opened 2021-03-31
Get LP bug description

ivoks:
Rather than just bug numbers, it would be great to also pick up bug title.


dosaboy (commented):
lgtm but for some reason the u/t didn't fail when run by github but it does if you run locally e.g. i get two errors like:

FAIL: test_add_known_bug_first (test_known_bugs_utils.TestKnownBugsUtils)
----------------------------------------------------------------------
  File "/home/user1/git/dosaboy/hotsos/tests/unit/test_known_bugs_utils.py", line 45, in test_add_known_bug_first
    self.assertEquals(ret,
AssertionError: {'known-bugs': [{'https://bugs.launchpad.net/bugs/1': 'Mi[34 chars]e'}]} != {'known-bugs': ['https://pad.lv/1']}
+ {'known-bugs': ['https://pad.lv/1']}
- {'known-bugs': [{'https://bugs.launchpad.net/bugs/1': 'Microsoft has a '
-                                                       'majority market share'}]}

----------------------------------------------------------------------
Ran 48 tests in 2.556s

I think we should add a mock to test_add_known_bug_first to mock out launchpadlib.

PR #57, opened 2021-04-05
Use json output for ceph commands

nicolasbock:
The `plain` output format is fragile when it comes to output changes
such as between Stein and Ussuri. Ussuri adds another column which is
hard to detect in this format.

Instead we should take advantage of the `json_output` data.


tests/unit/test_storage.py
```diff
@@ -86,33 +86,75 @@ def test_get_ceph_versions_mismatch_unavailable(self):
...
+    @mock.patch.object(storage_01ceph.helpers, "get_ceph_osd_df_tree",
+                       lambda: """ID CLASS WEIGHT  REWEIGHT SIZE   USE     DATA    OMAP META  AVAIL   %USE  VAR   PGS TYPE NAME
+-1       0.02939        - 30 GiB 3.0 GiB 8.4 MiB  0 B 3 GiB  27 GiB 10.03 1.00    - root default
+-3       0.00980        - 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00    -      host juju-1157aa-ceph-1
+ 0   hdd 0.00980  1.00000 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00  267        osd.0
+-7       0.00980        - 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00    -      host juju-1157aa-ceph-2
+ 2   hdd 0.00980  1.00000 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00  269        osd.2
+-5       0.00980        - 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00    -      host juju-1157aa-ceph-3
+ 1   hdd 0.00980  1.00000 10 GiB 1.0 GiB 2.8 MiB  0 B 1 GiB 9.0 GiB 10.03 1.00  268        osd.1
+                    TOTAL 30 GiB 3.0 GiB 8.4 MiB  0 B 3 GiB  27 GiB 10.03
```
dosaboy, line 115: I would really prefer to stick to using actual files for this kind of thing i.e. https://github.com/dosaboy/hotsos/blob/master/tests/unit/fake_data_root/sos_commands/ceph/ceph_osd_df_tree. I assume what you want here is for both files to be equivalent but one is non-json and the other json. If thats the case just go ahead and update both files and fixup existing unit tests that rely on their output. That will greatly reduce the amount of code in tests and the need for these extreme lambda functions. You can then also remove the global change in tox.ini since i assume it wont be needed anymore.

Also with your mock you are returning a string from get_ceph_osd_df_tree() but in reality it returns a list of string so this is not a valid mock value.

tools/test/.pylintrc
```diff
@@ -3,14 +3,22 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
+extension-pkg-allow-list=
```
dosaboy, line 6: These changes feel like more than "allow for
longer strings to allow for the additional artifacts in the storage
tests." And since they wont be needed if we move the test data to actual files maybe its not necessary?

tox.ini
```diff
@@ -21,6 +21,11 @@ deps =
     -r{toxinidir}/test-requirements.txt
 commands = nosetests {posargs} {toxinidir}/tests/unit
 
+[flake8]
+# Ignore long lines in flake8. This is still tested, albeit more
+# configurably, with pylint.
+ignore = E501
```
dosaboy, line 27: Not necessary?

tests/unit/test_storage.py
```diff
@@ -86,33 +86,75 @@ def test_get_ceph_versions_mismatch_unavailable(self):
...
+            "osd.4": 32,
+        }}
         storage_01ceph.get_ceph_pg_imbalance()
         self.assertEqual(storage_01ceph.CEPH_INFO, result)
 
+    @mock.patch.object(storage_01ceph.helpers, "get_ceph_osd_df_tree_json",
+                       lambda: "{}")
     @mock.patch.object(storage_01ceph.helpers, "get_ceph_osd_df_tree",
-                       lambda: [])
+                       lambda: "")
```
dosaboy, line 98: get_ceph_osd_df_tree() always returns a list not a string

dosaboy (changes requested):

dosaboy:
Since this pull request is very old and has a large number of conflicts I will go ahead and close. If we want to add support for json ceph osd tree I suggest we open a new pull request.

nicolasbock:
I completely agree @dosaboy Sorry for letting this one slide so long.

PR #66, opened 2021-04-09
Storage: add check for ceph crush map's mixed items buckets

taodd:
This patch will make hotsos able to report those crush buckets that contains mixed items.
mixed items bucket is always a problem which would cause ceph unable to recover

### DIFF
--- a/common/helpers.py
+++ b/common/helpers.py
@@ -4,6 +4,7 @@ import os
 import re
 import subprocess
 import sys
+import json
 
 # HOTSOS GLOBALS
 DATA_ROOT = os.environ.get('DATA_ROOT', '/')
@@ -189,6 +190,19 @@ def get_snap_list_all():
         return open(path, 'r').readlines()
 
     return []
+
+
+@catch_exceptions(OSError, subprocess.CalledProcessError, json.JSONDecodeError)
+def get_osd_crush_dump():
+    if DATA_ROOT == "/":
+        output = subprocess.check_output(['ceph', 'osd', 'crush', 'dump'])
+        return json.loads(output.decode('UTF-8'))
+
+    path = os.path.join(DATA_ROOT, "sos_commands/ceph/ceph_osd_crush_dump")
+    if os.path.exists(path):
+        f = open(path)
+        return json.load(f)
+    return {}
 
 
 @catch_exceptions(OSError, subprocess.CalledProcessError)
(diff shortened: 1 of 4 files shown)


taodd:
Ah, forgot to add one test file

taodd:
Now done

plugins/storage/01ceph
```diff
@@ -20,6 +20,49 @@ CEPH_SERVICES = [r"ceph-[A-Za-z]+",
 CEPH_INFO = {}
 
 
+def build_buckets_from_crushdump(crushdump):
```
dosaboy, line 23: Hi Dongdong,

I tried running this against a sosreport I have and got the following error:

  File "/home/ubuntu/git/dosaboy/hotsos/plugins/storage/01ceph", line 255, in <module>
    get_crushmap_mixed_buckets()
  File "/home/ubuntu/git/dosaboy/hotsos/plugins/storage/01ceph", line 45, in get_crushmap_mixed_buckets
    osd_crush_dump = helpers.get_osd_crush_dump()
  File "/home/ubuntu/git/dosaboy/hotsos/plugins/storage/common/helpers.py", line 30, in catch_exceptions_inner2
    return f(*args, **kwargs)
  File "/home/ubuntu/git/dosaboy/hotsos/plugins/storage/common/helpers.py", line 198, in get_osd_crush_dump
    return json.load(f)
  File "/usr/lib/python3.6/json/__init__.py", line 299, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 342, in decode
    raise JSONDecodeError("Extra data", s, end)
json.decoder.JSONDecodeError: Extra data: line 1 column 5 (char 4)

Looking at the contents of sos_commands/ceph/ceph_osd_crush_dump I see:

2021-04-05 15:19:45.664 7f4500a97700 -1 auth: unable to find a keyring on /etc/ceph/ceph.client.admin.keyring,/etc/ceph/ceph.keyring,/etc/ceph/keyring,/etc/ceph/keyring.bin,: (2) No such file or directory
2021-04-05 15:19:45.664 7f4500a97700 -1 monclient: ERROR: missing keyring, cannot use cephx for authentication
[errno 2] error connecting to the cluster

This is pretty common if e.g. sosreport is from a ceph-osd node that does not have privileges/keys to get that info. So I think we need to protect against this when we decode the contents of the file. I'd lile to also see a unit tests that covers the case where sos_commands/ceph/ceph_osd_crush_dump exists but contains non-json data.

dosaboy (changes requested):

plugins/storage/01ceph
taodd: Hmm, right, needs to think about the un-expected non-json output

PR #73, opened 2021-04-22
kubernetes snaps not listed

hemanthnakkina:
Following changes are done
- Corrected the path of snap list in sosreport
- Added kube-* to list of snaps
- Added calico-node to list of services


dosaboy (approved):
lgtm thanks!

PR #84, opened 2021-04-30
Add bashate check to `tox`

nicolasbock:
So that the shell scripts are linted as well.


nicolasbock:
I am going to work on fixing the linter issues in the next iteration @dosaboy

PR #85, opened 2021-04-30
Add base RabbitMQ queue statistics

nicolasbock:
Sometimes queues are imbalanced across a RabbitMQ cluster. Indicate a
large imbalance.


nicolasbock:
I am still working on the unit tests.

plugins/rabbitmq/parts/services.py
```diff
@@ -33,9 +38,122 @@ def get_running_services_info(self):
...
 
+    def get_queue_distribution(self):
+        """Get distribution of queues across cluster."""
+        queue_distribution = {}
+        RABBITMQ_INFO["queue distribution"] = queue_distribution
+        report_path = os.path.join(
+            constants.DATA_ROOT,
+            "sos_commands/rabbitmq/rabbitmqctl_report")
+        start_re = re.compile("Queues on ([^:]+):")
+        line_re = re.compile("^<(rabbit[^>.]*)(?:[.][0-9]+)+>.*$")
```
dosaboy, line 49: I wonder if analytics.LogSequenceStats can be of help to you here. It allows you to extract log sequences that have a start and end point then produce some stats on the results.

nicolasbock: I refactored the first method and will do the same for the other two in a follow up PR.

dosaboy (approved):

PR #91, opened 2021-05-11
Add RabbitMQ issue for severely imbalanced clusters

nicolasbock:
If one broker holds more than 2/3 of all queues raise an issue. The
threshold of 2/3 is somewhat arbitrary but a cluster that is this
imbalanced might experience some load bottle necks.

### DIFF
--- a/tests/unit/test_tools.py
+++ b/tests/unit/test_tools.py
@@ -48,6 +48,6 @@ class TestTools(utils.BaseTestCase):
                 output_filter.filter_master_yaml()
 
                 with open(ftmp.name) as fd:
-                    result = yaml.load(fd)
+                    result = yaml.load(fd, Loader=yaml.SafeLoader)
 
                 self.assertEqual(result, expected)
--- a/common/issue_types.py
+++ b/common/issue_types.py
@@ -26,3 +26,7 @@ class NeutronL3HAWarning(IssueTypeBase):
 
 class NetworkWarning(IssueTypeBase):
     pass
+
+
+class RabbitMQWarning(IssueTypeBase):
+    pass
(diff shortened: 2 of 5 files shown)


nicolasbock:
Would be nice to also unit test the issue that is raised. I'll see if I can add that.

common/issues_utils.py
```diff
@@ -31,8 +31,8 @@ def add_issue(issue):
     Fetch the current plugin issues.yaml if it exists and add new issue with
     description of the issue.
     """
-    if not os.path.isdir(PLUGIN_TMP_DIR):
-        raise Exception("plugin tmp dir  '{}' not found".
+    if PLUGIN_TMP_DIR is None or not os.path.isdir(PLUGIN_TMP_DIR):
```
dosaboy, line 34: This should never be None. If it is failing because of a unit test you need to make sure it gets set in the unit test. See https://github.com/dosaboy/hotsos/blob/master/tests/unit/test_storage.py#L21 for an example of how to do this.

dosaboy (changes requested):
Hi Nicolas, one comment for you. Also needs rebasing to fix conflicts.

common/issues_utils.py
nicolasbock: Fixed

tests/unit/test_tools.py
```diff
@@ -48,6 +48,6 @@ def test_output_filter(self):
                 output_filter.filter_master_yaml()
 
                 with open(ftmp.name) as fd:
-                    result = yaml.load(fd)
+                    result = yaml.load(fd, Loader=yaml.SafeLoader)
```
nicolasbock, line 51: This is a fix for a warning I get when running the `py3` tox test.

nicolasbock:
Rebased

nicolasbock:
Rebased and added test for the `RabbitMQWarning`

Issue #105, opened 2021-07-15
bluefs_buffered_io + kernel version check


lathiat:
Warn if bluefs_buffered_io=enable and kernel <5.4 (e.g. non-HWE for bionic): Ceph performance may be affected

bluefs_buffered_io was enabled by default in v13.2.0 and v14.2.0.
bluefs_buffered_io was disabled by default in v14.2.10 and v15.2.0.
bluefs_buffered io was re-enabled in the following point releases:
v14.2.22
v15.2.13
v16.2.0

Reference:
https://docs.google.com/document/d/1uv1Q8p2OR3hG_eEjbpP7OuS3rxmXUgpIlTG5JcpFDaY/edit#

pponnuvel:
Note: only on Ceph deployments that use bcache.

dosaboy:
The tricky thing here is that nodes running osds don't often have the osd config readily available to view (in a sosreport at least) but I think we can start by making this assertion based on package version knowing what defaults exist.

pponnuvel:
Raised https://github.com/sosreport/sos/issues/2613 to add it to sosreport.
I'll make a PR when I get time. If it gets accepted, we'll have the config of every OSD from a mon's sosreport.

If not, the table approach (ceph version mapped its default value of bluefs_buffered_io) is fine. The problem here is that once a customer corrects bluefs_buffered_io, we'll only see false positives from all subsequent sosreports.

dosaboy:
@pponnuvel the ideal would be if we could retrieve the config from an osd node so that we have it all in the same sos but i think that's be tricky due to location of keys. As a minimum we could just print a suggestion to check that bluefs_buffered_io is set to False given the right conditions.

pponnuvel:
@dosaboy We could get from individual OSD hosts, too. `ceph daemon osd.X config show` and the list of OSDs on that host can be retrieved with `mount | awk '/ceph/{l=split($3,a,"/"); split(a[l],b,"-"); print b[2]}'`. So it's possible to do that as well.

The reason I preferred to do from a mon is that we almost always get a sosreport from a mon but not from every single OSD.

Either way, we could get this info to sosreport which would be useful to know all the configs (not just bluefs_buffered_io). The version table can be an [alternative solution](https://dictionary.cambridge.org/dictionary/english/workaround) until this info gets added to sosreport.

dosaboy:
[...]

PR #106, opened 2021-07-15
[storage] check for ceph-osd bcache vulnerability

dosaboy:
If ceph osds are using bcache and bluefs_buffered_io is True
when using older kernels the host will be susceptible to major
performance issues as described in LP bug 1936136. If we think
these conditions are met we will raise an issue to warn the
user.

Resolves: #105

### DIFF
--- a/common/checks.py
+++ b/common/checks.py
@@ -2,6 +2,7 @@ import os
 
 import operator
 import re
+import subprocess
 import yaml
 
 from common import (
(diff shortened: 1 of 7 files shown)


pponnuvel (approved):
LGTM.

dosaboy:
just added a last minute cache_available_percent > 70 threshold for raising the alarm

PR #114, opened 2021-08-23
Changes for oom checks in hotsos

nkshirsagar:


plugins/kernel/pyparts/memory.py
```diff
@@ -9,10 +10,39 @@
...
+            for label in defs:
+                _results = results.find_by_tag(label)
+                if label == "oom":
+                    if _results:
+                        first_process_oomd = _results[0].get(3)
+                        time_oomd = _results[0].get(1) + " " +  _results[0].get(2)
+                        msg = ("oom-killer had been invoked, first for process ")
+                        issue = issue_types.MemoryWarning(msg + first_process_oomd + " at " + time_oomd);
+                        issues_utils.add_issue(issue);
+                        info['oom']=time_oomd
```
dosaboy, line 36: can we rename "oom"  to something like "oom-event"

plugins/kernel/pyparts/memory.py
```diff
@@ -9,10 +10,39 @@
...
+        """ See defs/events.yaml for definitions. """
+        info = {}
+        for defs in self.event_definitions.values():
+            for label in defs:
+                _results = results.find_by_tag(label)
+                if label == "oom":
+                    if _results:
+                        first_process_oomd = _results[0].get(3)
+                        time_oomd = _results[0].get(1) + " " +  _results[0].get(2)
+                        msg = ("oom-killer had been invoked, first for process ")
```
dosaboy, line 33: please use formatted string like:

process_name = _results[0].get(3)
"oom-killer invoked for process '{}' at {} {}".format(process_name, _results[0].get(1), _results[0].get(2))

plugins/kernel/pyparts/memory.py
```diff
@@ -9,10 +10,39 @@
     VMSTAT,
     BUDDY_INFO,
     SLABINFO,
+    KernelChecksBase
```
dosaboy, line 13: imported twice?

defs/events.yaml
```diff
@@ -85,6 +85,13 @@ openstack:
           expr: '(\S+ \d+) \d+:\S+\s+.+apparmor="DENIED".+\s+profile="(\S+neutron\S+)"\s+.+'
           hint: 'apparmor'
 kernel:
+  memory-check:
+    kernlog:
+      oom:
+        datasource: 'var/log/kern.log'
+        start:
+          expr: '(.+ \d+) (\d+:\d+:\d+) .+ (\S+) invoked oom-killer\: .+'
```
dosaboy, line 93: you can lose the .+ at the end since everything will match up until that point and we don't need anything beyond.

dosaboy (changes requested):
looks good, couple of things to fix. be good to have a unit test.

plugins/kernel/pyparts/memory.py
nkshirsagar: ack

plugins/kernel/pyparts/memory.py
nkshirsagar: corrected, thank you

defs/events.yaml
nkshirsagar: ack, done.

dosaboy (commented):
Everything looks good now thanks Nikhil. I small question I noticed that with the kern.log bits you added I get:

    - desc: oom-killer invoked for process 'nice' at Aug  3 08:32:23
      origin: kernel.memory
      type: MemoryWarning

i.e. it is saying that the process that got oom killed is "nice" but I think it should actually be "ruby"? I guess the issue is that the information you need is on the previous line. It is however possible to get that info by taking advantage of the fact that the events.yaml entry can have a start and end defined (i.e. a sequence) so that you can get that info from either the start or end match result.

Aug  3 10:31:24 compute4 kernel: [5490487.297212] Killed process 55438 (ruby) total-vm:1771344kB, anon-rss:1469448kB, file-rss:4992kB
Aug  3 08:32:23 compute4 kernel: [5489652.470650] nice invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=876

nkshirsagar:
> Everything looks good now thanks Nikhil. I small question I noticed that with the kern.log bits you added I get:
> 
> ```
> - desc: oom-killer invoked for process 'nice' at Aug  3 08:32:23
>   origin: kernel.memory
>   type: MemoryWarning
> ```
> 
> i.e. it is saying that the process that got oom killed is "nice" but I think it should actually be "ruby"? I guess the issue is that the information you need is on the previous line. It is however possible to get that info by taking advantage of the fact that the events.yaml entry can have a start and end defined (i.e. a sequence) so that you can get that info from either the start or end match result.
> 
> Aug 3 10:31:24 compute4 kernel: [5490487.297212] Killed process 55438 (ruby) total-vm:1771344kB, anon-rss:1469448kB, file-rss:4992kB
> Aug 3 08:32:23 compute4 kernel: [5489652.470650] nice invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=876

Hi Ed,

the logs I put are part of a larger bunch. I probably should have only included the "invoked oom-killer" line in the log. 

For eg,

nikhil@nikhil-Lenovo-Legion-Y540-15IRH-PG0:~/Downloads/sosreport-gcc-prod-vm1-2021-07-27-toyylmm/var/log$ cat kern.log | grep -i "invoked" | grep ruby
Jul 25 09:08:58 gcc-prod-vm1 kernel: [5491847.218834] ruby invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=876
Jul 25 09:31:44 gcc-prod-vm1 kernel: [5493212.754459] ruby invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=876
Jul 25 10:19:10 gcc-prod-vm1 kernel: [5496058.736570] ruby invoked oom-killer: gfp_mask=0x24000c0, order=0, oom_score_adj=876
...

It's better to point to where the issue begins in the logs with that timestamp because once processes start being terminated due to OOM there could be several more instances like we see here in this sos. 

Please let me know what you think.

Regards,
Nikhil.

dosaboy (approved):

Issue #110, opened 2021-08-17
create alert if softlockups detected in kern.log


dosaboy:

nivseg2:
This is as simple as just simply doing a grep of "Call" in the kernel log
to see if a kernel stack trace has been logged - usually either due to
a watchdog timer of some sort going off (common 120 sec one) or 
any other oops etc. or kernel issue which dumps kernel traces. They
are sometimes benign, sometimes merely a resource issue, or a bug
where an event that shouldn't occur does. Can look at an example 
stack trace and grab the key lines in question to describe what event
has occurred. I'll add that here.

PR #120, opened 2021-09-09
[kernel] alert error if Call Trace found

dosaboy:
If we detect a stacktrace in kern.log* we will
alert with a KernelError.

Resolves: #110


nivseg2 (approved):
Looks good.

Issue #122, opened 2021-09-10
Raise issue when Octavia o-hm0 port does not have ip address


dosaboy:
If the healthmanager port inside an octavia host does not have an address it will no be able to reach any amphora vms so we should alert when this happens.

PR #126, opened 2021-09-13
Add Octavia HM port to network configs

hemanthnakkina:
This patch adds Octavia Health manager IP address to the
network configs section if o-hm0 port exists. If there is
no IP address to o-hm0, will be notified in potential issues.

Resolves: #122

### DIFF
--- a/core/plugins/openstack/__init__.py
+++ b/core/plugins/openstack/__init__.py
@@ -288,6 +288,19 @@ class OpenstackBase(object):
         return interfaces
 
     @property
+    def octavia_bind_interfaces(self):
+        """
+        Fetch interface o-hm0 used by Openstack Octavia. Returned dict is
+        keyed by config key used to identify interface.
+        """
+        interfaces = {}
+        port = self.nethelp.get_interface_with_name('o-hm0')
+        if port:
+            interfaces.update({'o-hm0': port})
+
+        return interfaces
+
+    @property
     def bind_interfaces(self):
         """
         Fetch interfaces used by Openstack services and return dict.
--- a/plugins/openstack/pyparts/network.py
+++ b/plugins/openstack/pyparts/network.py
@@ -1,5 +1,9 @@
 import re
 
+from core.issues import (
+    issue_types,
+    issue_utils,
+)
 from core.cli_helpers import CLIHelper
 from core.plugins.openstack import OpenstackChecksBase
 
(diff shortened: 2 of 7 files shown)


plugins/openstack/pyparts/network.py
```diff
@@ -69,6 +73,17 @@ def get_config_network_info(self):
             config_info['nova'][key] = "{} ({})".format(port.addresses[0],
                                                         port.name)
 
+        for key, port in self.octavia_bind_interfaces.items():
+            if 'octavia' not in config_info:
+                config_info['octavia'] = {}
+
+            if len(port.addresses) > 0:
```
dosaboy, line 80: you can shorten this to just 'if port.addresses:'

plugins/openstack/pyparts/network.py
```diff
@@ -69,6 +73,17 @@ def get_config_network_info(self):
...
+                config_info['octavia'][key] = "{} ({})".format(
+                        port.addresses[0], port.name)
```
dosaboy, line 82: This assumes that the interface will only ever have one address which afaik is fine.

plugins/openstack/pyparts/network.py
```diff
@@ -69,6 +73,17 @@ def get_config_network_info(self):
...
+            else:
+                msg = ("Octavia Health manager port has no IP address")
```
dosaboy, line 84: Lets elaborate this a bit e.g.:

"No IP address found on Octavia healthmanager port ({}). Octavia will not be able to communicate with Amphora vms - please investigate.".format(port.name)

dosaboy (changes requested):

core/plugins/openstack/__init__.py
```diff
@@ -287,6 +287,19 @@ def neutron_bind_interfaces(self):
...
+    @property
+    def octavia_bind_interfaces(self):
+        """
+        Fetch interface o-hm0 used by Openstack Octavia. Returned dict is
+        keyed by config key used to identify interface.
+        """
+        interfaces = {}
+        port = self.nethelp.get_interface_with_name('o-hm0')
+        if port:
+            interfaces.update({'o-hm0': port})
```
dosaboy, line 299: Is there are an actual config key name we can use instead of "o-hm0"?

dosaboy (changes requested):

plugins/openstack/pyparts/network.py
hemanthnakkina: Done

plugins/openstack/pyparts/network.py
hemanthnakkina: Yes

plugins/openstack/pyparts/network.py
hemanthnakkina: o-hm0 is hardcoded in charm-octavia https://opendev.org/openstack/charm-octavia/src/branch/master/src/lib/charm/openstack/octavia.py#L39

core/plugins/openstack/__init__.py
hemanthnakkina: o-hm0 is hardcoded in charm-octavia https://opendev.org/openstack/charm-octavia/src/branch/master/src/lib/charm/openstack/octavia.py#L39

dosaboy (approved):
lgtm thanks hemanth!

Issue #131, opened 2021-09-17
Ceph OSD Map Trimming Warning


lathiat:
Report on ceph_report .osdmap_manifest.pinned_maps list too long

jq '.osdmap_manifest.pinned_maps | length'

$ cat report-0.txt |jq '.osdmap_manifest.pinned_maps | length'

Seems possibly 5000 is 'bad' and 500 is 'normal' from https://docs.ceph.com/en/latest/dev/mon-osdmap-prune/

Can be related to this bug:
https://tracker.ceph.com/issues/44184

Doesn't exist on luminous so needs to gracefully degrade if data is not there (plus sometimes its not valid json if its run on an osd node)

dosaboy:
@lathiat can you tell me where that information would be found in a sosreport?

hillpd:
You can get it from `ceph report` within a sosreport.

As an example:
`ceph report | jq '.osdmap_manifest.pinned_maps | length'`

PR #138, opened 2021-09-25
[storage] Detect if there are too many osdmaps pinned

pponnuvel:
Mons store a number of osdmaps (default = 500). But this can
get arbitrarily large if the cluster wasn't in healthy state
for a while. These are trimmed down automatically *if* the
cluster comes back to healthy state. However, too many
maps in itself can be a problem and can also indicate that
trimming isn't happening due to bugs.

Closes: #131


plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -25,6 +25,44 @@ def __init__(self):
         super().__init__()
         self.ceph_osd_df_tree = self.cli.ceph_osd_df_tree()
         self.ceph_versions = self.cli.ceph_versions()
```
dosaboy, line 27: Hi Pon, this mad me realise that the existing code needed tidying up and this __init__ method no longer exists. Can you move your ceph_report bits to check_osdmaps_size() and also if you are not actually using the cached ceph_report you don't need to cache it so you can remove that too (from self.cli_cache). Either that or do the json cleanup stuff as part of adding to self.cli_cache and use that instead of calling self.cli directly.

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -25,6 +25,44 @@ def __init__(self):
...
+            # any recovery/rebalance happens. We arbitrarily chose 5000 as the
+            # "bad" limit (10 x default).
+            if osdmaps_count >= 5000:
+                msg = ("Found {} pinned osdmaps. This can affect mon's"
+                       "performance and also indicate bugs such as "
+                       "https://tracker.ceph.com/issues/44184 and "
+                       "https://tracker.ceph.com/issues/47290"
+                       .format(osdmaps_count))
+                issue_utils.add_issue(issue_types.CephMapsWarning(msg))
+        except Exception:
```
dosaboy, line 64: Under what conditions would this raise an exception? I'd rather something a bit more explicit/descriptive than a catch-all with no explanation.

dosaboy (changes requested):
See comments and will need a rebase to resolve conflicts.

dosaboy (changes requested):

core/cli_helpers.py
```diff
@@ -349,6 +349,9 @@ def command_catalog(self):
             'ceph_volume_lvm_list':
                 [BinCmd('ceph-volume lvm list'),
                  FileCmd('sos_commands/ceph/ceph-volume_lvm_list')],
+            'ceph_report_json_decoded':
+                [BinCmd('ceph report', json_decode=True),
+                 FileCmd('sos_commands/ceph/ceph_report')],
```
dosaboy, line 354: I assume you are not doing json_decode=True for both because the sosreport one has the spurious data at the end of the file. Because of that I think we should create a custom FileCmd like:

class CephReportFileCmd(FileCmd):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.register_hook("pre-exec", self.sanitise_json_contents)

    def sanitise_json_contents(self, **kwargs):
        ...

And move the code that does the fix for the last line in the file to here.

PR #143, opened 2021-10-01
[storage] Fix unit test for mixed crush buckets

pponnuvel:

### DIFF
--- a/plugins/storage/pyparts/ceph_daemon_checks.py
+++ b/plugins/storage/pyparts/ceph_daemon_checks.py
@@ -190,7 +190,7 @@ class CephOSDChecks(ceph.CephChecksBase):
                 issue = issue_types.CephDaemonWarning(msg)
                 issue_utils.add_issue(issue)
 
-    def build_buckets_from_crushdump(self, crushdump):
+    def _build_buckets_from_crushdump(self, crushdump):
         buckets = {}
         # iterate jp for each bucket
         for bucket in crushdump["buckets"]:
@@ -217,7 +217,7 @@ class CephOSDChecks(ceph.CephChecksBase):
             return
 
         bad_buckets = []
-        buckets = self.build_buckets_from_crushdump(osd_crush_dump)
+        buckets = self._build_buckets_from_crushdump(osd_crush_dump)
         # check all bucket
         for bid in buckets:
             items = buckets[bid]["items"]
@@ -236,11 +236,11 @@ class CephOSDChecks(ceph.CephChecksBase):
                 bad_buckets.append(buckets[bid]["name"])
 
         if bad_buckets:
-            msg = ("mixed crush buckets identified (see --storage for more "
-                   "info)")
+            msg = ("mixed crush bucket types identified in buckets '{}'. "
+                   "This can cause data distribution to become skewed - "
+                   "please check crush map".format(bad_buckets))
             issue = issue_types.CephCrushWarning(msg)
             issue_utils.add_issue(issue)
-            self._output["mixed_crush_buckets"] = bad_buckets
 
     def check_bcache_vulnerabilities(self):
         has_bcache = False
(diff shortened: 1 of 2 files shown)


plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -236,11 +236,10 @@ def get_crushmap_mixed_buckets(self):
                 bad_buckets.append(buckets[bid]["name"])
 
         if bad_buckets:
-            msg = ("mixed crush buckets identified (see --storage for more "
-                   "info)")
+            msg = ("mixed crush buckets identified in buckets {}."
+                   .format(bad_buckets))
```
dosaboy, line 240: We could perhaps go a bit further with this message and allude to what should be checked/fixed e.g.

msg = ("mixed crush bucket types identified in buckets '{}'. This can cause data distribution to become skewed - please check crush map".format(bad_buckets))

tests/unit/fake_data_root/sos_commands/ceph/ceph_osd_crush_dump
```diff
@@ -90,7 +90,12 @@
...
                     "weight": 6396,
                     "pos": 2
-                }
+                },
+                {   
+                    "id": -9, 
+                    "weight": 6396,
+                    "pos": 3
+                }   
             ]
```
dosaboy, line 99: I would like if possible to keep the contents of this file "meaningful" since they are currently based on a real ceph cluster that is the same as the other files in the sos_commands/ceph dir. To do this kind of this i would either add comments either side of the additional text (it's json so should ok to add # comments) or put it inline in the unit test (which might not work so well here since its a lot of text).

tests/unit/fake_data_root/sos_commands/ceph/ceph_osd_crush_dump
```diff
@@ -207,6 +212,22 @@
             "weight": 6396,
             "alg": "straw2",
             "hash": "rjenkins1",
+            "items": [
+                {
+                    "id": -7,
```
dosaboy, line 217: same as above and also should this bucket unchanged? the id is changed from 2 -> -7 here.

dosaboy (changes requested):

tests/unit/fake_data_root/sos_commands/ceph/ceph_osd_crush_dump
dosaboy: hey @pponnuvel i see you moved this to the unit test code so I assume this change can be removed?

pponnuvel: @dosaboy Was still making changes (one more test). Done now & ready.

dosaboy: ah sorry about that, looks better now. One last thing on the unit test, ill comment inline.

tests/unit/test_storage.py
```diff
@@ -201,10 +281,23 @@ def test_ceph_base_interfaces(self):
...
+                return_value = json.loads(CEPH_OSD_CRUSH_DUMP)
+            inst = ceph_daemon_checks.CephOSDChecks()
+            inst.get_crushmap_mixed_buckets()
+            self.assertTrue(mock_issue_utils.add_issue.called)
+
+    @mock.patch.object(ceph_daemon_checks, 'issue_utils')
+    def test_get_crushmap_no_mixed_buckets(self, mock_issue_utils):
+        with mock.patch.object(ceph_core, 'CLIHelper') as mock_helper:
+            mock_helper.return_value = mock.MagicMock()
+            inst = ceph_daemon_checks.CephOSDChecks()
```
dosaboy, line 298: since the existing data does not have mixed buckets there is no need to mock cli_helpers so you can remove the above 3 lines.

dosaboy (changes requested):

tests/unit/test_storage.py
pponnuvel: ack, done

PR #147, opened 2021-10-06
[rabbitmq] detect bug LP 1943937

dosaboy:


defs/bugs.yaml
```diff
@@ -73,6 +73,13 @@ openstack:
       hint: 'AttributeError'
       reason: 'identified bug impacting OVN sbdb connections from ovn agents'
       path: 'var/log/neutron/neutron-ovn-metadata-agent.log'
+rabbitmq:
+  common:
+    1943937:
+      expr: 'operation queue.declare caused a channel exception not_found: failed to perform operation on queue ''\S+'' in vhost ''\S+'' due to timeout'
+      hint: 'not_found'
+      reason: 'known issue where rabbit server loses queues - this can cause clients to become stuck since their messages will never be consumed'
```
lathiat, line 81: Though we link to the LP bug, given how long it is and that the solution is non-obvious, I think I'd change this message to include the solution and that rolling restarts don't work. Something like:

'Known issue where RabbitMQ queues get stuck and clients trying to use them will just keep timing out. This stops many services in the cloud from working correctly. Resolution requires you to stop all RabbitMQ servers before starting them all again at the same time. A rolling restart or restarting them simultaneously will not work. See bug for more detail.'

lathiat (changes requested):
Would adjust the reason message but looks good otherwise.

Issue #142, opened 2021-09-29
[storage] check if mon is older than others


pponnuvel:
We currently check if all the daemons and clients are of the same version and if not, issue a warning.
We need to tweak slightly so that mon's version behind others, it should be considered as an ERROR.

PR #149, opened 2021-10-08
[storage] Check for lower mon version

nkshirsagar:
Check if there is any mon on a lower version than any
other component in the cluster.

Closes: #142

Signed-off-by: Nikhil Kshirsagar <nkshirsagar@gmail.com>

### DIFF
--- a/core/issues/issue_types.py
+++ b/core/issues/issue_types.py
@@ -36,6 +36,10 @@ class CephDaemonWarning(IssueTypeBase):
     pass
 
 
+class CephDaemonVersionsError(IssueTypeBase):
+    pass
+
+
 class JujuWarning(IssueTypeBase):
     pass
 
--- a/plugins/storage/pyparts/ceph_daemon_checks.py
+++ b/plugins/storage/pyparts/ceph_daemon_checks.py
@@ -160,6 +160,13 @@ class CephOSDChecks(ceph.CephChecksBase):
             issue = issue_types.CephCrushWarning(msg)
             issue_utils.add_issue(issue)
 
+    @staticmethod
+    def version_as_a_tuple(ver):
+        """
+        Return a version string as a tuple for easy comparison
+        """
+        return tuple(map(int, (ver.split("."))))
+
     def get_ceph_versions_mismatch(self):
         """
         Get versions of all Ceph daemons.
@@ -170,6 +177,11 @@ class CephOSDChecks(ceph.CephChecksBase):
 
         global_vers = set()
         daemon_version_info = {}
+
+        # these store highest ver and daemon name with highest ver
+        h_version = "0.0.0"
+        h_daemon = ""
+
         for daemon_type in versions:
             # skip the catchall
             if daemon_type == 'overall':
@@ -179,7 +191,11 @@ class CephOSDChecks(ceph.CephChecksBase):
             for version in versions[daemon_type]:
                 vers.append(version)
                 global_vers.add(version)
-
+                # store the highest version any component has
+                if self.version_as_a_tuple(version) > \
+                        self.version_as_a_tuple(h_version):
+                    h_version = version
+                    h_daemon = daemon_type
             if vers:
                 daemon_version_info[daemon_type] = vers
 
@@ -189,6 +205,15 @@ class CephOSDChecks(ceph.CephChecksBase):
                 msg = ('ceph daemon versions not aligned')
                 issue = issue_types.CephDaemonWarning(msg)
                 issue_utils.add_issue(issue)
+
+            # check if mon is lower than highest version we stored earlier
+            for version in versions.get('mon', []):
+                if self.version_as_a_tuple(version) < \
+                      self.version_as_a_tuple(h_version):
+                    msg = ("mon version {} is lower than {} version {}"
+                           .format(version, h_daemon, h_version))
+                    issue = issue_types.CephDaemonVersionsError(msg)
+                    issue_utils.add_issue(issue)
 
     def _build_buckets_from_crushdump(self, crushdump):
         buckets = {}
--- a/tests/unit/test_storage.py
+++ b/tests/unit/test_storage.py
@@ -340,6 +340,32 @@ class TestStoragePluginPartCephDaemonChecks(StorageTestsBase):
             inst.get_ceph_versions_mismatch()
             self.assertEqual(inst.output["ceph"]["versions"], result)
 
+    @mock.patch.object(ceph_daemon_checks.issue_utils, "add_issue")
+    def test_get_ceph_mon_lower_version(self, mock_add_issue):
+        issues = []
+
+        def fake_add_issue(issue):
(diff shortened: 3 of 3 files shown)


tests/unit/test_storage.py
```diff
@@ -305,6 +305,14 @@ def test_get_ceph_versions_mismatch(self):
         inst.get_ceph_versions_mismatch()
         self.assertEqual(inst.output["ceph"]["versions"], result)
 
+    def test_get_ceph_mon_lower_version(self):
+        result = {'mgr': ['15.2.13'],
```
pponnuvel, line 309: This test checks whether `result` and the versions found in sos_commands/ceph/ceph_versions differ.
That would still be the case if sos_commands/ceph/ceph_versions contains the mismatched one as it's not different to `result`.

You'd want to test if `result` is identified as a problematic version set. You can instead use `CEPH_VERSIONS_MISMATCHED` for this test (modify the mon version there to lower one) and test whether both mismatches are identified (similar to `test_get_ceph_pg_imbalance`).

core/issues/issue_types.py
```diff
@@ -36,6 +36,10 @@ class CephDaemonWarning(IssueTypeBase):
     pass
 
 
+class CephDaemonError(IssueTypeBase):
```
dosaboy, line 39: I'd like to change this CephDaemonVersionsError since CephDaemonError feels more appropriate for when an error has occured with a specific daemon.

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -160,6 +160,10 @@ def get_ceph_pg_imbalance(self):
             issue = issue_types.CephCrushWarning(msg)
             issue_utils.add_issue(issue)
 
+    @staticmethod
+    def version_compare(ver):
```
dosaboy, line 164: Can you name this in a way that better explains what it is to be used for and also a docstring and associated unit test. If you think this is a good way to compare ceph daemon version it might be better to put it somewhere like https://github.com/canonical/hotsos/blob/6a0d650a8d76b5a5f85f4ddc8c0a9f8939e1de7a/core/plugins/storage/ceph.py#L186

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -179,10 +188,23 @@ def get_ceph_versions_mismatch(self):
             for version in versions[daemon_type]:
                 vers.append(version)
                 global_vers.add(version)
-
+                if self.version_compare(str(version)) > \
```
dosaboy, line 191: if version is not already a string we should fix that in https://github.com/canonical/hotsos/blob/master/core/plugins/storage/ceph.py#L134 rather than having to convert on the fly.

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -179,10 +188,23 @@ def get_ceph_versions_mismatch(self):
...
+                        self.version_compare(h_version):
+                    h_version = version
+                    h_daemon = daemon_type
             if vers:
                 daemon_version_info[daemon_type] = vers
 
+        for daemon_type in versions:
+            if daemon_type == 'mon':
+                for version in versions[daemon_type]:
```
dosaboy, line 200: You could shorted the above 3 lines to:

for version in versions.get('mon', []):

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -179,10 +188,23 @@ def get_ceph_versions_mismatch(self):
...
+                    if self.version_compare(str(version)) < \
+                          self.version_compare(h_version):
+                        msg = ("mon version {} is lower than {} version {}"
+                               .format(version, h_daemon, h_version))
+                        issue = issue_types.CephDaemonError(msg)
+                        issue_utils.add_issue(issue)
+
```
dosaboy, line 207: Does this cover all scenarios where a mismatch occurs? My understanding is that we are looking to identify if one or more daemons has versions that break the ordering required by Ceph [1] and that could apply to more than just mons iiuc.
 
[1] https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/2/html/installation_guide_for_ubuntu/upgrading_ceph_storage_cluster

dosaboy (changes requested):

tests/unit/test_storage.py
dosaboy: One more thing. I found a bug in the  code that parses ceph versions output and have fixed it in https://github.com/canonical/hotsos/commit/5069fa2fb9647396dab6899d9aff82bc78ca4316. Please rebase your patch on top of master to pull that in (git rebase master).

plugins/storage/pyparts/ceph_daemon_checks.py
nkshirsagar: Hi Ed,

Looking at https://github.com/canonical/hotsos/issues/142 the description seems to suggest that we should issue an error if a mon's version is behind any other components version. (including other mons if present)

(For the general check of mismatched packages, i.e all packages not on same version, we already do issue a warning.)

I did see the doc you linked, especially the suggestion to upgrade in the following order,

Monitor nodes
OSD nodes
Ceph Object Gateway nodes
All other Ceph client nodes

So do you mean that we would need to also check if an OSD node is behind an object gateway node, or any other client node (rgw/mds)? , and similarly also check if an object gateway node is behind any other client node except a mon or osd?

tests/unit/test_storage.py
nkshirsagar: ack I will check on this

core/issues/issue_types.py
nkshirsagar: ack

plugins/storage/pyparts/ceph_daemon_checks.py
nkshirsagar: this method returns a tuple of ints of all the numbers in a version , so we can compare using < or > . version_compare could probably be called "return_version_as_tuple" perhaps. I could put it in the CephDaemonBase if that's a better place for it, I will discuss further and then make the changes as required.

plugins/storage/pyparts/ceph_daemon_checks.py
nkshirsagar: ack thank you

plugins/storage/pyparts/ceph_daemon_checks.py
nkshirsagar: I verified the conversion to string is not necessary, I will remove that

plugins/storage/pyparts/ceph_daemon_checks.py
nkshirsagar: As discussed, I am leaving the check as is, i.e we will issue an error if a mon is lower than any other cluster component. As discussed, for other components, the existing warning will suffice.

Issue #118, opened 2021-09-07
blueFS log can get too big


pponnuvel:
In older Ceph releases, we might hit the bug https://tracker.ceph.com/issues/45903

If the blueFS log is too big, we can detect and report this.

A possible way to manually shrink (Nautilus or newer): https://docs.ceph.com/en/latest/man/8/ceph-bluestore-tool/#bluefs-log-rescue

Trent: "You can diagnose this situation as teh “META” column from “ceph osd df tree” is large, 10-1200GB. Normally should be maybe 0-4GB".

PR #150, opened 2021-10-08
[storage] Detect if BlueFS metadata size is too big

pponnuvel:
Also modified to the json formatted ceph-osd-df-tree
command which is much easier to parse.

Closes: #118


core/cli_helpers.py
```diff
@@ -400,9 +400,12 @@ def command_catalog(self):
             'ceph_osd_dump':
                 [BinCmd('ceph osd dump'),
                  FileCmd('sos_commands/ceph/ceph_osd_dump')],
-            'ceph_osd_df_tree':
-                [BinCmd('ceph osd df tree'),
-                 FileCmd('sos_commands/ceph/ceph_osd_df_tree')],
+            'ceph_osd_df_tree_json_decoded':
```
dosaboy, line 403: While i agree that this provides a much better way to process the information, it seems it is only available in more recent sosreports so maybe we should leave both so that the code can have a fallback?

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -97,6 +96,36 @@ def check_osd_msgr_protocol_versions(self):
...
+                    # Usually the meta data is expected to be in 0-4G range
+                    # and we check if it's over 10G
+                    if meta_kb > OSD_META_LIMIT_KB:
+                        bad_meta_osds.append(device['name'])
+        except (KeyError, ValueError):
+            return
+
+        if bad_meta_osds:
+            msg = ("{} OSDs have metadata size larger than 10G. This indicates"
+                   " compaction failure/bug. Possibly affected by "
```
dosaboy, line 122: i prefer to be consistent about having whitespace at the end of the line rather than the start (makes it easier to spot mistakes)

core/cli_helpers.py
pponnuvel: It's available in the current sosreport package of Xenial, Bionic, etc (anyone using sosreport >= 3.8 will have it).

Do you think users might be running w/ older sosreport packages? (we could ask them to upgrade sosreport package, if needed). Otherwise, we're OK.

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -97,6 +98,63 @@ def check_osd_msgr_protocol_versions(self):
...
+            # Check the json data first
+            ceph_osd_df_tree_json = self.cli.ceph_osd_df_tree_json_decoded()
+            for device in ceph_osd_df_tree_json['nodes']:
+                if device['id'] >= 0:
+                    meta_kb = device['kb_used_meta']
+                    # Usually the meta data is expected to be in 0-4G range
+                    # and we check if it's over 10G
+                    if meta_kb > OSD_META_LIMIT_KB:
+                        bad_meta_osds.append(device['name'])
+        except Exception:
```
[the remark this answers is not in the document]
dosaboy: Also I tested with the latest version of sosreport in Focal and it does not contain the json decoded df tree:

root@juju-4eb7e9-2-lxd-0:/home/ubuntu# apt-cache policy sosreport
sosreport:
  Installed: 4.1-1ubuntu0.20.04.3
  Candidate: 4.1-1ubuntu0.20.04.3
  Version table:
 *** 4.1-1ubuntu0.20.04.3 500
        500 http://archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages
        100 /var/lib/dpkg/status
     3.9-1ubuntu2 500
        500 http://archive.ubuntu.com/ubuntu focal/main amd64 Packages
root@juju-4eb7e9-2-lxd-0:/home/ubuntu# ls sosreport-juju-4eb7e9-2-lxd-0-2021-10-11-errwpwa/sos_commands/ceph/ceph_osd_df_*
sosreport-juju-4eb7e9-2-lxd-0-2021-10-11-errwpwa/sos_commands/ceph/ceph_osd_df_tree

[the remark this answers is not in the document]
pponnuvel: > Can we find a better way to write this that doesn't rely on catching Exception?

I think it'll fail if when indexing the list if the list is empty/None.

I could handle the error condition, specific exceptions and use a flag. I thought the catch-all Exception conveys the intent better ("If *anything* goes wrong, maybe check the other file").

Modified as such. I don't particularly feel it's better but see if you like it ;-)

tests/unit/test_storage.py
```diff
@@ -132,6 +132,180 @@
     }
 """
 
+CEPH_OSD_DF_TREE = """
```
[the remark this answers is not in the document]
pponnuvel: I thought you don't want to modify them because the current one has been taken out of a real sosreport?

If copy this over to `sos_commands/ceph/json_output/ceph_osd_df_tree_--format_pretty`, it'll not match w/ its plain-text counterpart in `sos_commands/ceph/ceph_osd_df_tree`. If I replace both of them w/ from a real one, it may not be in sync w/ the rest of the stuff in sos_commands/ceph/ :D 

I can move this to tests/unit/fake_data_root if you prefer.

So I thought a string will not cause w/ any such issues (albeit being large & ugly). Perhaps we can move all these test-specific strings to another file and import it?

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -97,6 +96,29 @@ def check_osd_msgr_protocol_versions(self):
             msg = ("{} OSDs do not bind to v2 address".format(len(v1_osds)))
             issue_utils.add_issue(issue_types.CephOSDWarning(msg))
 
+    def check_ceph_bluefs_size(self):
+        """
+        Check if the BlueFS metadata size is too large
+        """
+        bad_meta_osds = []
+        ceph_osd_df_tree = self.cli.ceph_osd_df_tree_json_decoded()
+        for device in ceph_osd_df_tree['nodes']:
```
pponnuvel, line 105: Now that we always expect to have the data, this would fail for an empty list as we have decided not to catch exceptions which would otherwise hide buggy code or missing data in sosreport, etc.

pponnuvel: Raised https://github.com/canonical/hotsos/issues/157

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -107,39 +129,20 @@ def get_ceph_pg_imbalance(self):
...
-                    if pgs > OSD_PG_MAX_LIMIT:
-                        error_pgs[osd_id] = pgs
-
-                    margin = abs(100 - (float(100) / OSD_PG_OPTIMAL_NUM * pgs))
-                    # allow 30% margin from optimal OSD_PG_OPTIMAL_NUM value
-                    if margin > 30:
-                        suboptimal_pgs[osd_id] = pgs
-            except IndexError:
-                pass
+        ceph_osd_df_tree = self.cli.ceph_osd_df_tree_json_decoded()
```
dosaboy, line 134: you need to do:

if not ceph_osd_df_tree:
    return

dosaboy (changes requested):

plugins/storage/pyparts/ceph_daemon_checks.py
```diff
@@ -97,6 +96,29 @@ def check_osd_msgr_protocol_versions(self):
             msg = ("{} OSDs do not bind to v2 address".format(len(v1_osds)))
             issue_utils.add_issue(issue_types.CephOSDWarning(msg))
 
+    def check_ceph_bluefs_size(self):
+        """
+        Check if the BlueFS metadata size is too large
+        """
+        bad_meta_osds = []
+        ceph_osd_df_tree = self.cli.ceph_osd_df_tree_json_decoded()
```
dosaboy, line 104: you need to do:

if not ceph_osd_df_tree:
    return

Issue #104, opened 2021-07-08
RabbitMQ unsynchronised queue issues


lathiat:
Alert on the following log lines:
`rmq3    | 2019-06-27 09:20:59.926 [warning] <0.943.0> Mirrored queue 'rmq-two-queue' in vhost '/': Stopping all nodes on master shutdown since no synchronised slave is available

09:26:25.793 [error] Discarding message {'$gen_call',{<0.753.0>,#Ref<0.989368845.173015041.56949>},{info,[name,pid,slave_pids,synchronised_slave_pids]}} from <0.753.0> to <0.943.0> in an old incarnation (3) of this node (1)
`

And suggest to stop all rabbitmq nodes, verify they are stopped, then start them all again. This should recover the issue from past experience. 

Reference: https://github.com/rabbitmq/rabbitmq-server/issues/2045

lathiat:
Another item to catch:

=ERROR REPORT==== 8-Sep-2021::06:15:24 ===
Channel error on connection <0.7841.3050> (192.168.4.46:46970 -> 192.168.5.14:5672, vhost: 'openstack', user: 'nova'), channel 1:
operation queue.declare caused a channel exception not_found: failed to perform operation on queue 'reply_182c56778990474b8430713d586a4ba8' in vhost 'openstack' due to timeout

PR #151, opened 2021-10-09
Add log output for unsynchronized queues

nicolasbock:
See upstream issue [1] for more details.

[1] https://github.com/rabbitmq/rabbitmq-server/issues/2045

Closes: canonical/hotsos#104


tests/unit/fake_data_root/var/log/rabbitmq/rabbit@juju-ff6a31-19-lxd-18.log
```diff
@@ -1,2 +1,7 @@
+
 operation queue.declare caused a channel exception not_found: failed to perform operation on queue 'test_exchange_queue' in vhost 'nagios-rabbitmq-server-0' due to timeout
 
+Mirrored queue 'rmq-two-queue' in vhost '/': Stopping all nodes on master shutdown since no synchronised slave is available
+
+Discarding message {'$gen_call',{<0.753.0>,#Ref<0.989368845.173015041.56949>},{info,[name,pid,slave_pids,synchronised_slave_pids]}} from <0.753.0> to <0.943.0> in an old incarnation (3) of this node (1)
```
nicolasbock, line 6: Is there a way to add a second entry in `bugs.yaml` for the same issue, or use 2 `expr`?

dosaboy: yes, the expr field can be a list as well as a string - see https://github.com/canonical/hotsos/blob/6a0d650a8d76b5a5f85f4ddc8c0a9f8939e1de7a/defs/bugs.yaml#L17

defs/bugs.yaml
```diff
@@ -102,3 +102,9 @@ rabbitmq:
         restart or restarting them simultaneously will not work. See bug for
         more detail.
       path: 'var/log/rabbitmq/rabbit@*.log'
+    'rabbitmq/rabbitmq-server#2045':
```
dosaboy, line 105: This should be an LP bug id. I see your other comment about adding multiple search expressions for the same bug and yes that is already possible e.g.

123456:
  expr:
    - 'expr1'
    - 'expr2'

dosaboy (changes requested):

defs/bugs.yaml
dosaboy: you've got two options;

1. open a bug and use that here
2. move this to events.yaml which would allow you to pull some information e.g. frequency

nicolasbock: I moved everything to `events.yaml`. Please have another look.

core/plugins/rabbitmq.py
```diff
@@ -28,7 +28,7 @@ def plugin_runnable(self):
         if self.apt_check.core:
             return True
 
-        return False
+        return True
```
nicolasbock, line 31: To be honest I am not sure how this is supposed to be enabled.

plugins/rabbitmq/pyparts/synchronization_checks.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def __init__(self):
+        super().__init__(yaml_defs_group='synced-queues',
+                         callback_helper=EVENTCALLBACKS)
+
+    @EVENTCALLBACKS.callback
+    def no_sync(self, event):  # pylint: disable=W0613
+        msg = ("Transient mirrored classic queues are not deleted when there "
+               "are no replicas available for promotion. Please stop all "
+               "rabbitmq-server units and restart them.")
```
dosaboy, line 22: Should we perhaps be clearer about that the fact that it is necessary to stop all units before starting any back up again i.e. a rolling restart will not work here.

dosaboy (commented):
looks good, quick question

core/plugins/rabbitmq.py
nicolasbock: :+1:

PR #159, opened 2021-10-13
Check for masked systemd Openstack services

### LINKED ISSUE #68: Check for masked systemd Openstack services
We should report if core services are masked:

e.g. sos_commands/systemd/systemctl_list-unit-files:neutron-server.service                       masked

nicolasbock:
Fixes #68


core/checks.py
```diff
@@ -796,6 +798,21 @@ def _get_running_services(self):
                         self.services[svc]["ps_cmds"].append(ret.group(0))
                         break
 
+    def _get_masked_services(self):
+        """Get masked services."""
+        masked = re.compile('^([^\\s]+)\\s*masked')
+        core_services = [
+            'nova',
+            'neutron'
```
dosaboy, line 806: Any reason not to use https://github.com/dosaboy/hotsos/blob/ac566dc176ba5fb2a1098b4a21e01f2c555bacc6/core/plugins/openstack/__init__.py#L186?

That way we can easily do this for any service without having to define statically here.

core/__init__.py
```diff
@@ -1,5 +1,26 @@
 import os
 
+AGENT_DAEMON_NAMES = {
```
dosaboy, line 3: why has this been moved here? this belongs in its original place in core.plugins.openstack since it is plugin specific code.

dosaboy: Oh now i see why its cause you are implementing your code in core.checks. You should avoid putting plugin-specific code under core (outside of core.plugins). My advice would be to first implement your change in the openstack plugin itself then think about what elements of your implementation can be generalised and moved into core.* if needs be.

PR #161, opened 2021-10-14
Alert if kernel is marked as tainted

slashdd:
Indicates if running kernel mark itself as 'tainted'
due to out-of-tree, proprietary, ... modules.

If kernel is tainted:
... CPU: XX PID: YYY Comm: migration/27 Tainted: G OX ...

If not tainted:
CPU: X PID: YYY Comm: khungtaskd Not tainted

Signed-off-by: Eric Desrochers <eric.desrochers@canonical.com>

### DIFF
--- a/core/issues/issue_types.py
+++ b/core/issues/issue_types.py
@@ -8,6 +8,10 @@ class KernelError(IssueTypeBase):
     pass
 
 
+class KernelWarning(IssueTypeBase):
+    pass
+
+
 class MemoryWarning(IssueTypeBase):
     pass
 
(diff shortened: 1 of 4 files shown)


slashdd:
> Is there perhaps another way to determine if the kernel is tainted other than looking in kern.log? e.g would uname -a show that too? If so i'd prefer to add that to core.plugins.kernel as a field in the summary e.g.:
> 
Yes, we can query the tainted state by reading `/proc/sys/kernel/tainted` instead. If that returns 0, the kernel is not tainted.

And I just checked, sosreport captures it as part of the kernel plugin:
https://github.com/sosreport/sos/blob/main/sos/report/plugins/kernel.py#L105

> kernel: version: 5.0.0-37-generic tainted: true boot: ro quiet splash vt.handoff=1 ...

I'm +1 on the `Tainted: true` proposal, instead of a `KernelWarning`.

dosaboy:
closing since it has not had updates in a while.

PR #180, opened 2021-10-28
[storage] OSD superblock read error

pponnuvel:
Indicates possibly a bad disk/sector that needs replacing.

### DIFF
--- a/defs/events.yaml
+++ b/defs/events.yaml
@@ -272,3 +272,6 @@ storage:
       heartbeat-no-reply:
         expr: '^([0-9-]+)\S* \S+ \S+ \S+ osd.[0-9]+ .+ heartbeat_check: no reply from [0-9.:]+ (osd.[0-9]+)'
         hint: 'heartbeat_check'
+      superblock-read-error:
+        expr: '.+ unable to read osd superblock'
+        hint: 'osd superblock'
--- a/plugins/storage/pyparts/ceph_event_checks.py
+++ b/plugins/storage/pyparts/ceph_event_checks.py
@@ -136,3 +136,10 @@ class CephDaemonLogChecks(CephEventChecksBase):
     @EVENTCALLBACKS.callback
     def heartbeat_no_reply(self, event):
         return self.get_timings(event.results)
+
+    @EVENTCALLBACKS.callback
+    def superblock_read_error(self, event):  # pylint: disable=W0613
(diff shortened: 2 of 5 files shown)


defs/events.yaml
```diff
@@ -272,3 +272,6 @@ storage:
       heartbeat-no-reply:
         expr: '^([0-9-]+)\S* \S+ \S+ \S+ osd.[0-9]+ .+ heartbeat_check: no reply from [0-9.:]+ (osd.[0-9]+)'
         hint: 'heartbeat_check'
+      superblock-read-error:
```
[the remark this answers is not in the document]
pponnuvel: Added a callback and a unit test

plugins/storage/pyparts/ceph_event_checks.py
```diff
@@ -136,3 +136,11 @@ def long_heartbeat_pings(self, event):
     @EVENTCALLBACKS.callback
     def heartbeat_no_reply(self, event):
         return self.get_timings(event.results)
+
+    @EVENTCALLBACKS.callback
+    def superblock_read_error(self, event):  # pylint: disable=W0613
+        msg = ('Detected superblock read errors detected which indicates '
```
dosaboy, line 142: repetition: "detected"

pponnuvel: Done

lathiat:
These errors can happen in other cases.

I would modify the error message not to assume it's a bad disk but suggest that they check the log for messages before the superblock fail to determine the cause of the failure to open the block device. While also keeping the note that checksum failures could be caused by disk corruption or failing disk / check SMART.

Specifically it can also happen if
- there is corruption due to a bug, there was a couple of those
- If permissions on the block device are wrong for some reason (theres a udev rule that changes them) and it can't open the device you get the same error about the superblock nto opening, but no checksum error.

pponnuvel:
Hi @lathiat,

Those two cases slightly different. Although the error messages look similar, the messages are
different (won't be caught by the filter here) and this is specifically about catching fault-but-working kind of situations.


In those two cases, the OSD won't start at all. So I thought it'be hard to miss that. What I was aiming to catch is that there are errors but it somehow starts (after restarting the OSD or the host, for example). If so, warn about it.

If you think the other 2 would be useful as well, I could add similar filters to catch them.

Issue #170, opened 2021-10-19
[CEPH] Summarize CRUSH rules and Pool/Rule relationships in output


KellenRenshaw:
All data can be found in "ceph report" output.

For quick analysis, it would be useful to extract the CRUSH rules and list pools using each rule in the rule entry.

Rules example from ceph report:
```json
        "rules": [
            {
                "rule_id": 0,
                "rule_name": "replicated_rule",
                "ruleset": 0,
                "type": 1,
                "min_size": 1,
                "max_size": 10,
                "steps": [
                    {
                        "op": "take",
                        "item": -1,
                        "item_name": "default"
                    },
                    {
                        "op": "chooseleaf_firstn",
                        "num": 0,
                        "type": "rack"
                    },
                    {
                        "op": "emit"
                    }
                ]
            },
            {
                "rule_id": 1,
                "rule_name": "flash-pool",
                "ruleset": 1,
                "type": 1,
                "min_size": 1,
                "max_size": 10,
                "steps": [
                    {
                        "op": "take",
                        "item": -2,
                        "item_name": "default~ssd"
                    },
                    {
                        "op": "chooseleaf_firstn",
                        "num": 0,
                        "type": "rack"
                    },
                    {
                        "op": "emit"
                    }
                ]
            },
            {
                "rule_id": 2,
                "rule_name": "hybrid-pool",
                "ruleset": 2,
                "type": 1,
                "min_size": 1,
                "max_size": 10,
                "steps": [
                    {
                        "op": "take",
                        "item": -30,
                        "item_name": "default~hdd"
                    },
                    {
                        "op": "chooseleaf_firstn",
                        "num": 0,
                        "type": "rack"
                    },
                    {
                        "op": "emit"
                    }
                ]
            }
        ],
```

The "rule_id" value is used in the pool's configuration as the value in the "crush_rule" key/value pair.

Mockup of potential output
```yaml
crush rules:
  "replicated_rule":
    id: 0
    type: replicated
    pools: [ cinder-ceph, cinder-ceph-coolstuff ]
  "flash-pool":
    id: 1
    type: replicated
    pools: [ cinder-ceph-vroom ]
  "hybrid-pool":
    id: 2
    type: replicated
    pools: [ ]
```

hillpd:
This would definitely come in handy. I'd recommend adding the pool id to the list for quick reference.

Example:
```yaml
crush rules:
  "replicated_rule":
    id: 0
    type: replicated
[...]

PR #182, opened 2021-10-30
[storage] Report crush rules and the pools using them

pponnuvel:
A summary of crush rules and the pools which use those
crush rules are reported as part of standard report. This
is useful to quickly see crush rule <-> pools mapping.

A test run: https://pastebin.canonical.com/p/TYzNKZdpvr/

Closes: #170

### DIFF
--- a/core/plugins/storage/ceph.py
+++ b/core/plugins/storage/ceph.py
@@ -41,6 +41,8 @@ CEPH_REL_INFO = {
         'jewel': '10.0'},
     }
 
+CEPH_POOL_TYPE = {1: 'replicated', 3: 'erasure-coded'}
+
 
 class CephConfig(checks.SectionalConfigBase):
     def __init__(self, *args, **kwargs):
@@ -371,6 +373,7 @@ class CephChecksBase(StorageBase):
         super().__init__(*args, **kwargs)
         self.ceph_config = CephConfig()
         self._bcache_info = []
+        self._crush_rules = []
         self._local_osds = None
         self._cluster_osds = None
         self.apt_check = checks.APTPackageChecksBase(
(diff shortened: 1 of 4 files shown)


dosaboy (commented):
Hi Pon, is it possible to get a unit test for this change?

core/plugins/storage/ceph.py
```diff
@@ -514,6 +517,36 @@ def local_osds(self):
...
+                pool_str = pool['pool_name'] + ' (' + str(pool['pool']) + ')'
+                res_pool.append(pool_str)
+
+        return res_pool
+
+    @property
+    def crush_rules(self):
+        """
+        Returns a list of crush rules, mapped to the respective pools.
+        """
```
dosaboy, line 533: Since this is a property we should return self._crush_rules if it is already set so that we dont do the same work more than once e.g.

if self._crush_rules:
    return self._crush_rules

core/plugins/storage/ceph.py
```diff
@@ -514,6 +517,36 @@ def local_osds(self):
         self._local_osds = self._get_local_osds()
         return self._local_osds
 
+    def _get_pools(self, pools=None, crush_rule=-1):
```
dosaboy, line 520: can we give this a more specific name like _filter_pools_by_rule() and also don't set pool=None since it always needs to be provided for this method to be useful.

dosaboy (changes requested):

core/plugins/storage/ceph.py
pponnuvel: done

core/plugins/storage/ceph.py
pponnuvel: done

pponnuvel:
> Hi Pon, is it possible to get a unit test for this change?

done now.

PR #186, opened 2021-11-03
Warn in case unattended upgrades are enabled

nicolasbock:
Closes: #188

### DIFF
--- a/plugins/system/pyparts/general.py
+++ b/plugins/system/pyparts/general.py
@@ -2,6 +2,10 @@ import re
 
 from core.cli_helpers import CLIHelper
 from core.plugins.system import SystemChecksBase
+from core.issues import (
+    issue_types,
+    issue_utils,
+)
 
 YAML_PRIORITY = 0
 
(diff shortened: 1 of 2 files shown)


plugins/system/pyparts/general.py
```diff
@@ -32,6 +36,9 @@ def get_system_info(self):
 
         if self.unattended_upgrades_enabled:
             self._output['unattended-upgrades'] = "ENABLED"
+            issue_utils.add_issue(issue_types.SystemWarning(
+                'Unattanded upgrades are enabled which can lead to ' +
+                'unexpected issues. Please disable.'))
```
pponnuvel, line 41: If we always recommend it be disabled, why have it at all?

[0] [1] [2] aren't that direct and appears to suggest it's good to have it enabled for security updates. Perhaps, we may need to be specific? I know we've had systems break because of unattended upgrades. Just playing devil's advocate ;-)

[0] https://canonical.my.salesforce.com/ka44K000000bwoZ
[1] https://wiki.debian.org/UnattendedUpgrades
[2] https://help.ubuntu.com/community/AutomaticSecurityUpdates

nicolasbock: Thanks @pponnuvel ! I have updated the message. Hopefully it's a bit more specific.

pponnuvel (approved):
Thanks, Nicolas. Looks much clearer now.

Couple of typos: unattanded -> unattended, maintainence -> maintenance

nicolasbock:
Thanks for the careful review @pponnuvel ! Those spelling issues are really quite embarrassing 😅

nicolasbock:
@pponnuvel and @dosaboy Could you give this another read? Thanks!

PR #194, opened 2021-11-08
Drop unnecessary `channel_id`

nicolasbock:


nicolasbock:
@dosaboy Note that the `notifications` workflow will fail because this is a fork and forks are not allowed to use repository secrets.

Issue #195, opened 2021-11-08
Warn if bcache-tuning charm isn't deployed


pponnuvel:
If bcache is configured for Ceph OSDs and the deployment is managed via Juju/charms, we recommend that that bcache-tuning charm [0] is deployed to tweak certain parameters from defaults. 

Issue a warning f `bcache-tuning` isn't deployed in such cases.

[0] https://jaas.ai/bcache-tuning/

PR #208, opened 2021-11-12
[openstack/storage] Test if bache-tuning charm is enabled

pponnuvel:
If there are OSD units deployed with bcache devices, we want
to check if the default bcache params are tweaked with
bcache-tuning charm.

Closes: #195

### DIFF
--- a/plugins/storage/pyparts/bcache.py
+++ b/plugins/storage/pyparts/bcache.py
@@ -1,9 +1,38 @@
 import re
 
 from core.cli_helpers import CLIHelper
-from core.plugins.storage.bcache import BcacheChecksBase
+from core.plugins.storage.bcache import BcacheBase, BcacheChecksBase
+from core.plugins.juju import JujuChecksBase
+from core.issues import issue_types, issue_utils
 
 YAML_PRIORITY = 3
+
+
+class BcacheCharmChecks(BcacheChecksBase):
+
+    def bcache_tuning_unit(self):
+        """
+        Check if bcache-tuning charm is deployed for Ceph with bcache.
+        """
+        if not BcacheBase.bcache_enabled:
+            return
+
+        likely_ceph = False
+        inst = JujuChecksBase()
+        for u in inst.units:
+            if u.name.startswith('ceph-osd'):
+                likely_ceph = True
+            if u.name.startswith('bcache-tuning'):
+                return
+
+        if likely_ceph:
+            msg = ("Detected bcache devices but found no bcache-tuning unit. "
+                   "It's recommended that bcache-tuning charm is deployed to "
+                   "configure params for optimal performance.")
+            issue_utils.add_issue(issue_types.BcacheWarning(msg))
+
+    def __call__(self):
+        self.bcache_tuning_unit()
 
 
 class BcacheCSetChecks(BcacheChecksBase):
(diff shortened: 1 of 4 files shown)


plugins/storage/pyparts/ceph_cluster_checks.py
```diff
@@ -27,8 +27,8 @@ def check_health_status(self):
             return
 
         if self.health_status != 'HEALTH_OK':
-            msg = ("Ceph cluster is '{}'. Please check 'ceph status' for "
-                   "details".format(self.health_status))
+            msg = ("Ceph cluster is in '{}' state. Please check 'ceph status' "
```
pponnuvel, line 30: Unrelated to #195; a grammar fix.

plugins/juju/pyparts/units.py
```diff
@@ -1,10 +1,35 @@
...
 
 YAML_PRIORITY = 2
 
 
 class JujuUnitChecks(JujuChecksBase):
 
+    def bcache_tuning_unit(self):
+        """
+        Check if bcache-tuning charm is deployed for Ceph with bcache.
+        """
```
dosaboy, line 16: Hi Pon, since this is a bcache check I feel it really belongs in plugins/storage/pyparts/bcache.py.

dosaboy (changes requested):

plugins/juju/pyparts/units.py
pponnuvel: Thanks. Done now.

lathiat:
It may be useful to also report on what the settings are tuned to, and check the bcache is currently set to them. In some cases such as bcache recreated after boot, the tuning isn't applied. As the current charm only runs on boot. Though not all sosreports will have bcache info.

pponnuvel:
> It may be useful to also report on what the settings are tuned to, and check the bcache is currently set to them. In some cases such as bcache recreated after boot, the tuning isn't applied. As the current charm only runs on boot.

Hi @lathiat, It's already being checked separately. See https://github.com/canonical/hotsos/blob/master/defs/config_checks.yaml#L40

It checks the common settings we have linked to bugs in the past. This, of course, doesn't include all tuning parameters. If you have any tuning parameters not checked currently (in config_checks.yaml and the bcache.py) that would be useful to monitor or potentially cause issues, I can include them.

> Though not all sosreports will have bcache info.

Yeah, this is taken care of by checking bcache devices: https://github.com/canonical/hotsos/blob/master/core/plugins/storage/bcache.py#L11

taodd (approved):

plugins/storage/pyparts/bcache.py
```diff
@@ -1,11 +1,39 @@
...
-from core.plugins.storage.bcache import BcacheChecksBase
+from core.plugins.storage.bcache import BcacheBase, BcacheChecksBase
+from core.plugins.juju import JujuChecksBase
+from core.issues import issue_types, issue_utils
 
 YAML_PRIORITY = 3
 
 
+class BcacheCharmChecks(BcacheBase, JujuChecksBase):
+
```
dosaboy, line 12: So it is kindof personal preference but I feel it makes for clearer code to avoid multiple inheritance where possible and so I would suggest leaving this as inheriting from BcacheChecksBase and then just instantiate a JujuChecksBase object inside the bcache_tuning_unit method.

dosaboy (changes requested):

plugins/storage/pyparts/bcache.py
pponnuvel: Agree that's a better way. Modified as suggested.

Issue #169, opened 2021-10-19
[CEPH] Highlight potential issues with CRUSH map/tree


KellenRenshaw:
Using the " ceph report" output, it would be useful to highlight CRUSH trees that exhibit imbalances at the configured failure domain level.

This can done by looking at the CRUSH rules (part of "ceph report"), identifying the failure domains used in the CRUSH rule (the "type" in the rule), then checking that buckets of that type in each tree have equal weights.

PR #212, opened 2021-11-15
[storage] Report CRUSH trees with imbalances

KellenRenshaw:
Shows CRUSH trees that are imbalanced using
the failure domains configured in the CRUSH rule(s)
that reference them.

Closes: canonical/hotsos#169

### DIFF
--- a/plugins/storage/pyparts/ceph_cluster_checks.py
+++ b/plugins/storage/pyparts/ceph_cluster_checks.py
@@ -336,6 +336,73 @@ class CephClusterChecks(CephChecksBase):
             issue = issue_types.CephCrushWarning(msg)
             issue_utils.add_issue(issue)
 
+    def _is_bucket_imbalanced(self, buckets, start_bucket_id, failure_domain):
+        """
+        Recursively determine if a tree is balanced at the given failure domain
+        """
+        unbalanced = False
+        weight = -1
+
+        for item in buckets[start_bucket_id]["items"]:
+            if buckets[item["id"]]["type_name"] != failure_domain:
+                unbalanced = self._is_bucket_imbalanced(buckets,
+                                                        item["id"],
+                                                        failure_domain)
+                if unbalanced:
+                    return unbalanced
+            else:
+                if weight == -1:
+                    weight = item["weight"]
+                else:
+                    if weight != item["weight"]:
+                        unbalanced = True
+                        return unbalanced
+
+        return unbalanced
+
+    def check_crushmap_equal_buckets(self):
+        """
+        Report when buckets of the failure domain type in a
+        CRUSH rule referenced tree are unbalanced.
+
+        Uses the trees and failure domains referenced in the
+        CRUSH rules, and checks that all buckets of the failure
+        domain type in the referenced tree are equal.
+        """
+        osd_crush_dump = self.cli.ceph_osd_crush_dump_json_decoded()
+        if not osd_crush_dump:
+            return
+
+        buckets = {b['id']: b for b in osd_crush_dump["buckets"]}
+
+        to_check = []
+
+        for rule in osd_crush_dump['rules']:
+            taken = 0
+            fdomain = 0
+            for i in rule['steps']:
+                if i["op"] == "take":
+                    taken = i["item"]
+                if "type" in i and taken != 0:
+                    fdomain = i["type"]
+                if taken != 0 and fdomain != 0:
+                    to_check.append((rule["rule_id"], taken, fdomain))
+                    taken = fdomain = 0
+
+        for ruleid, tree, failure_domain in to_check:
+            unbalanced = \
+                self._is_bucket_imbalanced(buckets, tree, failure_domain)
+            if unbalanced:
+                msg = ("unbalanced crush buckets identified in CRUSH "
+                       "root '{}' using failure domain '{}'. "
+                       "Affected CRUSH rule id is '{}'. "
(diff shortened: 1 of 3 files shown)


dosaboy (changes requested):
Thanks for the patch Kellen. I think it is good although I want to re-read it one more time to make sure I understand what it is doing. A couple of things that could be tweaked are;

 * doctstring could provide a bit more detail about how you are determining unbalanced buckets (as well as commit message)
 * the unit test data in CEPH_OSD_CRUSH_DUMP_FULL is kinda huge an you could put that in e.g. tests/unit/fake_data_root/sos_commands/ceph/json_output/ceph_osd_dump_--format_json-pretty.unbalanced_buckets then use a mock to pull it in e.g. like in https://github.com/canonical/hotsos/blob/56c30b4de8e658a626f480d0aa94e0d85a753d30/tests/unit/test_storage.py#L335

dosaboy:
Oh one more thing. can you rebase+squash your commits e.g. git rebase -i master

tests/unit/test_storage.py
```diff
@@ -332,6 +1290,16 @@ def test_superblock_read_error(self, mock_issue_utils):
...
+    @mock.patch.object(ceph_cluster_checks, 'issue_utils')
+    def test_check_crushmap_equal_buckets(self, mock_issue_utils):
+        with mock.patch.object(ceph_core, 'CLIHelper') as mock_helper:
+            mock_helper.return_value = mock.MagicMock()
+            mock_helper.return_value.ceph_osd_crush_dump_json_decoded.\
+                return_value = json.loads(CEPH_OSD_CRUSH_DUMP_FULL)
+            inst = ceph_cluster_checks.CephClusterChecks()
+            inst.check_crushmap_equal_buckets()
+            self.assertTrue(mock_issue_utils.add_issue.called)
+
```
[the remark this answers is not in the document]
KellenRenshaw: Done.

KellenRenshaw:
Rebased and added more description.

plugins/storage/pyparts/ceph_cluster_checks.py
```diff
@@ -336,6 +336,76 @@ def get_crushmap_mixed_buckets(self):
...
+        domain type in the referenced tree are equal.
+        """
+        osd_crush_dump = self.cli.ceph_osd_crush_dump_json_decoded()
+        if not osd_crush_dump:
+            return
+
+        buckets = {}
+        for bucket in osd_crush_dump["buckets"]:
+            bid = bucket["id"]
+            buckets[bid] = bucket
```
dosaboy, line 379: can replace the above with a one-liner e.g.

buckets = {b['id']: b for b in osd_crush_dump['buckets']}

plugins/storage/pyparts/ceph_cluster_checks.py
```diff
@@ -336,6 +336,76 @@ def get_crushmap_mixed_buckets(self):
...
+            taken = 0
+            fdomain = 0
+            for i in rule['steps']:
+                if i["op"] == "take":
+                    taken = i["item"]
+                if "type" in i and taken != 0:
+                    fdomain = i["type"]
+                if taken != 0 and fdomain != 0:
+                    to_check.append((rule["rule_id"], taken, fdomain))
+                    taken = fdomain = 0
```
dosaboy, line 393: i think the above should be outside of the "if" shouldn't it? i.e. you want to reset before/after each step (not just the ones where you find a match).

plugins/storage/pyparts/ceph_cluster_checks.py
dosaboy: also since you are not actually using buckets until after processing the rules you could move it to line 394

dosaboy (commented):
looks good, I've left a couple of minor change requests

plugins/storage/pyparts/ceph_cluster_checks.py
KellenRenshaw: Moved it to 390, don't want to rebuild the buckets each time through the loop.

plugins/storage/pyparts/ceph_cluster_checks.py
KellenRenshaw: I want to reset only after both `taken` and `fdomain` are populated. It is an admittedly naive implementation. Resetting on each step in a rule would result in no rules to process.

It is ugly because the tree and failure domain are in separate steps, that you can't be guaranteed are directly adjacent.

KellenRenshaw (commented):
Working on the fixes, should have them pushed momentarily.

tests/unit/test_storage.py
```diff
@@ -336,8 +336,37 @@ def test_superblock_read_error(self, mock_issue_utils):
...
+        self.assertFalse(mock_issue_utils.add_issue.called)
+
+    @mock.patch.object(ceph_cluster_checks, 'issue_utils')
+    def test_check_crushmap_non_equal_buckets(self, mock_issue_utils):
+        '''
+        Verifies that the check_crushmap_equal_buckets() function
+        correctly raises an issue against a known bad CRUSH map.
+        '''
+        test_data_path = ("sos_commands/ceph/json_output/"
+                          "ceph_osd_dump_--format_json-pretty_unbalanced")
```
dosaboy, line 352: the above path should be ceph_osd_crush_dump_ ... (not ceph_osd_dump). not a biggie, we can fix after this lands.

Issue #207, opened 2021-11-12
Add `--json` command line argument to format output in `json` format


nicolasbock:
E.g. for consumption with `jq`.

dosaboy:
for this task we can simply do a json.dump(yaml.load()) in python

PR #213, opened 2021-11-16
Add `json` option to output

nicolasbock:
This change adds a `json` option to the hotsos output. The default is still
`yaml`.

Closes: canonical/hotsos#207

### DIFF
--- a/hotsos.sh
+++ b/hotsos.sh
@@ -28,6 +28,8 @@ export DATA_ROOT
 # Plugin args - prefix must be plugin name
 export SHOW_CPU_PINNING_RESULTS=false
 export AGENT_ERROR_KEY_BY_TIME=false
+# Output format - default is yaml
+export OUTPUT_FORMAT
 # Path to the end product that plugins can see along the way.
 export MASTER_YAML_OUT
 export USE_ALL_LOGS=false
@@ -46,6 +48,7 @@ PROGRESS_PID=
 FULL_MODE_EXPLICIT=false
 MINIMAL_MODE=false
 USER_PROVIDED_SUMMARY=
+OUTPUT_FORMAT="yaml"
 MASTER_YAML_OUT=`mktemp`
 MASTER_YAML_OUT_ORIG=`mktemp`
 SAVE_OUTPUT=false
@@ -140,10 +143,10 @@ OPTIONS
     --user-summary
         Used in conjunction with --short to allow an existing (full) summary to
         to be shortened.
-
+    --json
+        Output in json format.
 
 PLUGIN OPTIONS
-
   These options only apply to specific plugins.
 
   openstack:
@@ -196,6 +199,9 @@ while (($#)); do
             ;;
         -s|--save)
             SAVE_OUTPUT=true
+            ;;
+        --json)
+            OUTPUT_FORMAT=json
             ;;
         -a|--all)
             PLUGINS[all]=true
@@ -332,9 +338,12 @@ generate_summary ()
     else
         repo_info=`get_git_rev_info` || repo_info="unknown"
     fi
-    echo "hotsos:" > $MASTER_YAML_OUT
-    echo "  version: ${SNAP_REVISION:-"development"}" >> $MASTER_YAML_OUT
-    echo "  repo-info: $repo_info" >> $MASTER_YAML_OUT
+
+    cat <<EOF > "${MASTER_YAML_OUT}"
+hotsos:
+  version: ${SNAP_REVISION:-"development"}
+  repo-info: $repo_info
+EOF
 
     for plugin in ${PLUGIN_NAMES[@]}; do
         # skip this since not a real plugin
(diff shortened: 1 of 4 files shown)


pponnuvel (approved):

dosaboy:
@nicolasbock i ran this piped into jq but it does not produce valid json and i think it's for two reasons. firstly you are doing to coversion to yaml per plugin which means that output has multiple root levels. It also seems to get curtailed probably because of the progress icon. A better place to the do conversion would be on $MASTER_YAML_OUT i.e. on the total output. I would therefore suggest that tools/output_filter.py is a better place to do the conversion.

nicolasbock:
I had overlooked the header that the shell script adds. I updated `hotsos.sh` accordingly. Now I can pipe the output through `jq` and get:

```json
{
  "hotsos": {
    "version": "development",
    "repo-info": "0813e1a"
  }
}
{
  "system": {
    "hostname": "rubberducky",
    "os": "ubuntu jammy",
    "num-cpus": 8,
    "load": "2.09, 2.16, 2.27",
    "rootfs": "/dev/mapper/crypt-root 940601196 440108772 452641320  50% /",
    "unattended-upgrades": "disabled",
    "date": "Wed Nov 17 04:26:57 PM UTC 2021",

    "sysctl-mismatch": {
      "net.ipv4.conf.all.rp_filter": {
        "actual": "0",
        "expected": "2"
      },
      "net.ipv4.conf.all.mc_forwarding": {
        "actual": "0",
        "expected": "1"
      }
    },
    "potential-issues": [
      {
        "desc": "host sysctl not consistent with contents of systcl.d",
        "origin": "system.checks",
        "type": "SysCtlWarning"
      }
    ]
  }
}
{
  "kernel": {
    "version": "5.13.0-7619-generic",
    "boot": "ro cryptdevice=/dev/nvme0n1p4:crypt-root:allow-discards rd.luks.options=discard quiet splash vt.handoff=7",
    "cpu": {
      "smt": "enabled"
    },
    "memory-checks": "no issues found",
    "potential-issues": [
      {
        "desc": "kern.log contains 1 stacktraces.",
        "origin": "kernel.log_event_checks",
        "type": "KernelError"
      }
    ]
  }
}
```

nicolasbock:
Another observation: The `output_filter.py` is called only in `MINIMAL_MODE`.

nicolasbock:
Please have another look @dosaboy. I have moved the conversion code into a separate helper script.

hotsos.sh
```diff
@@ -140,10 +142,14 @@ OPTIONS
     --user-summary
         Used in conjunction with --short to allow an existing (full) summary to
         to be shortened.
-
+    --format [yaml|json]
+        The format of the output. Defaults to yaml.
+    --yaml
+        Output in yaml format.
+    --json
+        Output in json format.
```
dosaboy, line 150: Lets just use --json, the rest are all redundant

hotsos.sh
```diff
@@ -332,9 +352,13 @@ generate_summary ()
...
-    echo "hotsos:" > $MASTER_YAML_OUT
-    echo "  version: ${SNAP_REVISION:-"development"}" >> $MASTER_YAML_OUT
-    echo "  repo-info: $repo_info" >> $MASTER_YAML_OUT
+
+    cat <<EOF >> "${MASTER_YAML_OUT}"
+---
+hotsos:
+  version: ${SNAP_REVISION:-"development"}
+  repo-info: $repo_info
+EOF
```
dosaboy, line 361: necessary? If so can we lose the ---

hotsos.sh
nicolasbock: In a way it is redundant since `yaml` is the default, that's true. But shouldn't the user be able to specify the default output format even if it's the default and that choice is redundant? This is a common pattern, see e.g. the openstack or the juju clients.

hotsos.sh
nicolasbock: Fixed (removed the `---`)

dosaboy: this is not the same as the original code as it uses >> as opposed to >. Is there a need for this change? If there isnt can we remove it?

.gitignore
```diff
@@ -6,3 +6,4 @@ snap/.snapcraft
 *.snap
 .tox
 *.pyc
+venv
```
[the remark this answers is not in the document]
nicolasbock: I use that to install Python dependencies. But tox should be sufficient for that. I'll remove it.

hotsos.sh
nicolasbock: Ok will remove.

hotsos.sh
nicolasbock: Oh you are right. Will fix in next iteration.

hotsos.sh
```diff
@@ -29,6 +29,7 @@ export DATA_ROOT
 export SHOW_CPU_PINNING_RESULTS=false
 export AGENT_ERROR_KEY_BY_TIME=false
 # Path to the end product that plugins can see along the way.
+export OUTPUT_FORMAT
```
dosaboy, line 32: eek sorry forgot one last thing. the comment above this applies to the line below so your line needs to go above the comment.

nicolasbock: Oops, sorry, I should have noticed that.

dosaboy (commented):
lgtm

PR #237, opened 2021-12-03
Documentation feedback

sombrafam:


dosaboy (commented):
Thanks for this feedback Erlon. I have put some time into addressing this and some other issues, including further reducing the list of overrides where possible, and have pushed those changes up. Please take a look at the updated [README](https://github.com/canonical/hotsos/blob/master/defs/README.md). Another key change is that the all.yaml approach no longer exists and now each directory can simply have a file with the same name that contains overrides for that directory/group. This is more inline with how it would be if it was all defined in the same file (as it used to be). Again I have documented this so let me know what you think.

dosaboy:
Gonna close this out since the docs have moved on a lot since this was opened. Thank you for opening it and feel free to rebase and re-open it if  for further feedback if you want.

Issue #245, opened 2021-12-21
include cpu scaling_governor info in summary


dosaboy:
It is often useful to know how the cpu scaling_governor is configured and if cpufreq_utils is installed and setup. For example an Openstack hypervisor would be recommended to use a scaling_governor set to "performance" e.g.

\# cat /sys/devices/system/cpu/cpu*/policy*/scaling_governor
performance
...

PR #246, opened 2021-12-21
[openstack] check scaling_governor is performance

dosaboy:
Adds cpu scaling_governor mode to kernel plugin
output and adds scenario check that performance
mode is used if host is used as a nova hypervisor.

Resolves: #245


aserdean:
LGTM

nicolasbock (approved):

Issue #247, opened 2021-12-28
Suppress nova bug report if fixed?


pponnuvel:
The following is reported by hotsos even if the node in question has been upgraded to the a nova package with the fixed version:
```
openstack:
- desc: known nova bug identified
id: https://bugs.launchpad.net/bugs/1888395
origin: openstack.auto_bug_check
```
This is because hotsos trawls through old/logrotated logs, too. 

I wonder if this should be suppressed by checking whether the package version has the fix.
But at the same time, I am not sure if displaying this might help explain that the system had the problem in the (recent) past.
So this is as much of a question as an "issue".

PR #248, opened 2022-01-03
[openstack] skip bug check if package has fix

dosaboy:
Adds support to bug handler for bug def to specify a package
version range that contains the fix such that the check can
be skipped if the installed package version has the fix.

Resolves: #247

### DIFF
--- a/core/ycheck/bugs.py
+++ b/core/ycheck/bugs.py
@@ -1,5 +1,6 @@
 from core import constants
 from core.log import log
+from core.checks import DPKGVersionCompare
 from core.known_bugs_utils import (
     add_known_bug,
     BugSearchDef,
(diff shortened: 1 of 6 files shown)


lathiat:
A not entirely uncommon problem is that a package was upgraded but a service is not yet running the new version. I'm not sure how to handle that cleanly in all cases but there definitely have been support cases where that is true. For example libc6 was upgraded but openvswitch wasn't restarted to use the new version. 

Same with ceph as services are not restarted by default.. we may need to account for indicating explicitly cases where this is likely to be an issue. One possible way would be to generally alert when a service is not running the latest packaged version (this can be determined from lsof and ldd and such for C programs, not sure its so easy for python)

dosaboy:
@lathiat thats a good idea for a feature request. I've opened https://github.com/canonical/hotsos/issues/249 for this.

Issue #133, opened 2021-09-24
Add /proc/net/{snmp, netstat, softnet_stat} as sources


nivseg2:
To add a few more network checks, including implementing previous
issue #113, we need to check items from the following sources:

1. /proc/net/snmp
2. /proc/net/netstat
3. /proc/net/softnet_stat
    (The /proc/ directory is usually captured in default sosreport commands,
      but might not be. If not, sosreport uses the output of 'netstat -s' to capture
      the first two). 
4. output of 'tc -s qdisc show'
    (which sosreport saves as sos_commands/networking/tc_-s_qdisc_show)

If it helps design of hotsos to have these be implemented in helper infrastructure
can implement them separately or we can just use directly in the networking checks.
Feel free to kill this request if so.

nivseg2:
If a system has up to 128 CPUs, this can get very large to look at. 
Ideally, we'd still retain the one line of data per CPU for readability.

Example script output:

```
total packets 17174510040
000: proc 2196343775 12.79% drop      0 tm_sq 2 rx_rps 0 fl_lim 0
001: proc 1077756705  6.28% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
002: proc  847475188  4.93% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
003: proc  437068556  2.54% drop      0 tm_sq 9 rx_rps 0 fl_lim 0
004: proc  811068775  4.72% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
005: proc 1004788831  5.85% drop      0 tm_sq 3 rx_rps 0 fl_lim 0
006: proc  744461698  4.33% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
007: proc 1575548651  9.17% drop      0 tm_sq 6 rx_rps 0 fl_lim 0
008: proc  635717160  3.70% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
009: proc  322843259  1.88% drop      0 tm_sq 6 rx_rps 0 fl_lim 0
010: proc  120422423  0.70% drop      0 tm_sq 3 rx_rps 0 fl_lim 0
011: proc 3280724640 19.10% drop 135318 tm_sq 5 rx_rps 0 fl_lim 0
012: proc 1463335306  8.52% drop      0 tm_sq 4 rx_rps 0 fl_lim 0
013: proc 1029967683  6.00% drop      0 tm_sq 0 rx_rps 0 fl_lim 0
[...]

PR #255, opened 2022-01-12
[kernel] Add new netstats checks

dosaboy:
Adds a new part to the kernel plugin that performs
checks on /proc/net/netstat and /proc/net/snmp.

Partially-Resolves: #133


aserdean:
Nice one!

I tested on my system:
```
kernel:
  version: 5.11.0-44-generic
  boot: ro quiet splash
  cpu:
    smt: enabled
    cpufreq-scaling-governor: unknown
  memory-checks: no issues found
  potential-issues:
    KernelWarnings:
      - TCP retransmission rate 61.64 exceeds 1%. (origin=kernel.net_stats)
      - Failed reverse path filter test (drops=1). (origin=kernel.net_stats)
```

It could use a unit test but beside that LGTM.

dosaboy:
I have review this with jay and we decided to land as is and tweak later if needs be.

PR #256, opened 2022-01-12
Add speed to NetworkPort

dosaboy:
When we display port info such as state it is useful to
also have the port link speed. This is now included in
NetworkPort.

### DIFF
--- a/tests/unit/test_host_helpers.py
+++ b/tests/unit/test_host_helpers.py
@@ -1,5 +1,8 @@
+import mock
+
 import utils
 
+from core.cli_helpers import CLIHelper
 from core.host_helpers import HostNetworkingHelper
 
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_host_helpers.py
```diff
@@ -42,7 +42,8 @@ def test_get_interface_with_addr_not_exists(self):
     def test_get_interface_with_addr_exists(self):
         expected = {'br-ens3': {'addresses': ['10.0.0.49'],
                                 'hwaddr': '52:54:00:e2:28:a3',
-                                'state': 'UP'}}
+                                'state': 'UP',
+                                'speed': 'unknown'}}
```
aserdean, line 46: Maybe it's worth having one test with an actual value eg: `speed: 10000Mb/s`

aserdean (approved):
Just one suggestion.
LGTM.

tests/unit/test_host_helpers.py
dosaboy: ack, ill add that

Issue #263, opened 2022-01-25
support more openflow protocols


dosaboy:
implement suggestion from https://github.com/canonical/hotsos/pull/262#issuecomment-1021233786

PR #269, opened 2022-01-26
Add more constant OpenFlow versions

aserdean:
Fixes: https://github.com/canonical/hotsos/issues/263

### DIFF
--- a/core/cli_helpers.py
+++ b/core/cli_helpers.py
@@ -291,7 +291,8 @@ class OVSDPCTLFileCmd(FileCmd):
 
 
 class OVSOFCTLBinCmd(BinCmd):
-    OFPROTOCOL_VERSIONS = ['OpenFlow15', 'OpenFlow14', 'OpenFlow13']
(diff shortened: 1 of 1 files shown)


core/cli_helpers.py
```diff
@@ -628,7 +629,11 @@ def command_catalog(self):
                  FileCmd('sos_commands/openvswitch/'
                          'ovs-ofctl_-O_OpenFlow13_show_{bridge}'),
                  FileCmd('sos_commands/openvswitch/'
-                         'ovs-ofctl_show_{bridge}')],
```
dosaboy, line 631: i think we need to keep this since it does exist in some sosreports, i have also never seen 10-13 in a sos but no harm in having it i guess.

dosaboy (changes requested):

core/cli_helpers.py
aserdean: Updated.
Indeed it a is slim possibility for it to use versions prior of OF 1.3 but it is still used in some usecases.

Issue #231, opened 2021-11-30
Recalibrate PGs per OSD check


pponnuvel:
We currently check whether the number of PGs per within a certain limit from 200. But this produces long list of "sub optimal" PGs/OSD warning - essentially fase positives. We need to widen the allowed range so that 50 - 200 (upstream recommendation) is considered OK.

https://docs.ceph.com/en/latest/rados/operations/placement-groups/#choosing-the-number-of-placement-groups

---

Also if PG autoscaling is enabled, we can disable this checl: https://docs.ceph.com/en/latest/rados/operations/placement-groups/#autoscaling-placement-groups

because autoscaler could be in progress and the "current" value - at the time of sosreport generation - may not be the actual one. Besides, autoscaler is supposed to adjust the PG count optimally.

pponnuvel:
https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/5/html/storage_strategies_guide/placement_groups_pgs

PR #271, opened 2022-01-26
[storage] Use optimal PG range

KellenRenshaw:
From Issue #231, adjusted the number of pgs check to reflect
an optimal range of 50-200 PGs per OSD, with 30% of margin.

Partially-Resolves: #231


KellenRenshaw:
Picked the easy bit of #231, not sure about a good way to handle disabling the check since the autoscaler has both global and per pool configuration.

nicolasbock (approved):
:+1:

pponnuvel (approved):
LGTM

Issue #258, opened 2022-01-20
check for "experiencing BlueFS spillover" in ceph


nkshirsagar:
as described in https://tracker.ceph.com/issues/38745 , and case 00326782, hotsos should check for message "BlueFS spillover detected " to check for the rocksdb's level-sized issue.

$ ceph -s
  cluster:
    id:     815ea021-7839-4a63-9dc1-14f8c5feecc6
    health: HEALTH_WARN
            BlueFS spillover detected on 1 OSD(s) <--

Also, even if this situation has not been hit, we should check between ceph-osd versions 15.2.6->15.2.10 if the option bluestore_volume_selection_policy is not set to "use_some_extra" because then the vulnerability still exists.

So this is the second check to be made, to check for the vulnerability if the ceph packages fall between 15.2.6 and 15.2.10,

$ sudo ceph daemon osd.1 config show|grep -i policy
"bluestore_volume_selection_policy": "use_some_extra",

dosaboy:
The following can be used as a guide on how to implement this:
https://github.com/canonical/hotsos/blob/master/defs/scenarios/storage/ceph/auth_insecure_global_id_reclaim_allowed.yaml

nkshirsagar:
There are two parts to this check.

a) simply check if we see BlueFS spillover detected in the ceph status, or the detailed status, and report the bug if that string is found.

b) Check between ceph-osd versions 15.2.6->15.2.10 to see if bluestore_volume_selection_policy is not set to use_some_extra and if that is the case, report the vulnerability.

I can see that a) is easily implemented using the approach shown in https://github.com/canonical/hotsos/blob/master/defs/scenarios/storage/ceph/auth_insecure_global_id_reclaim_allowed.yaml but I am not sure if the scenarios can also take min-version, fixed-version, etc like how the bugs do, and how I'd check for the value of bluestore_volume_selection_policy. Would I need to add a property to the class CephMon(CephDaemonBase) and return that value? just like we do for osdmaps_count?

```
class CephMon(CephDaemonBase):

    def __init__(self):
        super().__init__('mon')

    @property
    def osdmaps_count(self):
        report = self.cli.ceph_report_json_decoded()
        if not report:
            return 0

        try:
            return len(report['osdmap_manifest']['pinned_maps'])
        except (ValueError, KeyError):
            return 0
```

dosaboy:
[...]

PR #275, opened 2022-01-31
[ceph] Detect bluefs spillover. Also check bluestore_volume_selection_policy

nkshirsagar:
[ceph] Detect bluefs spillover. Also check bluestore_volume_selection_policy

Resolves: #258

### DIFF
--- a/defs/bugs/storage/ceph.yaml
+++ b/defs/bugs/storage/ceph.yaml
@@ -1,0 +1,18 @@
+1959649:
+  requires:
+    - apt:
+        ceph-osd:
+          - min: 15.2.6
+            max: 15.2.10
+    - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
+      value: [use_some_extra]
+      op: ne
+    - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
+      value: []
+      op: ne
+  raises:
+    message: >-
+      Vulnerability to a known bug https://tracker.ceph.com/issues/38745.
+      RocksDB needs more space than the leveled space available so it is using
+      storage from the data disk.
+      Please set bluestore_volume_selection_policy of all OSDs to use_some_extra
--- a/defs/scenarios/storage/ceph/bluefs_spillover.yaml
+++ b/defs/scenarios/storage/ceph/bluefs_spillover.yaml
@@ -1,0 +1,21 @@
+checks:
+  bluefs_spillover:
+    input:
+      command: ceph_health_detail_json_decoded
+    expr: '.+experiencing BlueFS spillover'
+  health_warning:
+    requires:
+      property: core.plugins.storage.ceph.CephChecksBase.health_status
+      value: HEALTH_WARN
+conclusions:
+  bluefs_spillover:
+    decision:
+      and:
+        - bluefs_spillover
+        - health_warning
+    raises:
+      type: core.issues.issue_types.CephWarning
+      message: >-
+        Known ceph bug https://tracker.ceph.com/issues/38745 detected.
+        RocksDB needs more space than the leveled space available.
+        See www.mail-archive.com/ceph-users@ceph.io/msg05782.html
--- a/tests/unit/test_storage.py
+++ b/tests/unit/test_storage.py
@@ -21,6 +21,11 @@ from plugins.storage.pyparts import (
     ceph_service_info,
 )
 
+JOURNALCTL_BLUEFS_SPILLOVER= """
+Dec 21 14:07:53 juju-1 mongod.37017[17873]: [replication-18] CollectionCloner ns:juju.txns.log finished cloning with status: QueryPlanKilled: PlanExecutor killed: CappedPositionLost: CollectionScan died due to position in capped collection being deleted. Last seen record id: RecordId(204021366)
+Dec 21 14:07:53 juju-1 mongod.37017[17873]: [replication-18] collection clone for 'juju.txns.log' failed due to QueryPlanKilled: While cloning collection 'juju.txns.log' there was an error 'PlanExecutor killed: CappedPositionLost: CollectionScan died due to position in capped collection being deleted. Last seen record id: RecordId(204021366)'
+"""  # noqa
+
 CEPH_CONF_NO_BLUESTORE = """
 [global]
 [osd]
@@ -68,6 +73,7 @@ CEPH_VERSIONS_MISMATCHED_MINOR = """
     }
 }
 """  # noqa
+
 
 CEPH_OSD_CRUSH_DUMP = """
 {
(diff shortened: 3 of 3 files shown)


defs/bugs/storage/ceph.yaml
```diff
@@ -0,0 +1,18 @@
...
+  requires:
+    and:
+      - apt:
+          ceph-osd:
+            - min: 15.2.6
+              max: 15.2.10
+    or:
+      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
+        value: [use_some_extra]
+        operator: ne
```
hemanthnakkina, line 11: As per documentation, the tag should be op instead of operator
https://github.com/canonical/hotsos/tree/master/defs#requires

defs/bugs/storage/ceph.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
+        value: []
+        operator: ne
```
hemanthnakkina, line 14: As per documentation, the tag should be op instead of operator
https://github.com/canonical/hotsos/tree/master/defs#requires

hemanthnakkina (changes requested):

defs/bugs/storage/ceph.yaml
nkshirsagar: I see what you mean. I think the sos I tested with doesn't have ceph daemon config info so an empty list was returned, and it was then flagging it because it was checking it as eq instead of ne.. since as you pointed out,  it defaults to eq if op is not found.. I will change it.. thanks!

nkshirsagar: I changed it to op but I still see the issue being flagged in the bugs. The sos I'm testing with does not contain that information,
~~~
root@nikhil-Lenovo-Legion-Y540-15IRH-PG0:/home/nikhil/Downloads/sosreport-juju-6879b7-3-lxd-1-00326782-2022-01-19-ylwpcfg# grep -ri bluestore_volume_selection_policy *
root@nikhil-Lenovo-Legion-Y540-15IRH-PG0:/home/nikhil/Downloads/sosreport-juju-6879b7-3-lxd-1-00326782-2022-01-19-ylwpcfg# 
~~~
So this check,
~~~
38745:
  requires:
    and:
      - apt:
          ceph-osd:
            - min: 15.2.6
              max: 15.2.10
    or:
      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
        value: [use_some_extra]
        op: ne
      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
        value: []
        op: ne
  raises:
    message: >-
      Vulnerability to a known bug https://tracker.ceph.com/issues/38745.
      Please set bluestore_volume_selection_policy of all OSDs to use_some_extra
~~~
should ideally not be hit because I'd expect value: [] for property check of core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy.. so "ne" should not be triggered.. 
But the bug is still hit, and then I realized that this check is wrong..

~~~
    or:
      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
        value: [use_some_extra]
        op: ne
      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
        value: []
        op: ne 
~~~
Since it's "or" the first one is true ..

So I changed the bug check to 

~~~
38745:
  requires:
    and:
      - apt:
          ceph-osd:
            - min: 15.2.6
              max: 15.2.10
      - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
        value: [use_some_extra]
        op: ne
    or:
      and:
        - apt:
            ceph-osd:
              - min: 15.2.6
                max: 15.2.10
        - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
          value: []
          op: ne
~~~

Then I get a proper result, i.e the scenario is hit due to the presense of the spillover string in the ceph status, but the bug is not hit due to the sos not containing the property at all (resulting in empty string)

~~~
  potential-issues:
    CephWarnings:
      - 'This node has Ceph OSDs running on it but is not using cpufreq scaling_governor
        in "performance" mode (actual=ondemand). This is not recommended and can result
        in performance degradation. To fix this you can install cpufrequtils and set
        "GOVERNOR=performance" in /etc/default/cpufrequtils. You will also need to
        disable the ondemand systemd service in order for changes to persist. NOTE:
        requires node reboot to take effect. (origin=storage.auto_scenario_check)'
      - Known ceph bug https://tracker.ceph.com/issues/38745 detected. (origin=storage.auto_scenario_check)
~~~

nkshirsagar:
This seems to be working fine,

~~~
38745:
  requires:
    and:
      - apt:
          ceph-osd:
            - min: 15.2.6
              max: 15.2.10
    and:
        - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
          value: [use_some_extra]
          op: ne
        - property: core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
          value: []
          op: ne
  raises:
    message: >-
      Vulnerability to a known bug https://tracker.ceph.com/issues/38745.
      Please set bluestore_volume_selection_policy of all OSDs to use_some_extra
~~~

nkshirsagar:
Hi @dosaboy , I've put in more changes along the lines of what was discussed yesterday. I've got the unit test for the scenario working (needed to change line #10 in https://pastebin.canonical.com/p/YW74SDGC33/ to CLIHelper instead of CephHelper)

I've also opened LP bug - https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1959649 though would the URL that hotsos generates be like this, or would it end up as https://bugs.launchpad.net/1959649 ? (non existent URL)

The other major issue is the test case for the bug check isn't working, it fails with 

~~~
======================================================================
FAIL: test_bug_checks (test_storage.TestStorageBugChecks)
----------------------------------------------------------------------
  File "/home/nikhil/HDD_MOUNT/Downloads/hotsos_fork/hotsos/.tox/py3/lib/python3.8/site-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/home/nikhil/HDD_MOUNT/Downloads/hotsos_fork/hotsos/tests/unit/test_storage.py", line 678, in test_bug_checks
    self.assertEqual(len(bugs), 1)
AssertionError: 0 != 1
-------------------- >> begin captured logging << --------------------
root: DEBUG: loading bugs definitions for plugin=storage
root: DEBUG: loaded plugin 'storage' bugs - sections=1, events=1
root: DEBUG: bug=1959649 path=None
root: DEBUG: no search terms registered so nothing to do
root: DEBUG: requirements provided as groups
root: DEBUG: op=and has 3 requirement(s)
root: DEBUG: installed=True
root: DEBUG: package ceph-osd=15.2.13-0ubuntu0.20.04.1 within version ranges [{'min': '15.2.6', 'max': '15.2.10'}] (result=False)
root: DEBUG: requirement check: apt ceph-osd (result=False)
root: DEBUG: calling property core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
root: DEBUG: requirement check: property core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy <built-in function ne> ['use_some_extra'] (result=True)
root: DEBUG: calling property core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
root: DEBUG: requirement check: property core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy <built-in function ne> [] (result=True)
root: DEBUG: bugcheck '1959649' requirement not met - skipping check
--------------------- >> end captured logging << ---------------------
~~~

I've tried it this way,

~~~
class TestStorageBugChecks(StorageTestsBase):

    @mock.patch('core.ycheck.CLIHelper')
    @mock.patch('core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs')
    @mock.patch('core.ycheck.bugs.add_known_bug')
    def test_bug_checks(self, mock_add_known_bug, mock_cephdaemon,
                        mock_helper):
        bugs = []

        def fake_add_bug(*args, **kwargs):
            bugs.append((args, kwargs))

        mock_add_known_bug.side_effect = fake_add_bug
        mock_helper.return_value = mock.MagicMock()
        mock_helper.return_value.ceph_versions.return_value = \
            CEPH_VERSIONS_BLUEFS_SPILLOVER.split('\n')
        mock_cephdaemon.return_value = mock.MagicMock()
        mock_cephdaemon.return_value.bluestore_volume_selection_policy = \
            ['rocksdb_original']
        YBugChecker()()
        # This will need modifying once we have some storage bugs defined
        self.assertFalse(mock_add_known_bug.called)
        self.assertEqual(len(bugs), 1)
~~~

dosaboy:
@nkshirsagar to answer one of your questions above, the LP bug url is created with https://github.com/canonical/hotsos/blob/fdf61f24ff4c4d426883b479c7b2c28a94d82ed4/core/known_bugs_utils.py#L43 so it will be like https://bugs.launchpad.net/bugs/1959649

nkshirsagar:
> @nkshirsagar to answer one of your questions above, the LP bug url is created with
> 
> https://github.com/canonical/hotsos/blob/fdf61f24ff4c4d426883b479c7b2c28a94d82ed4/core/known_bugs_utils.py#L43
> 
> so it will be like https://bugs.launchpad.net/bugs/1959649

ack thank you..

Issue #276, opened 2022-01-31
Multiple operators of same type are not properly handled in requires


hemanthnakkina:
In requires section, if you specify multiple operators of same type then its not properly handled.
The operators in require section work correctly only when the operators of each type are specified only once.

For example, below are 4 properties with 2 or operators. Lets say results of each property are R1, R2, R3, R4.
As per existing code, it is evaluated to OR( R1, R2, R3, R4)
The expectation is  AND( OR(R1, R2), OR(R3, R4) )

```
<bug id>:
  requires:
    or:
      - property: P1
        value: [V1]
        operator: ne
      - property: P2
        value: [V2]
        operator: ne
    or:
      - property: P3
        value: [V3]
        operator: ne
      - property: P4
        value: [V4]
        operator: ne
```

The operations are handled after all properties are evaluated instead of for each operation [1]. 
The should be handled per each operation.

[1] https://github.com/canonical/hotsos/blob/master/core/ycheck/__init__.py#L524-L534

dosaboy:
hi @hemanthnakkina there is a simple reason for this which is that the above translates to (and is processed as) a dict (see https://github.com/canonical/hotsos/blob/d1336635e7c6a32ab9fcf1672b8188fb95044824/core/ycheck/__init__.py#L504) and keys must therefore be unique which means that one of the two "or" values gets overwritten e.g.

requires:
  or: <foo>
  or: <bar>

would give {'or': 'bar'}

If we wanted to be able to specify the same operator more than once we would need to use a list but I did not envisage a use case. We can certainly add support for that if there is a need.

PR #277, opened 2022-01-31
Handle operations in requests within the group

hemanthnakkina:
Currently the operations in requests are handled
together for all teh groups. So if there are
multiple groups with same operator then results of
all the groups are applied with the operator.

This patch changes the logic to apply operations
for only the results of the group.

Resolves: #276

### DIFF
--- a/core/ycheck/__init__.py
+++ b/core/ycheck/__init__.py
@@ -499,9 +499,10 @@ class YPropertyRequires(YPropertyOverrideBase):
                 return False
         else:
             log.debug("requirements provided as groups")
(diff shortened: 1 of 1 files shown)


dosaboy (changes requested):
@hemanthnakkina im afraid this won't quite work as expected because both the code and the existing yaml definitions expect these operators to use a dictionary not a list. If we want to switch to using a list of operators we will need to modify all the existing defs to do that and we will also need to modify the code that processes them to expect them to be a list and not a dict.

Issue #285, opened 2022-02-21
[ovn] Add checks for OVN


hemanthnakkina:
Following checks can be added for OVN

1. Inactivity probes in ovsdb-server for both NB/SB

    2022-02-17T00:53:18.224Z|27572|reconnect|ERR|ssl:10.130.11.110:50350: no response to inactivity probe after 60 seconds, disconnecting

2. Unreasonably long polling time for ovsdb-server for NB/SB and for ovn-controller

2022-02-17T04:37:50.941Z|368641|timeval|WARN|Unreasonably long 27886ms poll interval (27242ms user, 616ms system)

3. Add WARN/ERR count for OVN logs - ovn-northd, ovsdb-server-nb, ovsdb-server-nb, ovn-controller

PR #286, opened 2022-02-22
[ovn] Add checks for OVN services

hemanthnakkina:
* Add the ovn-controller, ovn-northd, ovsdb-server-sb, ovsdb-server-nb
  logs to the error-and-warnings section
* Capture inactive probes from both ovsdb-server NB/SB logs
* Capture Unreasonably long poll interval occurences from
  ovn-controller and ovsdb-server NB/SB logs
* Fix ovn-controller-unreasonably-long-poll-interval regex

Resolves: #285


dosaboy:
lgtm thanks hemanth!

Issue #284, opened 2022-02-21
Not clear what action needs to be taken for `unbalanced crush buckets` warning


nobuto-m:
We've got the following warning from hotsos. However, the description wasn't crystal clear what needs to be checked in the crush map nor what action needs to be taken to remediate the situation (if it's not a false positive).

```
  potential-issues:
    CephCrushWarnings:
      - unbalanced crush buckets identified in CRUSH root 'default' using failure
        domain 'rack'. Affected CRUSH rule id is '0'. This can cause data distribution
        to become skewed - please check crush map (origin=storage.ceph_cluster_checks)
```

pponnuvel:
@nobuto-m Can you share `ceph osd crush dump -f json-pretty` where this happens?

nobuto-m:
Unfortunately this was an air-gapped deployment and we weren't able to spot the original sosreport stored somewhere, but we only have hotsos output.

KellenRenshaw:
I have been working on trying to be clearer in the output as to where one should look in the Ceph cluster's CRUSH map to determine if the check is alerting on something that does need remediation.

@nobuto-m 
Is this output easier to understand? It is for a test case configured for rack failure domains that are deliberately different to cause the check to be raised.
```
Found one or more unbalanced buckets in the cluster's CRUSH map.
This can cause uneven data distribution in the cluster. 
Verify that "ceph osd crush tree --show-shadow" conforms to the expected layout for the cluster.
Transient issues such as "out" OSDs, or cluster expansion/maintenance can trigger this warning.
Affected CRUSH tree(s) and bucket types are tree default at the rack buckets.
```

I don't currently see a way that this check can provide remediation information. That would require the check (and therefore hotsos) to know the desired state of the cluster.

We know that having all the buckets (at the configured failure domains) be the same size (or of size 0) is the optimal case, thus we alert on differences from that optimal case. However, the check has no way of knowing if those differences are intentional, transient, or the result of error.

For example, if the check finds that a set of buckets are of differing sizes, it doesn't have a way to know what the "correct" size should be.

nobuto-m:
A potential cause of triggering this warning in the specific deployment was 20 nodes were distributed into 3 racks as 7+7+6. Was that case to surface it from a code point of view? If so, we might want to see the actual number in the output so we can clearly know why it was triggered and can triage it appropriately.

PR #288, opened 2022-02-23
[storage] Clarify unbalanced check message

KellenRenshaw:
This changes the output of the unbalanced bucket check to be clearer
about the part of the tree that should be inspected.

Fixes: #284

### DIFF
--- a/defs/scenarios/storage/ceph/ceph-mon/crushmap_bucket_checks.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/crushmap_bucket_checks.yaml
@@ -24,10 +24,10 @@ conclusions:
     raises:
       type: CephCrushWarning
       message: >-
-        Found one or more unbalanced crush buckets e.g. crush root '{root}' using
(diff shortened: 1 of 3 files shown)


KellenRenshaw:
The failure seem unrelated to the changes that I made, happy to retest if needed.

pponnuvel:
I think the reworded message is clearer. Running against a customer issue we've discussed shows:
```
    CephCrushWarnings:
      - Found one or more unbalanced buckets in the cluster's CRUSH map. This can
        cause uneven data distribution or PG mapping issues in the cluster. Verify
        that "ceph osd crush tree --show-shadow" conforms to the expected layout for
        the cluster. Transient issues such as "out" OSDs, or cluster expansion/maintenance
        can trigger this warning. Affected CRUSH tree(s) and bucket types are tree
        default~nvme at the rack level. (origin=storage.auto_scenario_check)
```
P.S: `ceph osd crush tree --show-shadow` isn't part of sos data. So perhaps we may want to add that in sosreports.

Issue #287, opened 2022-02-23
add support for html formatted output


dosaboy:
This may be needed for applications that display their text as html.

Use https://docs.python.org/3/library/html.html

PR #289, opened 2022-02-24
Add support for html encoding output

dosaboy:
A new --html flag is provided to allow encoding the
output to html.

Resolves: #287


aserdean:
LGTM

PR #290, opened 2022-02-27
Expand messages raised for rabbitmq issues

nicolasbock:
OpenStack uses classic transient mirrored queues. In case a broker is lost and
re-added to the cluster, the mirrored queue will not be synchronized but will
receive new messages. Over time, the new broker's queue will end up being
synchronized once all messages from before have expired.

In case nodes are lost and re-added the following two scenarios are possible:

1. If the current primary broker is lost and there are no fully synchronized
   secondaries, RabbitMQ will stop the queue.

```
09:26:25.793 [error] Discarding message {'$gen_call',{<0.753.0>,#Ref<0.989368845.173015041.56949>},{info,[name,pid,slave_pids,synchronised_slave_pids]}} from <0.753.0> to <0.943.0> in an old incarnation (3) of this node (1)
```

2. When the brokers are back up after such a shut-down, RabbitMQ will not
   restart the queue and fail with

```
09:31:08.102 [error] Discarding message {'$gen_call',{<0.928.0>,#Ref<0.989368845.173015041.62682>},{info,[name,pid,slave_pids,synchronised_slave_pids]}} from <0.928.0> to <0.943.0> in an old incarnation (3) of this node (1)
```

Which is just another symptom of the same underlying problem.

In either case the solution is to shut down the cluster and restart it.

### DIFF
--- a/defs/scenarios/rabbitmq/cluster_logchecks.yaml
+++ b/defs/scenarios/rabbitmq/cluster_logchecks.yaml
@@ -35,5 +35,5 @@ conclusions:
       type: core.issues.issue_types.RabbitMQWarning
       message: >-
         Messages were discarded because transient mirrored classic queues are
-        not syncronized. Please stop all rabbitmq-server units and restart the
+        not synchronized. Please stop all rabbitmq-server units and restart the
         cluster. Note that a rolling restart will not work.


nicolasbock:
@dosaboy My current understanding is that this error is pretty serious as the queues that are affected by it are not usable. But I want to experiment a bit to test this hypothesis.

I will also have to actually update the error message :smile:

dosaboy:
For context, this mainly relates to the issues implemented for https://github.com/canonical/hotsos/issues/104 which has further context in upstream bug https://github.com/rabbitmq/rabbitmq-server/issues/2045. What i'd like to achieve here is a message that clearly explains why it thinks it has found the issue along with a reference to the upstream bug for more context. If we think it is appropriate we can also provide a bit more info on further debug and workaround actions but I know that the message is very limited format-wise so might be best to defer that to a KB article or something similar.

dosaboy:
Closing this for now since @nicolasbock confirmed no plan to progress. We can re-open later if we want.

Issue #293, opened 2022-03-04
Detect Octavia bug 2008099


hemanthnakkina:
Detect Octavia bug https://storyboard.openstack.org/#!/story/2008099

The LB failover does not work if session persistence is set on LB pool. This is fixed and available in Ussuri Octavia package but the environments with older versions of Octavia can hit this bug.

Error message that can be used for regex:
ERROR octavia.amphorae.drivers.haproxy.exceptions [req-517fc350-0d44-49d9-8c42-bf0ee08743cc - 77dd427b317c41bc92306aa341174958 - - -] Amphora agent returned unexpected result code 400 with response {'message': 'Invalid request', 'details': "[ALERT] 058/224306 (1748) : Proxy '2d4fb3e0-d743-48f6-a69a-c37e2c1246f2:8a692eba-7d40-4f8d-8a74-77d8c4f3dba8': unable to find local peer 'xFiQsE9fy1TMp7CDRHT-ClSZOEI' in peers section '3705f8d100144614a7475324946a3a5f_peers'.\n[WARNING] 058/224306 (1748) : Removing incomplete section 'peers 3705f8d100144614a7475324946a3a5f_peers' (no peer named 'xFiQsE9fy1TMp7CDRHT-ClSZOEI').\n[ALERT] 058/224306 (1748) : Fatal errors found in configuration.\n"}

PR #294, opened 2022-03-04
[openstack] detect octavia bug 2008099

hemanthnakkina:
Add bug check to detect bug
https://storyboard.openstack.org/#!/story/2008099

Resolves: #293

### DIFF
--- a/defs/bugs/openstack/octavia.yaml
+++ b/defs/bugs/openstack/octavia.yaml
@@ -1,4 +1,13 @@
 2008099:
+  requires:
+    apt:
+      octavia-common:
+        # ussuri
+        - min: 6.0.0
+          max: 6.2.0-0ubuntu1~cloud0~1
+        # victoria
+        - min: 7.0.0
+          max: 7.1.0-0ubuntu1~cloud0~1
   input:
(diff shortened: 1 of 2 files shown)


defs/bugs/openstack/octavia.yaml
```diff
@@ -0,0 +1,10 @@
+2008099:
+  input:
+    path: 'var/log/octavia/octavia-worker.log'
+  expr: '.+octavia.amphorae.drivers.haproxy.exceptions .+ Removing incomplete section ''peers'
+  hint: octavia.amphorae.drivers.haproxy.exceptions
+  raises:
```
dosaboy, line 6: Since the bug is fixed in mentioned package version can you add a version criteria like in https://github.com/canonical/hotsos/pull/294/files.

dosaboy (changes requested):

defs/bugs/openstack/octavia.yaml
hemanthnakkina: Done as new commit

pponnuvel: Squash them?

PR #297, opened 2022-03-16
[ceph] Detect OSD communication issues

pponnuvel:


pponnuvel:
OK, coverted to a scenario. But it fails with `undefined override 'osd_slow_heartbeats'`

PR #306, opened 2022-03-27
Add Pacemaker Plugin

### LINKED ISSUE #296: detect LP 1874719 - Use the hostname as the node name instead of hardcoded 'node1'
Check for [LP: 1874719] (https://bugs.launchpad.net/charm-hacluster/+bug/1874719).

dnegreira:
This aims to add a pacemaker plugin to hotsos.
It also adds a check to verify if a hostname with 'node1'
is configured on pacemaker, and presents a warning with the
correct action to remove it via juju.

Closes #296

### DIFF
--- a/client.py
+++ b/client.py
@@ -39,6 +39,8 @@ from plugin_extensions.storage import (
 )
 from plugin_extensions.vault.summary import VaultSummary
 
+from plugin_extensions.pacemaker.summary import PacemakerSummary
+
 
 class HotSOSSummary(plugintools.PluginPartBase):
     """
@@ -70,6 +72,7 @@ PLUGIN_RUN_ORDER = [
     'juju',
     'maas',
     'kernel',
+    'pacemaker',
 ]
 
 
@@ -160,6 +163,10 @@ PLUGIN_CATALOG = {'hotsos': {
                   'vault': {
                      'summary': {
                          'objects': [VaultSummary],
+                         'part_yaml_offset': 0}},
+                  'pacemaker': {
+                     'summary': {
+                         'objects': [PacemakerSummary],
                          'part_yaml_offset': 0}}}
 
 
--- a/core/plugins/pacemaker.py
+++ b/core/plugins/pacemaker.py
@@ -1,0 +1,55 @@
+import re
+
+from core import checks
+from core.plugintools import PluginPartBase
+from core.cli_helpers import CLIHelper
+
+PACEMAKER_PKGS_CORE = ['pacemaker', 'crmsh', 'corosync']
+PACEMAKER_SVC_EXPR = ['pacemaker[a-zA-Z-]*',
+                      'corosync']
+
+
+class PacemakerBase(object):
+
+    @property
+    def crm_status(self):
+        return CLIHelper().pacemaker_crm_status
+
+    @property
+    def offline_nodes(self):
+        crm_status = self.crm_status()
+        for line in crm_status:
+            regex_match = re.search(r'.*OFFLINE.*\[\s(.*)\s\]',
+                                    line)
+            if regex_match:
+                return regex_match.group(1).split()
+        return []
+
+    @property
+    def online_nodes(self):
+        crm_status = self.crm_status()
+        for line in crm_status:
+            regex_match = re.search(r'.*Online.*\[\s(.*)\s\]',
+                                    line)
+            if regex_match:
+                return regex_match.group(1).split()
+        return []
+
+
+class PacemakerChecksBase(PacemakerBase, PluginPartBase):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.apt_check = checks.APTPackageChecksBase(
+            core_pkgs=PACEMAKER_PKGS_CORE)
+        svc_exprs = PACEMAKER_SVC_EXPR
+        self.svc_check = checks.ServiceChecksBase(service_exprs=svc_exprs)
+        self.pacemaker = PacemakerBase()
+
+    @property
+    def apt_packages_all(self):
+        return self.apt_check.all_formatted
+
+    @property
+    def plugin_runnable(self):
+        return len(self.apt_check.core) > 0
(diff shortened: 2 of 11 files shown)


core/plugins/pacemaker.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @property
+    def offline_nodes(self):
+        offline_nodes = []
+        crm_status = self.crm_status()
+        for line in crm_status:
+            regex_match = re.search(r'.*OFFLINE.*\[\s(.*)\s\]',
+                                    line)
+            if regex_match:
+                offline_nodes = regex_match.group(1).split()
```
dosaboy, line 26: could just do:

return regex_match.group(1).split()

core/plugins/pacemaker.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class PacemakerBase(object):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.status = PacemakerStatus()
+
+
+class PacemakerChecksBase(PacemakerBase, PluginPartBase):
```
dosaboy, line 50: use this instead:

def __init__():
...
self.pacemaker = PacemakerBase()

core/plugins/pacemaker.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @property
+    def apt_packages_all(self):
+        return self.apt_check.all_formatted
+
+    @property
+    def plugin_runnable(self):
+        return len(self.apt_check.core) > 0
+
+    @property
```
dosaboy, line 67: There is no need to re-implement these since PacemakerBase() can be used directly from yaml defs.

defs/scenarios/pacemaker/node1_found.yaml
```diff
@@ -0,0 +1,13 @@
+checks:
+  node1-found:
+    requires:
+      property: core.plugins.pacemaker.PacemakerStatus.offline_nodes
+      op: contains
+      value: node1
```
dosaboy, line 6: nice!

defs/scenarios/pacemaker/node1_found.yaml
```diff
@@ -0,0 +1,13 @@
...
+conclusions:
+  node1-found-needs-removal:
+    decision: node1-found
+    raises:
+      type: core.issues.PacemakerWarning
+      message: >-
+       A node with the hostname node1 is currently configured and enabled on pacemaker. This is related to bug: https://bugs.launchpad.net/charm-hacluster/+bug/1874719. You can remove the node by running the following command on the application-hacluster leader: juju run-action <application>-hacluster/leader delete-node-from-ring node=node1 --wait
```
dosaboy, line 13: Can you wrap the line at char 79 like python code. Makes it much more readable (to be at least). The newlines wont appear in the output message.

client.py
```diff
@@ -70,97 +72,102 @@ def summary(self):
     'juju',
     'maas',
     'kernel',
+    'pacemaker',
 ]
 
 
 PLUGIN_CATALOG = {'hotsos': {
```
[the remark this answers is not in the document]
dnegreira: fixed on cab18285d94b5ececa730ae507fe39fc2d2da899

core/plugins/pacemaker.py
dnegreira: fixed on cab18285d94b5ececa730ae507fe39fc2d2da899

core/plugins/pacemaker.py
dnegreira: fixed on cab18285d94b5ececa730ae507fe39fc2d2da899

core/plugins/pacemaker.py
dnegreira: fixed on cab18285d94b5ececa730ae507fe39fc2d2da899

defs/scenarios/pacemaker/pacemaker_node1_found.yaml
```diff
@@ -0,0 +1,20 @@
+checks:
+  node1-found:
+    requires:
+      property: core.plugins.pacemaker.PacemakerBase.offline_nodes
+      op: contains
+      value: node1
```
dosaboy, line 6: I tried running this against a node that had pacemaker installed but not configured and it failed because offline_nodes returns None if there aren't any. Instead you should have it return [] so that this always works.

core/plugins/pacemaker.py
```diff
@@ -0,0 +1,55 @@
...
+    def offline_nodes(self):
+        crm_status = self.crm_status()
+        for line in crm_status:
+            regex_match = re.search(r'.*OFFLINE.*\[\s(.*)\s\]',
+                                    line)
+            if regex_match:
+                return regex_match.group(1).split()
+        return None
```
dosaboy, line 26: See my comment on your scenario but in short offline/online_nodes need to return [] if there aren't any so that operator.contains(nodes, <value>) does not fail.

tests/unit/test_pacemaker.py
```diff
@@ -0,0 +1,91 @@
...
+            mock_helper.return_value.pacemaker_crm_status.\
+                return_value = crm_status
+
+            mock_add_issue.side_effect = fake_add_issue
+            YScenarioChecker()()
+            self.assertTrue(mock_add_issue.called)
+            msg = (
+                "A node with the hostname node1 is currently configured and "
+                "enabled on pacemaker. This is related to bug:"
+                " https://bugs.launchpad.net/charm-hacluster/+bug/1874719."
```
dosaboy, line 84: nit: please put whitespace at the end of the line

dosaboy (changes requested):

defs/scenarios/pacemaker/pacemaker_node1_found.yaml
dnegreira: Fixed on c82b72d1a3be6ba00b6d2de421383f4c853ac395

core/plugins/pacemaker.py
dnegreira: Fixed on c82b72d1a3be6ba00b6d2de421383f4c853ac395

tests/unit/test_pacemaker.py
dnegreira: Fixed on c82b72d1a3be6ba00b6d2de421383f4c853ac395

Issue #206, opened 2021-11-12
Don't include openstack client packages in core package set


dosaboy:
The openstack plugin defines a core and non-core set of packages and uses the existence of core package as a means of deciding whether the plugin should be runnable. Currently a node that has only client packages qualifies for the openstack to run which is not what we want.

PR #307, opened 2022-03-28
[openstack] exclude client packages from core

dosaboy:
Don't include openstack client packages as part of the
core set if packages required to determine if a
service/project is installed since this leads to
false positives.

Resolves: #206


dnegreira:
* tested on a VM with two clients installed
* tested on a nova-compute node on openstack
*  tox working properly
LGTM

PR #311, opened 2022-04-07
Initial SSL Implementation

dnegreira:


defs/scenarios/openstack/openstack_apache2_certificates.yaml
```diff
@@ -0,0 +1,18 @@
+requires:
```
dosaboy, line 1: this will need to be a check  in itself since anything outside of checks: will be ignored.

hotsos/core/checks.py
```diff
@@ -766,3 +769,46 @@ def core(self):
...
+        try:
+            with open(self.certificate_path, "rb") as fd:
+                self.certificate = fd.read()
+        except IOError:
+            log.debug("Unable to read SSL certificate file %s",
+                      self.certificate_path)
+            raise
+
+    @property
+    def certificate_expire_date(self):
```
dosaboy, line 795: I think it might make sense to have a SSLCertificate class with these properties that way you can have more than one cert and have this class focus on checks rather than providing cert properties.

setup.py
```diff
@@ -9,6 +9,7 @@
         'pyyaml',
         'progress',
         'simplejson',
+        'cryptography',
```
dosaboy, line 12: also need to add this to requirements.txt

hotsos/core/checks.py
```diff
@@ -766,3 +766,19 @@ def core(self):
...
+
+    def __init__(self, certificate):
+        """
+        @param certificate: Expects an SSLCertificate object.
+        """
+        self.certificate = certificate
+
+    @property
+    def certificate_expires_soon(self):
+        "returns True if certificate expires in less than 60 days"
```
dosaboy, line 781: Can you use a normal doctring format i.e.

""" Returns True if certificate expires in less than 60 days  """

hotsos/core/ssl.py
```diff
@@ -0,0 +1,46 @@
+from hotsos.core.log import log
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+from datetime import datetime
+
+
+class SSLCertificate(object):
+
```
dosaboy, line 8: del whitespace

hotsos/core/ssl.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    This class instantiates an SSLCertificate object.
+    """
+
+    def __init__(self, certificate_path):
+        """
+        @param certificate_path: path to the certificate that we want to load
```
dosaboy, line 15: @param should go after the description

hotsos/core/ssl.py
```diff
@@ -0,0 +1,46 @@
...
+        in order to gather information from.
+        If the file in the path fails to read, it raises an exception.
+        """
+        self.certificate_path = certificate_path
+        try:
+            with open(self.certificate_path, "rb") as fd:
+                self.certificate = fd.read()
+        except OSError as e:
+            log.debug("Unable to read SSL certificate file %s, error: %s",
```
dosaboy, line 24: use log.error if you are going to raise the exception

we might want to just raise a log.warning and not raise the exception since this will mean the whole plugin fails

hotsos/core/ssl.py
```diff
@@ -0,0 +1,46 @@
...
+    @property
+    def certificate_expire_date(self):
+        "return datetime() of when the certificate expires"
+        cert = x509.load_pem_x509_certificate(self.certificate,
+                                              default_backend())
+        return cert.not_valid_after
+
+    @property
+    def certificate_days_to_expire(self):
+        "return int(days) remaining until the certificate expires"
```
dosaboy, line 37: Use a doctsring i.e. """ ...  """

hotsos/core/checks.py
```diff
@@ -766,3 +766,19 @@ def core(self):
...
+        if self.certificate.certificate_days_to_expire <= 60:
```
dosaboy, line 782: I think i'd prefer to have the expiry (60) as a constant/property and have the scenario call it from format-dict rather than statically set in the comment. that way if ever it changes we only have to change in one place.

hotsos/core/plugins/openstack/__init__.py
```diff
@@ -878,6 +879,60 @@ def release_name(self):
...
         return relname
 
+    @property
+    def apache2_ssl_config_file(self):
+        return os.path.join(HotSOSConfig.DATA_ROOT,
+                            'etc/apache2/sites-enabled',
+                            'openstack_https_frontend.conf')
+
+    @property
+    def ssl_enabled(self):
```
dosaboy, line 889: you could shorted this to:

return os.path.exists(self.apache2_ssl_config_file)

hotsos/core/ssl.py
```diff
@@ -0,0 +1,46 @@
...
+        today = datetime.today()
+        certificate_date = self.certificate_expire_date
+        days_to_expire = certificate_date - today
+        return int(days_to_expire.days)
+
+    @property
+    def path(self):
+        "return str(path) where the certificate file is located"
```
dosaboy, line 45: again docstring """ ... """

dosaboy (changes requested):

defs/scenarios/openstack/openstack_apache2_certificates.yaml
```diff
@@ -0,0 +1,22 @@
...
+      and:
+        - ssl_enabled
+        - apache2_certificate_expiring
+    raises:
+      type: hotsos.core.issues.OpenstackWarning
+      message: >-
+        The following certificates will expire in less than 60 days:
+        {apache2-certificates-path}
+      format-dict:
+        apache2-certificates-path: '@checks.apache2_certificate_expiring.requires.value_actual'
```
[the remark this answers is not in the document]
dnegreira: This is already fixed on 0d7456582fe0870ef0ead23011620a0686cbb292

hotsos/core/checks.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

hotsos/core/ssl.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

hotsos/core/ssl.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

hotsos/core/ssl.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

hotsos/core/plugins/openstack/__init__.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

hotsos/core/ssl.py
dnegreira: fixed on 59fc4895706a98108a6fc7c588544aceb54b1787

tests/unit/test_openstack.py
```diff
@@ -1150,3 +1389,63 @@ def test_pkgs_from_mixed_releases_found(self, mock_cli):
...
+            conf = "etc/apache2/sites-enabled/openstack_https_frontend.conf"
+            certificate_path = "etc/apache2/ssl/keystone/cert_10.5.100.2"
+            os.makedirs(os.path.dirname(os.path.join(dtmp, conf)))
+            os.makedirs(os.path.dirname(os.path.join(dtmp, certificate_path)))
+            with open(os.path.join(dtmp, conf), 'w') as fd:
+                fd.write(APACHE2_SSL_CONF)
+            with open(os.path.join(dtmp, certificate_path), 'w') as fd:
+                fd.write(CERTIFICATE_FILE)
+            base = openstack_core.OpenstackBase()
+            self.assertTrue(len(base.apache2_certificates_list) > 0)
```
dosaboy, line 1422: would this be more accurate?

self.assertEqual(len(base.apache2_certificates_list), 1)

tests/unit/test_openstack.py
```diff
@@ -1150,3 +1389,63 @@ def test_pkgs_from_mixed_releases_found(self, mock_cli):
...
+
+    def test_ssl_expiration_false(self):
+        # TODO we might need to change the assert here when the time comes :)
```
dosaboy, line 1425: We can't have a test whose state will change at some point in the future. I suggest that you just mock SSLCertificate.certificate_days_to_expire to return < 60 instead of relying on the contents of a real cert.

dosaboy (changes requested):

Issue #310, opened 2022-04-05
'potential_issues' and 'known_bugs' data would be helpful to have context metadata


ddstreet:
For issues, in addition to the actual human-readable text description, if there was metadata that included more context about the problem, it would be very helpful, especially to programmatic parsing, e.g. saucery.

This probably would be more useful to include the metadata only if a flag was passed, not by default, some flag like --full or something.

The specific context would include as much data as possible about:
- the filename(s), of course with full path inside the sosreport
  - if from a sos_command output, this could indicate what command group and specific subcommand instead of file path, if that's easier
- the specific line number(s) (per file, if there are multiple files)
- the 'seriousness' or severity of the issue/bug (the scale is arbitrary of course, but should be defined and consistent)
- some indicator to allow correlating multiple same-issue entries, i.e. if 5 occurrences of the same problem happened, each should be listed and each should be able to programmatically be identified as separate occurrences of the same issue (this might be simply nesting the occurrences under a top-level issue, but the top-level issue should be uniquely identifiable between separate sosreport analyses)
- reference (preferably a URI) to an external tracker/reference/etc for the problem - probably a bug report, in LP or github or whereever; it might be useful to standardize this so it's programmatically easy to identify a LP bug vs. github bug, etc (while consumers of this could parse the URI hostname, if this is provided as a URL, it would be better to standardize a field of this object to indicate specifically what type of external reference it is; then that standard field can be extended later to add more external reference types). It's possible there may be multiple external references for one issue, e.g. a bug link as well as a KB link

More context/metadata may be useful later, this is the main things I can think of now that would be useful.

dosaboy:
Leaving this open to allow further improvements to machine-readable issue context. The patch landed provides just basic context.

PR #314, opened 2022-04-07
Support machine readable output

dosaboy:
Resolves: #310


nicolasbock (approved):

dosaboy:
I have Partially-Resolves in the commit msg to indicate that this patch provides the basis for issue context but will leave open the issue as a means of refining the information we actually provide in --machine-readable context.

Issue #300, opened 2022-03-21
warn on ceph-mgr if "ceph osd df tree" is not showing up on monitor's sosreport


taodd:
When ceph-mgr is hang or down, the "ceph osd df tree" might be hanging and there wouldn't be an output file on the monitor's sosreport. 
So, we should warn the user that a restart of ceph-mgr might be required if we don't see the ceph_osd_df_tree file in the sosreport.

pponnuvel:
Plenty of (sos report) ceph commands can hang if ceph-mgr is experiencing problems or unreponsive. So we should probably expand this check widely (not just "ceph osd df tree").

dosaboy:
In the case of sosreport, would the outcome not be that the ceph plugin would timeout? In which case we would see this in https://github.com/canonical/hotsos/blob/c88375d8700bf53faed4e5de55c34bd0bdc66187/core/plugins/sosreport.py#L43

pponnuvel:
> In the case of sosreport, would the outcome not be that the ceph plugin would timeout? In which case we would see this in
> 
> https://github.com/canonical/hotsos/blob/c88375d8700bf53faed4e5de55c34bd0bdc66187/core/plugins/sosreport.py#L43

Yes, that's a strong indicator of unresponsive mgr.

It's also possible that the active one hangs (in response to a command) and a standby mgr becomes active frequently which might be able to process subsequent command(s). So if _any_ of the ceph_osd_* or ceph_pg_* files are empty on mon's sosreport, that's also a sign of problematic mgr (because we always expect these cmds to work and contain data when run from a mon node).

PR #322, opened 2022-04-12
[plugin/ceph] detect unresponsive ceph-mgr

pponnuvel:

Resolves: #300

### DIFF
--- a/defs/scenarios/storage/ceph/ceph-mon/unresponsive_mon_mgr.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/unresponsive_mon_mgr.yaml
@@ -1,0 +1,75 @@
+checks:
+  # In older sosreports, the plugin is called 'ceph' (= mon + mgr + osd) and
+  # in the newer ones it's called 'ceph_mon'.
+  sosreport_hung_ceph_mon_old:
+    requires:
+      property:
+        path: hotsos.core.plugins.sosreport.SOSReportChecksBase.timed_out_plugins
+        ops: [[contains, ceph]]
+  sosreport_hung_ceph_mon_new:
+    requires:
+      property:
+        path: hotsos.core.plugins.sosreport.SOSReportChecksBase.timed_out_plugins
+        ops: [[contains, ceph_mon]]
+  sosreport_hung_ceph_mgr:
+    requires:
+      property:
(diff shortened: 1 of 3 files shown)


defs/scenarios/storage/ceph/ceph-mon/unresponsive_mon_mgr.yaml
```diff
@@ -0,0 +1,45 @@
+checks:
+  # In older sosreports, the plugin is called 'ceph' (= mon + mgr + osd) and
+  # in the newer ones it's called 'ceph_mon'.
+  sosreport_hung_ceph_mon_old:
+    requires:
+      property:
+        path: hotsos.core.plugins.sosreport.SOSReportChecksBase.timed_out_plugins
+        ops: [[eq, ceph]]
```
dosaboy, line 8: timed_out_plugins is a list of plugins so you will need to do:

ops: [[contains, ceph]]

otherwise this will never match. same goes for the other checks.

tests/unit/storage/test_ceph_mon.py
```diff
@@ -864,3 +864,20 @@ def test_crushmap_bucket_checks_unbalanced_buckets(self, mock_helper):
                "check crush map.")
         issues = list(IssuesManager().load_issues().values())[0]
         self.assertEqual([issue['desc'] for issue in issues], [msg])
+
+    @mock.patch('hotsos.core.ycheck.YDefsLoader._is_def',
+                new=utils.is_def_filter('ceph-mon/unresponsive_mon_mgr.yaml'))
+    @mock.patch('hotsos.core.plugins.sosreport.SOSReportChecksBase.'
+                'timed_out_plugins', 'ceph_mon')
```
dosaboy, line 871: timed_out_plugins returns a list - https://github.com/canonical/hotsos/blob/main/hotsos/core/plugins/sosreport.py#L43

dosaboy (changes requested):

tests/unit/storage/test_ceph_mon.py
pponnuvel: done

defs/scenarios/storage/ceph/ceph-mon/unresponsive_mon_mgr.yaml
pponnuvel: done

Issue #326, opened 2022-04-19
snap failing to load python3-cryptography


dosaboy:
On dists other than Focal I get:

Traceback (most recent call last):
  File "/snap/hotsos/460/bin/hotsos", line 5, in <module>
    from hotsos.cli import main
  File "/snap/hotsos/460/lib/python3.8/site-packages/hotsos/cli.py", line 16, in <module>
    from hotsos.client import HotSOSClient, PLUGIN_CATALOG
  File "/snap/hotsos/460/lib/python3.8/site-packages/hotsos/client.py", line 11, in <module>
    from hotsos.plugin_extensions.openstack import (
  File "/snap/hotsos/460/lib/python3.8/site-packages/hotsos/plugin_extensions/openstack/summary.py", line 2, in <module>
    from hotsos.core.plugins.openstack import (
  File "/snap/hotsos/460/lib/python3.8/site-packages/hotsos/core/plugins/openstack/__init__.py", line 14, in <module>
    from hotsos.core.ssl import SSLCertificate
  File "/snap/hotsos/460/lib/python3.8/site-packages/hotsos/core/ssl.py", line 2, in <module>
    from cryptography import x509
  File "/snap/hotsos/460/lib/python3.8/site-packages/cryptography/x509/__init__.py", line 6, in <module>
    from cryptography.x509 import certificate_transparency
  File "/snap/hotsos/460/lib/python3.8/site-packages/cryptography/x509/certificate_transparency.py", line 10, in <module>
    from cryptography.hazmat.bindings._rust import x509 as rust_x509
ImportError: /lib/x86_64-linux-gnu/libpthread.so.0: version `GLIBC_PRIVATE' not found (required by /snap/core20/current/lib/x86_64-linux-gnu/librt.so.1)

I have played around with LD_PRELOAD/LD_LIBRARY_PATH to no avail so for now will have the code fail gracefully if the module fails to load to allow eveything else to run.

dosaboy:
re-opening and assigning to @dnegreira who said he would look into the real fix for this

nicolasbock:
Why is it importing from the host?
```
ImportError: /lib/x86_64-linux-gnu/libpthread.so.0: version `GLIBC_PRIVATE' not found (required by /snap/core20/current/lib/x86_64-linux-gnu/librt.so.1)
```
Or am I misreading this error message?

PR #328, opened 2022-04-19
Fix cryptography dependency for snap builds

dnegreira:
Since the python cryptography module>3.4, it is obligatory to
depend on the Rust compiler.
This brings further dependencies on other libraries for the snap
and did not allowed for hotsos to run correctly.

This patch aims to pin cryptography to 3.4, and introduce build
options to exclude the rust compiler dependency.

Closes: #326


requirements.txt
```diff
@@ -2,4 +2,4 @@ click
 progress
 pyyaml
 simplejson
-cryptography
+cryptography==3.4
```
nicolasbock, line 5: Why pin it at such an ancient version? Upstream is at 36.0.2. If there is a reason, could you include it in the commit message and requirements.txt?

dnegreira: Hi @nicolasbock, you can see it in the commit: https://github.com/canonical/hotsos/pull/328/commits/1bf39c8a7518c996535016c65d8f7394e9300243 and in the description of the PR as well.

Is there anything I am missing?

dnegreira: Oh, i'll add a comment on requirements.txt

nicolasbock: > Is there anything I am missing?

Yes, the why :)

This version might be great, I was just wondering why you pin it to a release that is from February 7 2021.

dnegreira: @nicolasbock fixed on https://github.com/canonical/hotsos/commit/2dcc79dd9a0acf20de1dd24ef48fdb60441f4275 LMK if it is ok like that or you want more info.
OTOH I would also like to be able to build the snap with `cryptography` with the greatest and latest, but this was the quick fix I could find in order for us to not lose sight on when a certificate is expiring. 
I can create a new issue to track this work. WDYT?

dnegreira: @nicolasbock fixed on https://github.com/canonical/hotsos/commit/2dcc79dd9a0acf20de1dd24ef48fdb60441f4275 LMK if it is ok like that or you want more info.
OTOH I would also like to be able to build the snap with `cryptography` with the greatest and latest, but this was the quick fix I could find in order for us to not lose sight on when a certificate is expiring. 
I can create a new issue to track this work. WDYT?

nicolasbock: Does this build with `3.4.8`? That's the last version in the `3.4.x` branch.

dnegreira: @nicolasbock good point, yes it does. I've just tested it and pushed the deps with 3.4.8 on https://github.com/canonical/hotsos/commit/b3c6ff910859269170c5997b7fe143fdb0d8fc4e

requirements.txt
```diff
@@ -2,4 +2,7 @@ click
 progress
 pyyaml
 simplejson
-cryptography
+# Pin to cryptography==3.4 since it is the last dependency version that
+# requires to be build with Rust and was breaking the snap.
+# More info on issue: https://github.com/canonical/hotsos/issues/326
```
nicolasbock, line 7:
```suggestion
# Pin to cryptography==3.4.8 since the cryptography-3.4.x branch is the
# last branch that can be built without the Rust toolchain..
# More info on issue: https://github.com/canonical/hotsos/issues/326
# [TODO] Bump cryptography to more recent release
```

setup.py
```diff
@@ -9,7 +9,10 @@
         'pyyaml',
         'progress',
         'simplejson',
-        'cryptography',
+        # Pin to cryptography==3.4 since it is the last dependency version that
+        # requires to be build with Rust and was breaking the snap.
+        # More info on issue: https://github.com/canonical/hotsos/issues/326
```
nicolasbock, line 14:
```suggestion
        # Pin to cryptography==3.4.8 since the cryptography-3.4.x branch is the
        # last branch that can be built without the Rust toolchain..
        # More info on issue: https://github.com/canonical/hotsos/issues/326
        # [TODO] Bump cryptography to more recent release
```

nicolasbock (changes requested):
Nit: A small change in the comment.

Otherwise this looks great!

requirements.txt
dnegreira: Fixed on https://github.com/canonical/hotsos/commit/985654fca5885b803c178de214a597569b76882b

setup.py
dnegreira: Fixed on https://github.com/canonical/hotsos/commit/985654fca5885b803c178de214a597569b76882b

nicolasbock (approved):
Thanks @dnegreira !

I can confirm that the snap works now, i.e. it doesn't die because of the `cryptography` package.

PR #333, opened 2022-04-21
Collect systemd service start time

dosaboy:
The ServiceChecksBase not supports retreiving the most
recent start time of a service and we also add support
to the Ydef requires.systemd property to check if
a service was started after another or not.

This is a pre-requisite for issue 331

### DIFF
--- a/defs/README.md
+++ b/defs/README.md
@@ -343,21 +343,29 @@ REQ_TYPE
     snap: package name
 
   SYSTEMD
-    Takes a systemd service and, optionally, state. Returns True if service
-    exists and, if provided, state matches. A short and long form are
-    supported as follows where the long form supports provided more than one
-    <service>: <state> pair and optionally an operator to be used instead of
-    the default 'eq' when comparing state.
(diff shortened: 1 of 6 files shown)


defs/README.md
```diff
@@ -358,6 +360,7 @@ REQ_TYPE
     systemd:
       <service name>: <service state>
```
hemanthnakkina, line 361: The sytanx mentioned here and in the example in test_ycheck.py  does not match.

    systemd:
        neutron-openvswitch-agent:
          state: enabled
          started-after: openvswitch-switch

hemanthnakkina (changes requested):
Please see comment inline. The syntax used for systemd in documentation and the exmaples+implementation didnot match. Example changed the syntax a bit to a dict under dict.

defs/README.md
dosaboy: nice catch, fixed.

hemanthnakkina:
LGTM!, lets wait for CI builds

hemanthnakkina (approved):

Issue #331, opened 2022-04-20
Check if openvswitch is restarted after neutron-openvswitch-agent


hemanthnakkina:
Restart of openvswitch without restart of neutron-openvswitch-agent can cause inconsistent l2pop flows [1].
This issue is not resolved upstream.

Need a check in hotsos if openvswitch is restarted after neutron-openvswitch-agent

[1] https://bugs.launchpad.net/neutron/+bug/1794991/comments/66

hemanthnakkina:
Following regex will be helpful in capturing the timestamp the openvswitch/neutron-openvswitch-agent are restarted from the file sosreport-<>/sos_commands/systemd/systemctl_status_--all

.* neutron-openvswitch-agent.service - Openstack Neutron .*? since (.*?);
.* ovs-vswitchd.service - Open vSwitch .*? since (.*?);

dosaboy:
This could be somewhat related to https://github.com/canonical/hotsos/issues/249

As discussed I think a fairly clean way to implement this kind of check would be to add a new "started-after" field to the [requires.systemd](https://github.com/canonical/hotsos/blob/e5b4fc0d100a27eaa4c49a66a7f9df3727fdcbe2/hotsos/core/ycheck/__init__.py#L1107) property so we can do e.g.

```
requires:
  systemd:
    neutron-openvswitch-agent:
      started-after: openvswitch-switch
```

dosaboy:
I have implemented the underlying functionality to implement this check here - https://github.com/canonical/hotsos/pull/333

PR #334, opened 2022-04-22
Add bug check for LP#1794991

hemanthnakkina:
Add check to determine if neutron-openvswitch-agent is
restarted after restart of openvswitch.

Resolves: #331

### DIFF
--- a/defs/scenarios/openstack/neutron/bugs.yaml
+++ b/defs/scenarios/openstack/neutron/bugs.yaml
@@ -2,6 +2,11 @@ checks:
   isolcpus_enabled:
     requires:
       property: hotsos.core.plugins.kernel.KernelBase.isolcpus_enabled
+  has_1794991:
+    requires:
+      systemd:
+        openvswitch-switch:
+          started-after: neutron-openvswitch-agent
   has_1883089:
     input:
       path: 'var/log/neutron/neutron-l3-agent.log'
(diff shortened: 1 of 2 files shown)


aserdean:
lgtm

defs/scenarios/openstack/neutron/bugs.yaml
```diff
@@ -47,6 +52,15 @@ checks:
       path: 'var/log/neutron/neutron-l3-agent.log'
     expr: '.+Gateway interface for router \S+ was not set up; router will not work properly'
 conclusions:
+  lp1794991:
+    decision: has_1794991
+    raises:
+      type: LaunchpadBug
+      bug-id: 1794991
+      message: >-
+        neutron-openvswitch-agent is not restarted after restart of
```
dosaboy, line 61: Suggested amendment to msg:

This host may be affected by a bug in Openstack Neutron ML2 whereby if the neutron-openvswitch-agent service is not restarted after a restart of openvswitch-switch this can lead to inconsistent l2pop flows. If you think this node is impacted you can fix this with a restart of neutron-openvswitch-agent.

hemanthnakkina: Modified the message and added unit test cases

dosaboy (approved):
lgtm thanks hemanth!

PR #336, opened 2022-04-22
Fix recursion in _is_bucket_imbalanced

KellenRenshaw:
The weight was not properly being passed between the recursions. This PR fixes that. Needs #288 to be applied first.


dosaboy:
@KellenRenshaw your other patch is now merged so you can go ahead and rebase this one

Issue #325, opened 2022-04-14
Add SSL checks for OVN


dnegreira:
Since OVN also uses SSL, we should check those certificates.

dosaboy:
lets revisit this because ... https://bugs.launchpad.net/charm-ovn-central/+bug/1979539

PR #338, opened 2022-05-02
OVN SSL Checks

dnegreira:
This patch aims to add SSL certificates validation for OVN.

It currently verifies the certificate name by the package name,
and adds information to the openvswitch summary in case ovn-host
or is ovn-central is installed.

Closes: #325

### DIFF
--- a/hotsos/core/plugins/openvswitch.py
+++ b/hotsos/core/plugins/openvswitch.py
@@ -1,7 +1,9 @@
 import re
+import os
 
 from hotsos.core import plugintools
 from hotsos.core import host_helpers
+from hotsos.core.config import HotSOSConfig
 from hotsos.core.ycheck.events import YEventCheckerBase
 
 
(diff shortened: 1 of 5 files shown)


hotsos/core/plugins/openvswitch.py
```diff
@@ -204,3 +206,65 @@ def get_results_stats(self, results, key_by_date=True):
...
+
+class OVNChecksBase(OpenvSwitchChecksBase, plugintools.PluginPartBase):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.ovn_package_names = ['ovn-central', 'ovn-host']
+        self.certificate_expire_days = 60
+
+    @property
+    def ovn_type(self):
```
dosaboy, line 219: Since this is used to determine if ovn is installed I think it would be better to replace that with a check like is ovn-common package installed and/or is ovn-host/ovn-northd etc systemd service enabled. That could be done entirely in the yaml check and therefore not require this python code. Also, ovn uses a number of certs for communication between different ovn services and we should be sure that we are checking them all.

hotsos/core/plugins/openvswitch.py
```diff
@@ -204,3 +206,65 @@ def get_results_stats(self, results, key_by_date=True):
...
+                try:
+                    ssl_certificate = host_helpers.SSLCertificate(
+                                      certificate_path)
+                    return ssl_certificate
+                except OSError:
+                    pass
+        return None
+
+    @property
+    def ssl_certificate_path(self):
```
dosaboy, line 252: Some ovn services may have more than one cert or some hosts may be running more than one service each with its own cert so this approach of assuming a single cert path wont work i think.

hotsos/core/plugins/openvswitch.py
```diff
@@ -204,3 +206,65 @@ def get_results_stats(self, results, key_by_date=True):
...
+        return self.ssl_certificate.path
+
+    @property
+    def ssl_enabled(self):
```
dosaboy, line 256: I think this is somewhat redundant in the context of ovn (it isnt a rest api like in openstack and it is always using ssl certs).

dosaboy (changes requested):

dosaboy:
Abandoning this pull request since it has not had any updates in months. Feel free to re-open and update whenever you want.

Issue #339, opened 2022-05-03
Unit tests failing on jammy


dnegreira:
Seems that `nose` [stopped working with jammy](https://paste.ubuntu.com/p/4yRVPjNBNc/).

Since `nose` [is now on maintenance mode for years](https://nose.readthedocs.io/en/latest/) and it breaks with python3.10, I believe we should be looking into other alternatives for unit testing.

Shall we go to `nose2` ? Any other idea/alternative?

dosaboy:
[stestr](https://github.com/mtreinish/stestr) might be a good alternative, it;s what openstack uses

PR #340, opened 2022-05-03
Switch to use stestr instead of nose

dosaboy:
Resolves: #339


dosaboy:
stestr supports running tests in parallel but we cant enable that yet because hotsosconfig is global and not threadsafe.

nicolasbock (approved):
LGTM

Issue #346, opened 2022-05-12
False positives for issue #331


hemanthnakkina:
Issue #331 is creating lot of false positives when openvswitch-switch is restarted some seconds after neutron-openvswitch-agent.
This can happen during restart of the node.

PR #347, opened 2022-05-12
Avoid service start check false positives.

dosaboy:
Checks for service start ordering can produce
false positives if they are started at almost
the same time e.g. on boot. This patch allows
a grace period of 120s i.e. service A must
have been started at least 120s after service
B in order for A started-after B to be True.

Resolves: #346


hemanthnakkina (approved):
LGTM

PR #348, opened 2022-05-13
Split ceph event defs into mon/osd

dosaboy:
This splits the event defs for storage.ceph into
storage.ceph.mon and storage.ceph.osd so that only
searches relevant to each are run and this is gated
on whether or not mon or osd are avaiable on that
node.


pponnuvel (approved):
LGTM

PR #350, opened 2022-05-19
Add checks for Openstack and Ceph EOL

### LINKED ISSUE #132: Report EOL of Openstack/Ceph releases
We can detect whether a particular version of Openstack/Ceph has been EOL'ed or going to be EOL'ed (Openstack has fixed dates, so we could report it ahead of EOL happening).

We probably need to report this as a warning, noting/differentiating the upstream EOL as well as Ubuntu/UCA EOLs.

Aim is inform users that they need to (or consider) upgrading to avoid going out of support and/or running into "old" bugs.

tpsilva:
Resolves #132

### DIFF
--- a/defs/scenarios/openstack/eol.yaml
+++ b/defs/scenarios/openstack/eol.yaml
@@ -1,0 +1,18 @@
+checks:
+  is_eol:
+    requires:
+      property:
+        path: hotsos.core.plugins.openstack.OpenstackBase.days_to_eol
+        ops: [[le, 0]]
+conclusions:
+  is-eol:
+    decision: is_eol
+    raises:
+      type: OpenstackWarning
+      message: >-
+        This node is running a version of Openstack that is
+        End of Life (release={release}) which means it has
+        limited support and is likely not receiving updates
+        anymore. Please consider upgrading to a newer release.
+      format-dict:
+        release: hotsos.core.plugins.openstack.OpenstackBase.release_name
--- a/defs/scenarios/storage/ceph/ceph-mon/eol.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/eol.yaml
@@ -1,0 +1,18 @@
+checks:
+  is_eol:
+    requires:
+      property:
+        path: hotsos.core.plugins.storage.ceph.CephChecksBase.days_to_eol
+        ops: [[le, 0]]
+conclusions:
+  is-eol:
+    decision: is_eol
+    raises:
+      type: CephWarning
+      message: >-
+        This node is running a version of Ceph that is
+        End of Life (release={release}) which means it
+        has limited support and is likely not receiving
+        updates anymore. Please consider upgrading to a newer release.
+      format-dict:
+        release: hotsos.core.plugins.storage.ceph.CephChecksBase.release_name
--- a/hotsos/core/plugins/openstack/common.py
+++ b/hotsos/core/plugins/openstack/common.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import os
 
 from hotsos.core import host_helpers
@@ -19,6 +20,20 @@ from hotsos.core.plugins.openstack.exceptions import (
     OVSDBAPP_EXCEPTIONS,
     MASAKARI_EXCEPTIONS,
 )
+
+# NOTE(tpsilva): when updating this, refer to the Charmed Openstack supported
+# versions page: https://ubuntu.com/openstack/docs/supported-versions
+OST_EOL_INFO = {
+    'xena': datetime(2023, 4, 30),
+    'wallaby': datetime(2024, 4, 30),
+    'victoria': datetime(2022, 4, 30),
+    'ussuri': datetime(2030, 4, 30),
+    'train': datetime(2021, 2, 28),
+    'stein': datetime(2022, 4, 30),
+    'rocky': datetime(2020, 2, 29),
+    'queens': datetime(2028, 4, 30),
+    'mitaka': datetime(2024, 4, 30)
+}
 
 
 OST_REL_INFO = {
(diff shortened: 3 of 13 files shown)


hotsos/core/plugins/openstack/common.py
```diff
@@ -19,6 +20,18 @@
     OVSDBAPP_EXCEPTIONS,
 )
 
+OST_EOL_INFO = {
+    'xena': datetime(2023, 4, 30),
```
pponnuvel, line 24: Include `yoga` as well?

hotsos/core/plugins/storage/ceph.py
```diff
@@ -31,6 +32,16 @@
 
 CEPH_LOGS = "var/log/ceph/"
 
+CEPH_EOL_INFO = {
+    'quincy': datetime(2032, 4, 30),
```
pponnuvel, line 36: Tthese are Ubuntu EOL dates, right?

ceph upstream has a much shorter cycle: https://docs.ceph.com/en/quincy/releases/index.html

I don't know if we (can) actually support that far out once upstream stopped accepting patches. So I am wondering if we need to use upstream dates. @hillpd comments?

tpsilva: That was actually something that came to mind while implementing this. Waiting for input from @hillpd or maybe @dosaboy to get some answers on this.

pponnuvel: Had a discussion in Ceph OH.

hillpd thinks we should use what our service description dates. But there's some complication there
as to what the right EOL date is for a Ceph pkg:

    If it's UCA/Ceph package then use UCA EOL date.
    If it's Ubuntu/Ceph package then use Ubuntu EOL date (which is what's currently set here in this PR).

dosaboy: @tpsilva @pponnuvel sorry for delayed response. I agree that using service description dates makes sense and since the majority of users will be using UCA for non-lts versions it makes sense to track UCA dates.

hotsos/plugin_extensions/openstack/summary.py
```diff
@@ -7,8 +15,24 @@ class OpenstackSummary(OpenstackServiceChecksBase):
...
 
+    def check_eol(self):
+        if self.release_name != 'unknown':
+            eol = OST_EOL_INFO[self.release_name]
+            today = datetime.now()
+            delta = (eol - today).days
+
+            if delta <= 0:
+                IssuesManager().add(
+                    OpenstackError("{} is EOL".format(self.release_name)))
```
dosaboy, line 30: Once this is moved to a scenario I think we should currently only raise an issue for releases that are already EOL (not ones that are about to be). This is is just to reduce noise. Using the summary info we would always be able to tell exactly when EOL will be if required.

dosaboy (changes requested):

defs/scenarios/openstack/eol.yaml
```diff
@@ -0,0 +1,12 @@
...
+    requires:
+      property:
+        path: hotsos.core.plugins.openstack.OpenstackBase.days_to_eol
+        ops: [[le, 0]]
+conclusions:
+  is-eol:
+    decision: is_eol
+    raises:
+      type: OpenstackWarning
+      message: This version of Openstack is EOL
```
dosaboy, line 12: try this:

message: >-
  This node is running a version of Openstack that is End of Life (release={release}) which means it has limited support and is likely not receiving updates anymore. Please consider upgrading to a newer release.
format-dict:
  release: hotsos.core.plugins.openstack.OpenstackBase.release_name

defs/scenarios/storage/ceph/ceph-mon/eol.yaml
```diff
@@ -0,0 +1,12 @@
...
+    requires:
+      property:
+        path: hotsos.core.plugins.storage.ceph.CephChecksBase.days_to_eol
+        ops: [[le, 0]]
+conclusions:
+  is-eol:
+    decision: is_eol
+    raises:
+      type: CephWarning
+      message: This version of Ceph is EOL
```
dosaboy, line 12: try this:

message: >-
This node is running a version of Ceph that is End of Life (release={release}) which means it has limited support and is likely not receiving updates anymore. Please consider upgrading to a newer release.
format-dict:
release: hotsos.core.plugins.storage.ceph.CephChecksBase.release_name

hotsos/core/plugins/openstack/common.py
```diff
@@ -20,6 +21,18 @@
     MASAKARI_EXCEPTIONS,
 )
 
+OST_EOL_INFO = {
```
dosaboy, line 24: If you could add a comment to say where you are getting these dates from that would be helpful for when we need to update them in the future.

hotsos/core/plugins/storage/ceph.py
```diff
@@ -31,6 +32,16 @@
 
 CEPH_LOGS = "var/log/ceph/"
 
+CEPH_EOL_INFO = {
```
dosaboy, line 35: If you could add a comment to say where you are getting these dates from that would be helpful for when we need to update them in the future.

hotsos/plugin_extensions/openstack/summary.py
```diff
@@ -10,27 +10,31 @@ def __summary_release(self):
         return self.release_name
 
     @idx(1)
+    def __summary_days_to_eol(self):
```
dosaboy, line 13: to keep ordering consistent across plugins can we keep this below services and dpkg (so basically put days_to_eol at 4)

dosaboy (changes requested):

dosaboy (approved):
lgtm thanks tiago

PR #352, opened 2022-05-31
Add ydefs path requirement type

dosaboy:
You can now test if an fs path exists or not when defining a
requires by doing e.g.

requires:
  path: foo

### DIFF
--- a/defs/README.md
+++ b/defs/README.md
@@ -165,8 +165,15 @@ format
 ```
 input:
   command: core.cli_helpers.CLIHelpers command
-  path: filesystem path (relative to DATA_ROOT)
+  path: FS_PATH
   options: OPTIONS
+
+or
+
+input: FS_PATH
+
+FS_PATH
+  This is a filesystem path that must be relative to DATA_ROOT.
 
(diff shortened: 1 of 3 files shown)


nkshirsagar:
tested this with a scenario like,

~~~
# This file is used to define overrides applicable to contents of this
# directory including subdirectories.
requires:
  property: hotsos.core.plugins.mysql.MySQLChecksBase.plugin_runnable

checks:
  max_connections_greater_than_4199:
    requires:
      config:
        handler: hotsos.core.plugins.mysql.MySQLConfig
        assertions:
          max_connections:
            ops: [[gt, 4190]]
  file_not_exists:
    requires:
      path: /etc/systemd/system/mysql.service.d/charm-nofile.conf

conclusions:
  max_connections_problem:
    decision:
      and:
        - max_connections_greater_than_4199
        - file_not_exists
    raises:
      type: MySQLWarning
      message: >-
        Max Connections is higher than 4190 but there is no
        charm-nofile.conf seen.
~~~
was able to build hotsos.

Issue #353, opened 2022-05-31
ceph report needs 'first_line_filter'


pponnuvel:
Some ceph commands have "extra" output in files collected by sosreport (due to stderr being captured) - which makes the json invalid.

We previously have a 'last_line_filter' to workaround it. But it's been observed that this can happen to be the first line as well.

The output was from a old Ceph (mimic), so it could be that what used to first line has now become last line. So this is likely to be an edge case. But a 'first_line_filter' can be implemented to fix it.

dosaboy:
@pponnuvel should we look for the same string at both ends or would it be something different? e.g. for ceph report we look for last_line_filter='report', should just add first_line_filter='report' as well?

PR #355, opened 2022-06-01
[ceph] Another filter for 'ceph report'

pponnuvel:
Older sosreports may have the extra 'report' line as
the first line of 'ceph report' which makes it an invalid
json file. Added a filter that removes the first line if
it starts with 'report'.

Fixes: #353

### DIFF
--- a/hotsos/core/host_helpers/cli.py
+++ b/hotsos/core/host_helpers/cli.py
@@ -379,13 +379,14 @@ class CephJSONFileCmd(FileCmd):
     end of the file (typically from stderr) which causes it to be invalid json
     so we have to strip that final line before decoding the contents.
     """
-    def __init__(self, *args, last_line_filter=None,
+    def __init__(self, *args, first_line_filter=None, last_line_filter=None,
                  **kwargs):  # pylint: disable=W0613
         super().__init__(*args, **kwargs)
-        if last_line_filter:
+        if first_line_filter or last_line_filter:
             self.register_hook('pre-exec', self.format_json_contents)
             self.register_hook('post-exec', self.cleanup)
             self.orig_path = None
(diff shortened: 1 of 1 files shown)


pponnuvel:
Noticed the same 'first line' <-> 'last line' swap in few other files too. So extended it a bit.

hotsos/core/host_helpers/cli.py
```diff
@@ -395,7 +396,12 @@ def format_json_contents(self, *args, **kwargs):  # pylint: disable=W0613
         with open(self.path) as f:
             lines = f.readlines()
 
-        if lines and lines[-1].startswith(self.last_line_filter):
+        if self.first_line_filter:
+            line_filter = self.first_line_filter
```
hemanthnakkina, line 400: Will it be simple to use python inbuilt function string strip() to handle this?
strip() can remove both leading and trailing characters.

hemanthnakkina (changes requested):
Please see inline comment

hotsos/core/host_helpers/cli.py
pponnuvel: I am not sure I understand.

Do you mean loading the entire file into a string and then removing the parts from either end and writing that string back?

What we try to do here is remove the first or last line which looks like in either form:
```
report xyz
{
 # json formatted output
}
```
or 
```
{
 # json formatted output
}
report xyz
```

The extra line "report xyz" makes invalid json and parsing fails. So removing that extra line makes it work.

hemanthnakkina:
@pponnuvel ignore my comment. I thought strip() will remove substrings but the function takes characters and strip the characters.  I am thinking if you can do substring and indexing to remove first or last line which will avoid creating temporary file.

PR #359, opened 2022-06-16
Add check for amd + iommu=pt

dosaboy:
Resolves: #319

### DIFF
--- a/defs/scenarios/kernel/amd_iommu_pt.yaml
+++ b/defs/scenarios/kernel/amd_iommu_pt.yaml
@@ -1,0 +1,23 @@
+checks:
+  cpu_vendor_is_amd:
+    property:
+      path: hotsos.core.plugins.kernel.sysfs.CPU.vendor
+      ops: [[eq, 'authenticamd']]
+  iommu_not_pt:
+    property:
+      path: hotsos.core.plugins.kernel.KernelBase.boot_parameters
+      ops: [[contains, 'iommu=pt'], [not_]]
+conclusions:
+  mixed-pkg-releases:
+    decision:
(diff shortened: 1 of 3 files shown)


tests/unit/test_system.py
```diff
@@ -156,3 +164,41 @@ def test_unattended_upgrades(self):
...
+    @mock.patch('hotsos.core.ycheck.engine.YDefsLoader._is_def',
+                new=utils.is_def_filter('amd_iommu_pt.yaml'))
+    def test_amd_iommu_pt_pass(self, mock_cli, mock_kernel):
+        mock_cli.return_value = mock.MagicMock()
+        mock_cli.return_value.lscpu.return_value = \
+            FAKE_AMD_LSCPU.splitlines(keepends=True)
+
+        mock_kernel.return_value = mock.MagicMock()
+        mock_kernel.return_value.boot_parameters = \
+            ['intel_iommu=on', 'iommu=pt']
```
gerald-yang-tw, line 200: does this mean both intel_iommu and iommu must be set to on and pt?
I am only sure for AMD EPYC series CPU, iommu must be set to pt, but not sure about intel_iommu

dosaboy: hi @gerald-yang-tw no only iommu=pt is needed:

  iommu_not_pt:
    property:
      path: hotsos.core.plugins.kernel.KernelBase.boot_parameters
      ops: [[contains, 'iommu=pt'], [not_]]

I just put both to test for false positives.

gerald-yang-tw: Got it, thanks Ed

gerald-yang-tw:
LGTM

Issue #356, opened 2022-06-14
Add check for bcache deadlock issue


dosaboy:
https://vanko.io/project/linux-block/patch/20220522170736.6582-5-colyli@suse.de/

nkshirsagar:
I think the only way to check this one would be to examine the stack trace.. there's no other way of realizing this has been hit..

dosaboy:
@nkshirsagar thats fair. Do you have an example stack trace for this issue? As a side topic we are working on https://github.com/canonical/hotsos/issues/367 which will hopefully provide an easy way to do this but we will need to ensure it exposes whatever information is needed by this check so an example would be helpful.

nkshirsagar:
https://pastebin.canonical.com/p/SN2JzgGgCq/ is what I see when I reproduce it using the customers DD dumps of the bcache devices that already are having this problem. We should ignore the function names which have ? in front of them. It's always bcache-register that will be blocked, so we search for that, and then the ones to search for in the fn traces are,

__schedule+0x24e/0x890
schedule+0x2c/0x80
closure_sync+0x18/0x90 [bcache]
bch_journal+0x123/0x380 [bcache]
bch_journal_meta+0x47/0x70 [bcache]
bch_btree_set_root+0x1c2/0x1f0 [bcache]
bch_btree_set_root+0x1c2/0x1f0 [bcache]
I still haven't been able to reproduce it myself from scratch to get the bcache devices into this state, so still working on that.

dosaboy:
ok so, now that we have https://github.com/canonical/hotsos/blob/f123587f344baea29051091dfecb5320515010d0/hotsos/core/plugins/kernel/kernlog.py#L339 I believe you could create e.g. a BcacheTraceHandler(TraceHandlerBase)

nkshirsagar:
thank you @dosaboy  - Does this look ok? I will also need to write out a unit test for this ..
 
```
diff --git a/defs/scenarios/kernel/kernlog_checks.yaml b/defs/scenarios/kernel/kernlog_checks.yaml
index 03419561..fa410260 100644
--- a/defs/scenarios/kernel/kernlog_checks.yaml
+++ b/defs/scenarios/kernel/kernlog_checks.yaml
@@ -9,6 +9,10 @@ checks:
     property:
       path: hotsos.core.plugins.kernel.CallTraceManager.oom_killer
       ops: [[length_hint]]
+  has_bcache_deadlock_invoked:
+    property:
[...]

PR #369, opened 2022-07-06
[bcache] Check if we see a deadlock during bcache registration

nkshirsagar:
Resolves: #356

### DIFF
--- a/defs/scenarios/kernel/kernlog_calltrace.yaml
+++ b/defs/scenarios/kernel/kernlog_calltrace.yaml
@@ -9,6 +9,10 @@ checks:
     property:
       path: hotsos.core.plugins.kernel.CallTraceManager.oom_killer
       ops: [[length_hint]]
+  has_bcache_deadlock_invoked:
+    property:
+      path: hotsos.core.plugins.kernel.CallTraceManager.calltrace-bcache
+      ops: [[length_hint]]
 conclusions:
   stacktraces:
     # Give this one lowest priority so that if any other call trace types match they
@@ -30,3 +34,15 @@ conclusions:
         {numreports} reports of oom-killer invoked in kern.log - please check.
       format-dict:
         numreports: '@checks.has_oom_killer_invoked.requires.value_actual:len'
+  bcache-deadlock-invoked:   
+    priority: 2
+    decision: has_bcache_deadlock_invoked
+    raises:
+      type: MemoryWarning
+      message: >-
+        Bcache cache set registration deadlock has occurred. This is caused
+        by a bug that has been fixed in kernel 5.15.11 (current is {kver}).
+        See https://www.spinics.net/lists/stable/msg566639.html for full
+        detail.
+      format-dict:
+        kver: hotsos.core.plugins.kernel.KernelBase.version
--- a/hotsos/core/plugins/kernel/kernlog/calltrace.py
+++ b/hotsos/core/plugins/kernel/kernlog/calltrace.py
@@ -206,6 +206,56 @@ class MemFieldsNodeZoneMem(MemFieldsBase):
                 'local_pcp', 'free_cma']
 
 
+class BcacheDeadlockType(TraceTypeBase):
+
+    def __init__(self):
+        self._search_def = None
+        self.deadlocks = []
+
+    @property
+    def name(self):
+        return 'calltrace-bcache'
+
+    @property
+    def searchdef(self):
+        if self._search_def:
+            return self._search_def
+
+        start = SearchDef(r'{}.+\s+task bcache-register:(\d+) blocked for '
+                          r'more than (\d+) seconds.'.
+                          format(KERNLOG_PREFIX))
+        body = SearchDef(r'(\S+) schedule(\S+) closure_sync(\S+)'
+                         r' bch_journal_meta(\S+) bch_btree_set_root(\S+)'
+                         r' bch_journal_replay(\S+) register_bcache(\S+)')
+        end = SearchDef(r'{}.+\s+do_syscall_(\S+)'.format(KERNLOG_PREFIX))
+        self._search_def = SequenceSearchDef(start, tag='bcache', body=body,
+                                             end=end)
+        return self._search_def
+
+    def apply(self, result):
+        """
+        Run through the results.
+        @param results: list of searchtools.SearchResult objects.
+        """
+        if len(result) > 0:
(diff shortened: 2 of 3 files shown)


hotsos/core/plugins/kernel/kernlog/calltrace.py
```diff
@@ -206,6 +206,58 @@ def fields(self):
                 'local_pcp', 'free_cma']
 
 
+class BcacheTraceType(TraceTypeBase):
+
+    def __init__(self):
+        self._search_def = None
+        self.generics = []
```
dosaboy, line 213: can you rename "generics" to something else e.g. "traces"

hotsos/core/plugins/kernel/kernlog/calltrace.py
```diff
@@ -206,6 +206,58 @@ def fields(self):
...
+    @property
+    def searchdef(self):
+        if self._search_def:
+            return self._search_def
+
+        start = SearchDef(r'{} (\S+) task bcache-register:(d+)'
+                          r' blocked for more than (d+) seconds.'.
+                          format(KERNLOG_PREFIX))
+        body = SearchDef(r'(\S+) schedule(\S+) closure_sync(\S+)'
+                         r' bch_journal_meta(\S+) bch_btree_set_root(\S+)'
```
dosaboy, line 228: Given that you are not extracting any information from results of this search is it possible to reduce the size of this expression? Remember that the longer the search expression the more CPU cycles it will take to run it and if you are storing items from the result this will also consume memory. Try to reduce the search to its absolute minimum.

hotsos/core/plugins/kernel/kernlog/calltrace.py
nkshirsagar: I've called it deadlocks

hotsos/core/plugins/kernel/kernlog/calltrace.py
nkshirsagar: I have reduced the search terms but I need to keep some in there to not run the chance of hitting some other bcache trace and confusing it for this deadlock

nkshirsagar:
I've written the unit test this way (please see the test_bcache_deadlock() function)

```
class TestKernelScenarioChecks(TestKernelBase):

    @mock.patch('hotsos.core.ycheck.engine.YDefsLoader._is_def',
                new=utils.is_def_filter('kernlog_calltrace.yaml'))
    def test_stacktraces(self):
        with tempfile.TemporaryDirectory() as dtmp:
            setup_config(DATA_ROOT=dtmp)
            os.makedirs(os.path.join(dtmp, 'var/log'))
            klog = os.path.join(dtmp, 'var/log/kern.log')
            with open(klog, 'w') as fd:
                fd.write(KERNLOG_STACKTRACE)

            YScenarioChecker()()

        msg = ('1 reports of stacktraces in kern.log - please check.')
        issues = list(IssuesStore().load().values())[0]

    @mock.patch('hotsos.core.ycheck.engine.YDefsLoader._is_def',
                new=utils.is_def_filter('kernlog_calltrace.yaml'))
    def test_bcache_deadlock(self):
        with tempfile.TemporaryDirectory() as dtmp:
            setup_config(DATA_ROOT=dtmp)
            os.makedirs(os.path.join(dtmp, 'var/log'))
            klog = os.path.join(dtmp, 'var/log/kern.log')
            with open(klog, 'w') as fd:
                fd.write(KERNLOG_BCACHE_DEADLOCK)

            YScenarioChecker()()

        msg = ('bcache register deadlock seen in kern.log - please check.')
        issues = list(IssuesStore().load().values())[0]

    @mock.patch('hotsos.core.ycheck.engine.YDefsLoader._is_def',
                new=utils.is_def_filter('kernlog_calltrace.yaml'))
    def test_oom_killer_invoked(self):
        YScenarioChecker()()
        msg = ('1 reports of oom-killer invoked in kern.log - please check.')
        issues = list(IssuesStore().load().values())[0]
~~~
But the test fails,

```
{0} tests.unit.test_kernel.TestKernelScenarioChecks.test_bcache_deadlock [0.022777s] ... FAILED

Captured traceback:
~~~~~~~~~~~~~~~~~~~

      File "/usr/lib/python3.8/unittest/mock.py", line 1325, in patched
    return func(*newargs, **newkeywargs)

      File "/home/nikhil/HDD_MOUNT/Downloads/my_new_hotsos/hotsos/tests/unit/test_kernel.py", line 288, in test_bcache_deadlock

      File "/usr/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)

      File "/usr/lib/python3.8/unittest/case.py", line 1118, in assertListEqual
    self.assertSequenceEqual(list1, list2, msg, seq_type=list)

      File "/usr/lib/python3.8/unittest/case.py", line 1100, in assertSequenceEqual
    self.fail(msg)

      File "/usr/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)

    AssertionError: Lists differ: ['1 reports of stacktraces in kern.log - please check.'] != ['bcache register deadlock seen in kern.log - please check.']

First differing element 0:
'1 reports of stacktraces in kern.log - please check.'
'bcache register deadlock seen in kern.log - please check.'

- ['1 reports of stacktraces in kern.log - please check.']
+ ['bcache register deadlock seen in kern.log - please check.']
```

File: defs/scenarios/kernel/kernlog_calltrace.yaml [hunk 3]
```diff
@@ -30,3 +34,12 @@ conclusions:
...
         numreports: '@checks.has_oom_killer_invoked.requires.value_actual:len'
+  bcache-deadlock-invoked:   
+    priority: 2
+    decision: has_bcache_deadlock_invoked
+    raises:
+      type: MemoryWarning
+      message: >-
+        bcache register deadlock seen in kern.log.
+        See https://www.spinics.net/lists/stable/msg566639.html.
+        This issue is fixed in kernel 5.15.51.
```
dosaboy, line 45: Could we update this message with the following:

message: >-
  Bcache cache set registration deadlock has occurred. This is caused by a bug that has been fixed
  in kernel 5.15.11 (current is {kver}). See https://www.spinics.net/lists/stable/msg566639.html
  for full detail.
format-dict:
  kver: hotsos.plugins.core.kernel.KernelBase.version

dosaboy (approved):
lgtm!

PR #371, opened 2022-07-12
Supporting arm64 snap

xtrusia:
Adding architecture arm64


lathiat (changes requested):
I wonder if we should use this instead
```
architectures:
  - build-on: [amd64, arm64]
    run-on: all
```

from: https://snapcraft.io/docs/architectures

(i feel like ideally you would build-on all but thats not documented as supported)

Issue #301, opened 2022-03-22
juju "unknown relation"


lathiat:
Juju “unknown relation” issue during hook. Not resolvable.

Two bugs, one fixed in 2.9.0 (which you can find on Google):
https://bugs.launchpad.net/juju/+bug/1895040

But this bug is also related, fixed in 2.9.13, not obvious from bug/pull request:
https://bugs.launchpad.net/charm-ceph-mon/+bug/1940983
https://github.com/juju/juju/pull/13286/files

juju upgrade (controller then model) is the fix

Example
```
2022-03-15 09:03:38 ERROR juju.worker.uniter agent.go:31 resolver loop error: unknown relation: 24
2022-03-15 09:03:38 INFO juju.worker.uniter uniter.go:309 unit "ams/0" shutting down: unknown relation: 24
2022-03-15 09:03:38 ERROR juju.worker.dependency engine.go:671 "uniter" manifold worker returned unexpected error: unknown relation: 24

```

PR #375, opened 2022-07-24
Check for juju unknown relation error

dosaboy:
Resolves: #301


lathiat (approved):
Looks good to me. I added a comment to https://bugs.launchpad.net/juju/+bug/1895040 clarifying that the same error was sometimes also fixed by commits for Bug #1940983 even though it's not documented as such. To help people understand why 2.9.13+ is needed not just 2.9.0+

PR #379, opened 2022-08-01
Add workflow for snap upload and release

tpsilva:


tpsilva:
@dosaboy for this to work, you'll need to add your snapcraft token to a repository secred called STORE_LOGIN. Descriptions on how to do this can be found at: https://github.com/snapcore/action-publish#store-login

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: snapcore/action-build@v1
+      id: build
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
+      with:
+        snap: ${{ steps.build.outputs.snap }}
+        release: stable
```
dosaboy, line 20: can we publish to edge instead of stable so that we get a chance to test a build before it gets released?

tpsilva: Done

nicolasbock:
Why is a separate action necessary to publish to snapcraft @tpsilva ? If the repository is connected to the snap it will build automatically every time there is a new commit on `main`. What am I missing?

Issue #243, opened 2021-12-14
Not available on non-x86_64 architectures (ARM64 etc)


lathiat:
Currently the snap cannot be installed on ARM64 hosts, such as an Ubuntu VM running on a new Apple Silicon Mac.

Launchpad can remote-build open source snaps for various architectures including ARM64. The requirements are documented here: https://snapcraft.io/docs/remote-build

See also: https://snapcraft.io/docs/architectures

We should enable all archictures (including armhf and arm64 but there is no reason not to enable the other architectures) and add launchpad remote-builds as part of the snap release process.

hotsos runs fine from a source directory on arm64.

lathiat:
It's not clear to me how the snap is currently released but if we are not already doing it, it probably makes the most sense to setup the automated Build from Github and then just merge in the architecture list.

If the Build from Github is already in use I can just submit a PR for the new architectures.

If it's not enabled then I can do the hard work there we just need to get permissions setup on the snap store for the wider team including me.

xtrusia:
only adding architectures: [amd64, arm64] and I let it build from github https://github.com/xtrusia/hotsos/blob/main/snap/snapcraft.yaml . i removed confinement field as test.. and hotsos-test is published successfully. not sure this will be working on official hotsos as well

dosaboy:
@xtrusia we now have automated builds so I believe we can try re-enabling arch: all to see if it will work

PR #382, opened 2022-08-03
Retry enabling multi-arch snap build

dosaboy:
Resolves: #243

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,6 +8,7 @@ description:
 confinement: classic
 grade: stable
 base: core20
+architectures: [amd64, arm64]
 environment:
   REPO_INFO_PATH: $SNAP/repo-info
   LC_ALL: C.UTF-8


nicolasbock (approved):

lathiat (changes requested):
I am fairly confident this wont work, as per:
https://snapcraft.io/docs/architectures

"run-on: all" only works for snaps that are purely architecture independant (e.g. python) which we are not, if for no other reason than the LD_PRELOAD hack (but we also seemed to have some other bits). We experienced this in the first attempt.

In order to build on all architectures, it seems you can simply delete the "architectures" entirely and according to the above URL

"Continuous Integration (CI) systems, such as [build.snapcraft.io](https://build.snapcraft.io/), can use the architectures keyword to determine which architectures to build a snap on. With none specified, a snap is built on all architectures."

The alternative would be to explicitly set:
architectures:

amd64
arm64
etc
But I would like to try just deleting it entirely first.

Unclear if a launchpad build does the same but I would have assumed so

dosaboy:
@lathiat fair enough, i'll make that change

PR #389, opened 2022-08-11
Adding mysql-router ssl issue

xtrusia:
Checking if client_ssl_cert and client_ssl_mode is set
properly

### DIFF
--- a/defs/scenarios/mysql/mysql_router.yaml
+++ b/defs/scenarios/mysql/mysql_router.yaml
@@ -1,0 +1,29 @@
+checks:
+  using_mysql_router:
+    apt: mysql-router
+  client_ssl_mode:
+    config:
+      handler: hotsos.core.plugins.mysql.MySQLRouterConfig
+      assertions:
+        key: client_ssl_mode
+        ops: [[eq, PREFERRED]]
+  client_ssl_cert:
+    config:
+      handler: hotsos.core.plugins.mysql.MySQLRouterConfig
+      assertions:
+        key: client_ssl_cert
+        ops: [[ne, null]]
+conclusions:
+  client_ssl_cert_problem:
+    decision:
+      and: [ using_mysql_router, client_ssl_mode ]
+      not: client_ssl_cert
+    raises:
+      type: LaunchpadBug
+      bug-id: 1959861
+      message: >-
+          This host is running MySQL Router and
+          has client_ssl_mode configured
+          but client_ssl_cert is not set.
+          This will cause mysql-router to error
+          when restarted.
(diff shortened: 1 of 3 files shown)


xtrusia:
in test code, I refered to DPKG_L_ROUTER but actually I need to check >= mysql router 8.0.23, Could you please advice anything to check version >= 8.0.23 ?

defs/scenarios/mysql/mysql_router.yaml
```diff
@@ -0,0 +1,26 @@
...
+      handler: hotsos.core.plugins.mysql.MySQLRouterConfig
+      assertions:
+        key: client_ssl_mode
+        ops: [[eq, PREFERRED]]
+  client_ssl_cert:
+    config:
+      handler: hotsos.core.plugins.mysql.MySQLRouterConfig
+      assertions:
+        key: client_ssl_cert
+        ops: [[ne, None]]
```
dosaboy, line 15: i think this needs to be:

[[ne, null]]

This is because Python None translates to null in yaml. if you use None in yaml it actually gives you string "None".

defs/scenarios/mysql/mysql_router.yaml
```diff
@@ -0,0 +1,26 @@
...
+conclusions:
+  client_ssl_cert_problem:
+    decision:
+      and: [ using_mysql_router, client_ssl_mode ]
+      not: client_ssl_cert
+    raises:
+      type: MySQLWarning
+      message: >-
+          MySQL Router has client_ssl_mode configuration but no client_ssl_cert
```
dosaboy, line 24: I'd like propose we work like this:

This host is running MySQL Router and has client_ssl_mode configured but
client_ssl_cert is not set. This will cause mysql-router to error when
restarted.

dosaboy: Also is there a launchpad bug for this issue? If there is we should add it here as well.

dosaboy (changes requested):

defs/scenarios/mysql/mysql_router.yaml
xtrusia: done

defs/scenarios/mysql/mysql_router.yaml
```diff
@@ -0,0 +1,29 @@
...
+      assertions:
+        key: client_ssl_cert
+        ops: [[ne, null]]
+conclusions:
+  client_ssl_cert_problem:
+    decision:
+      and: [ using_mysql_router, client_ssl_mode ]
+      not: client_ssl_cert
+    raises:
+      type: MySQLWarning
```
dosaboy, line 22: Ok so I should have spotted this in my last review but for the reason I have just raised in https://github.com/canonical/hotsos/issues/395 the type here needs to be "LaunchpadBug" and this will mean that in your unit test instead of calling IssuesManager().load_issues() you need to call IssuesManager().load_bugs().

dosaboy (changes requested):

dosaboy:
lgtm thanks seyeong!

Issue #393, opened 2022-08-15
bcache check produces false positive


pponnuvel:
bcache check incorrectly reports the result.
```
 potential-issues:
    BcacheWarnings:
      - bcache config writeback_percent expected to be ge 10 but actual=10. (origin=storage.auto_scenario_check)
```

When sequential_cutoff is 4.0M, it complains about a different writeback_percent (which is fine):
```
2022-08-15 12:17:53,364.364 625879 DEBUG None [-] run_collection:start (assertions)
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] ops=[['eq', '0.0k']], input=4.0M
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] op=eq, input=4.0M, expected=0.0k, force_expected=True
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] assertion: sequential_cutoff (4.0M) eq 0.0k result=False
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] 139923836819872: caching key=key with value=sequential_cutoff
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] 139923836819872: caching key=ops with value=eq 0.0k
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] 139923836819872: caching key=value_actual with value=4.0M
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] ops=[['eq', 'writethrough [writeback] writearound none']], input=writethrough [writeback] writearound none
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] op=eq, input=writethrough [writeback] writearound none, expected=writethrough [writeback] writearound none, force_expected=True
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] assertion: cache_mode (writethrough [writeback] writearound none) eq writethrough [writeback] writearound none result=True
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] 139923836819872: caching key=key with value=cache_mode
2022-08-15 12:17:53,365.365 625879 DEBUG None [-] 139923836819872: caching key=ops with value=eq writethrough [writeback] writearound none
2022-08-15 12:17:53,366.366 625879 DEBUG None [-] 139923836819872: caching key=value_actual with value=writethrough [writeback] writearound none
2022-08-15 12:17:53,366.366 625879 DEBUG None [-] ops=[['ge', 10]], input=10
[...]

PR #394, opened 2022-08-15
Fix config check issue message

dosaboy:
The config check property will store the last key/ops/actual_value
to be checked and the bcache store checks where always reporting the
last config checked when raising an issue which was not necessarily
the one that failed its assertion. This fixes that problem by
introducing a new cache key 'assertion_results' that contains all
assertions so that can be included in a message.

Resolves: #393

### DIFF
--- a/defs/scenarios/storage/bcache/cacheset.yaml
+++ b/defs/scenarios/storage/bcache/cacheset.yaml
@@ -30,11 +30,10 @@ conclusions:
     raises:
       type: BcacheWarning
       message: >-
-        bcache cacheset config {key} expected to be {op} but actual={value_actual}.
+        One or more of the following bcache cacheset config assertions failed:
+        {assertions}
       format-dict:
-        key: '@checks.cset_has_invalid_config.requires.key'
-        op: '@checks.cset_has_invalid_config.requires.ops'
-        value_actual: '@checks.cset_has_invalid_config.requires.value_actual'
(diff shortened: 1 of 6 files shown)


defs/scenarios/storage/bcache/cacheset.yaml
```diff
@@ -47,5 +46,4 @@ conclusions:
         bcache cache_available_percent is {actual} (i.e. approx. 30%) which
         implies this node could be suffering from bug LP 1900438 - please check.
       format-dict:
-        actual: '@checks.cset_config_lp1900438_limit.requires.value_actual'
-
+        actual: '@checks.cset_config_lp1900438_limit.requires.value_actual'
```
pponnuvel, line 49: Missing \n?

pponnuvel:
Works as expected now.
Sample test run:
```
    BcacheWarnings:
      - 'One or more of the following bcache bdev config assertions failed: sequential_cutoff
        eq "0.0k"/actual="4.0M", cache_mode eq "writethrough [writeback] writearound
        none"/actual="writethrough [writeback] writearound none", writeback_percent
        ge 10/actual="10" (origin=storage.auto_scenario_check)'
```

And a single op failure looks like:

```
    BcacheWarnings:
      - 'One or more of the following bcache bdev config assertions failed: sequential_cutoff
        eq "0.0k"/actual="4.0M" (origin=storage.auto_scenario_check)'
```

pponnuvel (approved):
LGTM

Issue #395, opened 2022-08-17
Update docs for raising known-bugs vs potential-issues


dosaboy:
If the [raises](https://github.com/canonical/hotsos/tree/main/defs#raises) property contains a ```bug-id:``` then it needs to use a ```type:``` that is implemented from BugTypeBase in order for it to be recognised as a bug rather than an issue (known-bugs vs. potential-issues). The documentation is a little unclear about this so needs updating. The code should also error if the conditions are not met since it will produce unexpected results.

PR #396, opened 2022-08-17
Ensure bug-id and bug type provided when raising bugs

dosaboy:
When raising a bug (known-bugs) both bug-id: and a bug
type: must be provided. A ScenarioException will now be
raised if this is not the case.

Resolves: #395


xtrusia:
LGTM, thanks Ed

xtrusia:
LGTM, thanks Ed

xtrusia (commented):
LGTM, thanks Ed

PR #401, opened 2022-08-21
[plugin/ceph] Recalibrate OSD metadata limit

### LINKED ISSUE #400: [ceph] false positives when checking metadata size larger than 10G
```
    - desc: Found 96 Ceph OSDs with metadata size larger than 10G. This could be the
        result of a compaction failure/bug and this host may be affected by https://tracker.ceph.com/issues/45903.
        A workaround (>= Nautilus) is to manually compact using 'ceph-bluestore-tool'.
```

This warning is generating a lot of false positives and needs to be re-evaluated.

The check is using a fixed 10GiB value and should probably be looking at metadata as a percentage of total utilization.

pponnuvel:
Modified the hardcoded value of OSD metadata limit from 10G to 5% of the actual OSD device size.

Fixes #400.

### DIFF
--- a/defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
@@ -11,10 +11,11 @@ conclusions:
       type: CephTrackerBug
       bug-id: 45903
       message: >-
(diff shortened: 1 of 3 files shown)


dosaboy:
lgtm

PR #403, opened 2022-08-21
[plugin/ceph] exclude unused crush rules from checks

### LINKED ISSUE #402: Exclude unused crush rules from balancing check
Sometimes there are crush rules that are not used by any of the pools. So they are not really unbalancing any tree(s).

pponnuvel:
If a crush rule isn't used by any of the pools, then it
doesn't need to be checked whether its tree is balanced.

Fixes #402.


KellenRenshaw:
LGTM, thanks for catching the oversight!

dosaboy (approved):
lgtm

PR #405, opened 2022-08-22
Cleanup ceph storage event unit tests

dosaboy:
The unit tests were running mon and osd events in one go in
test_ceph_osd. They now run osd and mon events separately in
test_ceph_osd and test_ceph_mon. Also splits non-mon logs from
fake_data_root/openstack/var/log/ceph/ceph-osd.log.1 and puts
them in fake_data_root/storage/ceph-mon.

Also ensures event pre-requisites are checked and enforced.


nkshirsagar (approved):
LGTM

PR #407, opened 2022-08-23
[plugin/ceph] Use the total used device size

pponnuvel:
The previous commit incorrectly used the total device
size instead of the total used size.

### DIFF
--- a/defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
@@ -11,11 +11,11 @@ conclusions:
       type: CephTrackerBug
       bug-id: 45903
       message: >-
-        Found {num_bad_meta_osds} Ceph OSDs with metadata usage > {limit_percent}% of its
-        total device usage. This could be the result of a compaction failure. Possibly related
+        Found OSDs {bad_meta_osds} with metadata usage > {limit_percent}% of its total
+        device usage. This could be the result of a compaction failure. Possibly related
         to the bug https://tracker.ceph.com/issues/45903 if Ceph < 14.2.17. To manually
         compact the metadata, use 'ceph-bluestore-tool' which is available since 14.2.0.
       format-dict:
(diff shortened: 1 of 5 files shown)


pponnuvel:
The functional test fails because the metadata usage is really high ( > 99.96% - the fake_data_root seems to be from a cluster no  data).  I'll leave until @dosaboy gets back as editing ceph_osd_df_tree_--format_json-pretty in fake_data_root may not be good idea. If not, we'd need another approach to handle the functest failure.

hillpd (approved):
Looks good!

defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
```diff
@@ -11,11 +11,11 @@ conclusions:
...
       bug-id: 45903
       message: >-
-        Found {num_bad_meta_osds} Ceph OSDs with metadata size > {limit_percent}% of its
-        device size. This could be the result of a compaction failure. Possibly related
+        Found OSDs {bad_meta_osds} OSDs with metadata usage > {limit_percent}% of its
+        total device usage. This could be the result of a compaction failure. Possibly related
         to the bug https://tracker.ceph.com/issues/45903 if Ceph < 14.2.17. To manually
         compact the metadata, use 'ceph-bluestore-tool' which is available since 14.2.0.
       format-dict:
-        num_bad_meta_osds: '@checks.bluefs_osds_have_oversize_metadata.requires.value_actual:len'
```
dosaboy, line 19: i think its undocumented but there you can use :comma_join instead of :len

defs/scenarios/storage/ceph/ceph-mon/bluefs_size.yaml
```diff
@@ -11,11 +11,11 @@ conclusions:
       type: CephTrackerBug
       bug-id: 45903
       message: >-
-        Found {num_bad_meta_osds} Ceph OSDs with metadata size > {limit_percent}% of its
```
dosaboy, line 14: The second "OSDs" can be removed i think.

dosaboy (changes requested):

pponnuvel:
Addressed the comments @dosaboy

PR #412, opened 2022-09-01
[plugin/ceph] Fine tune suboptimal PG count warning

### LINKED ISSUE #404: Don't warn about PGs/OSD being sub-optimal if pg autoscaler is enabled
Since Octopus, PG autoscaler module is enabled by default. So it can automatically adjust the PG count(s) if necessary.
This warning can also be a incorrect for new cluster with not much data and PGs will be increased as needed by autoscaler.

So I propose that if autoscaler is enabled and it doesn't warn about "too few" or "too many" PGs, we shouldn't issue a warning. Can still list the out-of-range PGs/OSD but no warning.

pponnuvel:
The suboptimal PGs per PSD warning can at times be false.
A specific case is when a cluster is newly deployed. In
that case, there wouldn't much data, and so, either the
admin hasn't scaled the PGs yet or if the PG autoscaler
is enabled, then it doesn't think PGs need to be scaled
for the current utilisation.

So we now warn on suboptimal PGs if:
  - autoscaler isn't enabled, and
  - *every* pool in the cluster is utilised under 2%

The suboptimal PGs will still be reported in the summary - just the warning won't issued now unless it meets the (above noted) criteria.

Fixes #404.

### DIFF
--- a/defs/scenarios/storage/ceph/ceph-mon/pg_imbalance.yaml
+++ b/defs/scenarios/storage/ceph/ceph-mon/pg_imbalance.yaml
@@ -9,6 +9,14 @@ checks:
       property:
         path: hotsos.core.plugins.storage.ceph.CephCluster.osds_pgs_suboptimal
         ops: [[length_hint], [gt, 0]]
+  cluster_has_non_empty_pools:
+    requires:
+      property:
+        path: hotsos.core.plugins.storage.ceph.CephCluster.cluster_has_non_empty_pools
+  autoscaler_disabled_for_any_pool:
+    requires:
+      property:
+        path: hotsos.core.plugins.storage.ceph.CephCrushMap.autoscaler_disabled_for_any_pool
 conclusions:
   cluster-osds-with-pgs-above-max:
     decision: cluster_has_osds_with_pgs_above_max
@@ -21,7 +29,11 @@ conclusions:
       format-dict:
         limit: hotsos.core.plugins.storage.ceph.CephCluster.OSD_PG_MAX_LIMIT
   cluster-osds-with-suboptimal-pgs:
-    decision: cluster_has_osds_with_suboptimal_pgs
+    decision:
+      and:
+        - cluster_has_osds_with_suboptimal_pgs
+        - cluster_has_non_empty_pools
+        - autoscaler_disabled_for_any_pool
     raises:
       type: CephCrushWarning
       message: >-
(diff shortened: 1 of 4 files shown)


pponnuvel:
@hillpd @lathiat We have briefly talked about this before. Please review.

pponnuvel:
Converted to draft until reviews are done.

hotsos/core/plugins/storage/ceph.py
```diff
@@ -247,12 +247,25 @@ def crushmap_equal_buckets_pretty(self):
...
+                return True
+        return False
+
 
 class CephCluster(object):
     OSD_META_LIMIT_PERCENT = 5
     OSD_PG_MAX_LIMIT = 500
     OSD_PG_OPTIMAL_NUM_MAX = 200
     OSD_PG_OPTIMAL_NUM_MIN = 50
+    POOL_USAGE_LIMIT_PERCENT = 2
```
pponnuvel, line 268: This "2%" is arbitrary - sometimes FE run tests/benchmarks and there's the device_health_metrics that's there as well. If handover takes a few weeks then there might be some data in one of these pools as well.

So realistically we can't expect "0%" utilisation in any cluster.

defs/scenarios/storage/ceph/ceph-mon/pg_imbalance.yaml
```diff
@@ -9,6 +9,14 @@ checks:
       property:
         path: hotsos.core.plugins.storage.ceph.CephCluster.osds_pgs_suboptimal
         ops: [[length_hint], [gt, 0]]
+  pool_utilisation_over_limit:
```
hillpd, line 12: I think the intent here is to check whether the cluster has any non-empty pools.  The current name begs the question, "What limit?" and the arbitrary 2% value isn't explained anywhere in the code.

What do you think about changing the scenario name to be `cluster_has_non_empty_pools` or maybe `cluster_has_ingested_user_data`?

hotsos/core/plugins/storage/ceph.py
hillpd: The 2% check seems reasonable enough. This constant was a bit confusing for me. I'd recommend adding a comment here to clarify the intent, or refactor this to align with the scenario name change above.

hillpd (changes requested):

defs/scenarios/storage/ceph/ceph-mon/pg_imbalance.yaml
pponnuvel: Agree. I have re-named few the vars & the function and commented on the rationale for choosing 2%. Please take another look.

hillpd (approved):
Changes look good. Thanks for the updates!

dosaboy (approved):
i think in the future we can leverage variables and varops to reduce the need for doing as much in python but for now lgtm

PR #413, opened 2022-09-01
[plugin/ceph] Detect PG autoscaling failures

### LINKED ISSUE #411: Report PG autoscaler ignores pools due to overlapping roots

pponnuvel:
When the PG autoscaler module finds pools with overlapping
roots, it doesn't scale them.

Fixes #411.


lathiat:
Its worth noting that in some versions of Ceph (including octopus currently, and our version of pacific) it does actually scale them, but incorrectly with maybe 2-4x times the number of PGs.

Maybe we need to change the message to detail both situations.

dosaboy (approved):
lgtm

PR #414, opened 2022-09-02
[plugin/ceph] Fix laggy PGs check

pponnuvel:
Laggy PG check incorrectly consider backfill_wait,
snaptrim_wait, etc due to a bug in the check which
we exclude here.

### DIFF
--- a/hotsos/core/plugins/storage/ceph.py
+++ b/hotsos/core/plugins/storage/ceph.py
@@ -429,11 +429,11 @@ class CephCluster(object):
             return []
 
         laggy_pgs = []
+        # The states we consider problematic for network related issues
+        laggy_like_states = ['laggy', 'wait']
         for pg in self.pg_dump['pg_map']['pg_stats']:
-            for state in ['laggy', 'wait']:
-                if state in pg['state']:
-                    laggy_pgs.append(pg)
-                    break
+            if any(x in pg['state'].split('+') for x in laggy_like_states):
+                laggy_pgs.append(pg)
 
         return laggy_pgs
 


hotsos/core/plugins/storage/ceph.py
```diff
@@ -430,10 +430,10 @@ def laggy_pgs(self):
 
         laggy_pgs = []
         for pg in self.pg_dump['pg_map']['pg_stats']:
-            for state in ['laggy', 'wait']:
-                if state in pg['state']:
-                    laggy_pgs.append(pg)
-                    break
+            # Looking for 'laggy' or 'wait', but not 'backfill_wait', etc
+            r = re.findall(r'(laggy|^wait|[+]wait$|[+]wait[+])', pg['state'])
```
lathiat, line 434: -1 on this. Although this does match currently, it's not really precise in that this would still match other strings with "laggy" in them.

I think I'd prefer to switch to splitting the string on + and then check the list, e.g.
`
states = pg['state'].split('+')
any(elem in states for elem in ['wait', 'laggy'])
`

Then check if laggy and wait are in the list.
It is apparently not super syntactically simple to check for two items but:

I tried to write a regex to do this but had trouble making it non-greedy so it would correct match both wait and laggy in the same string one after another, e.g. 
`r = re.findall(r'(?:(?:^|\+)(laggy|wait)(?:\+|$)?)', 'active+remapped+laggy+wait+laggy')`

lathiat (changes requested):

hotsos/core/plugins/storage/ceph.py
pponnuvel: I didn't particularly worry about 'laggy' because there are no other [states](https://docs.ceph.com/en/quincy/rados/operations/pg-states/#placement-group-states) with 'laggy' states. Even if there's one in future, it's probably a specialisation of the same and we'd probably want to include that as well.

pponnuvel: But in the interest of accuracy, I've used 'split' as suggested ;-)

dosaboy (approved):
lgtm

PR #417, opened 2022-09-06
Fix support for defining search as list of exprs

dosaboy:
The search property supports different ways of defining a search
expression which can be provided as a string or list of strings.
The latter form was no longer working so this patch fixes that.


pponnuvel (approved):
LGTM

Works with my PR as well (that previously identified this issue).

Issue #418, opened 2022-09-08
Add masakari scenarios


dosaboy:
Add some scenarios to cover masakari failures e.g. compute nodes where pacemaker-remote is not running or installed.

PR #419, opened 2022-09-08
[openstack] add masakari scenario

dosaboy:
Add scenario to check if pacemaker-remote if installed and enabled when compute host is running masakari.

Resolves: #418


dnegreira:
LGTM!

PR #425, opened 2022-09-23
Switch to new shared-memory plug for /dev/shm

lathiat:
Python multiprocessing requires access to files in /dev/shm to work.

Previously snapd only allowed access to /dev/shm files prefixed with the snap name which most applications don't support natively so a snapcraft-preloads LD_PRELOAD hack was used to override and re-write the path being used.

This is now replaced in snapd by a new shared-memory plug that uses a private /dev/shm directory for each snap. This is less clunky and also removes one of the problems for arm64 snaps that were built on amd64 as the LD_PRELOAD was architecture specific. Although we should continue to build arm64 snaps on arm64 using snapcraft remote/hosted builds as per the snapcraft docs.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,7 +14,6 @@ environment:
   LC_ALL: C.UTF-8
   LANG: C.UTF-8
   PYTHONPATH: $SNAP/lib/python3.8/site-packages
-  LD_PRELOAD: $SNAP/libpreload-semaphores.so
 
 # dpkg won't work without this
 layout:
@@ -63,7 +62,7 @@ parts:
         - -lib/systemd/system/cryptdisks.service
     stage:
       - $exclude-systemd-files
-  sem-open-preload:
-    source: https://github.com/snapcore/snapcraft-preloads.git
-    source-subdir: semaphores
-    plugin: make
+
+plugs:
+  shared-memory:
+    private: true


dosaboy:
thanks for the patch i'm gonna have a go at building this locally and running some tests

lathiat:
Forgot this link: https://snapcraft.io/blog/private-shared-memory-support-for-snaps

It no longer seems to be a candidate per that blog, as it works on 20.04 etc and the backported release on those versions is 2 minor versions newer.

dosaboy:
Seems to work well. My only concern is whether enough people will have sufficiently new versions of snapd to have shared-memory support as the docs say it's available as "part of the snapd 2.55.3 release, and it is available both in the Ubuntu 22.04 archive as well the candidate channel (in the Snap Store) on older releases of the distribution"

lathiat:
> Seems to work well. My only concern is whether enough people will have sufficiently new versions of snapd to have shared-memory support as the docs say it's available as "part of the snapd 2.55.3 release, and it is available both in the Ubuntu 22.04 archive as well the candidate channel (in the Snap Store) on older releases of the distribution"

This was only true at the time of that blog post, per snaps generally people are always auto-updated to the latest. Unlike some snaps (such as LXD) snapd has no channels other than latest/stable or latest/edge so you can't sit on an older version other than preventing snapd upgrade itself.

dosaboy:
@lathiat fair enough, lets land it and in the hopefully very unlikely circumstance that we get reports of problems we can have a rethink.

Issue #428, opened 2022-09-26
check ovn logs for certificate expiry or invalid certificates


aserdean:
```2022-09-26T09:51:47.795Z|08398|stream_ssl|WARN|SSL_accept: error:1417C086:SSL routines:tls_process_client_certificate:certificate verify failed```
this would apply for `ovsdb-server-sb.log`, `ovn-controller-log`, `ovn-northd.log`
`2022-09-26T09:47:51.340Z|06924|stream_ssl|WARN|SSL_accept: error:14094415:SSL routines:ssl3_read_bytes:sslv3 alert certificate expired`
this would apply for `ovsdb-server-nb.log`
check if the uptime of central services is newer then certificate issues time.

PR #429, opened 2022-09-26
[openvswitch] check for ovn ssl cert errors

dosaboy:
Resolves: #428


dosaboy:
@aserdean i will have a go at implementing the service uptime part next. I did however check and I can see "SSL routines:ssl3_read_bytes:sslv3 alert certificate expired" for all services in the logs I have, do you think that's not right?

dosaboy:
> @dosaboy: cool.
> 
> Searching for `SSL routines:ssl3_read_bytes:sslv3 alert certificate expired` in all logs is correct.
> 
> It would be nice to also check for the following message: `SSL routines:tls_process_server_certificate:certificate verify failed`. In case the certificate is not expired but invalid.

ack, will do

dosaboy:
still need to add tests

Issue #433, opened 2022-10-04
ignore mon re-elections happening immediately after node boot


dosaboy:
e.g. don't warn on re-elections that happen within an hour or so of a reboot since we would expect that to happen

PR #435, opened 2022-10-04
[ceph] only check mon elections min 1 hour after boot

dosaboy:
We expect some re-elections are a boot/reboot especially when a cluster is first deployed so ignore the first hour after a boot.

Resolves: #433

### DIFF
--- a/hotsos/core/ycheck/engine/properties/search.py
+++ b/hotsos/core/ycheck/engine/properties/search.py
@@ -2,7 +2,7 @@ from datetime import (
     datetime,
     timedelta,
 )
-from hotsos.core.host_helpers import CLIHelper
+from hotsos.core.host_helpers import UptimeHelper, CLIHelper
 from hotsos.core.log import log
 from hotsos.core.search import (
     SearchDef,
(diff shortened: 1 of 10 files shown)


hotsos/core/ycheck/engine/properties/search.py
```diff
@@ -43,11 +43,18 @@ def search_period_hours(self):
...
-        time (in the case of a sosreport would time sosreport was created).
+        Result must have occurred within search-result-age-hours from current
+        time (in the case of a sosreport would be time sosreport was created).
         """
         return int(self.content.get('search-result-age-hours', 0))
 
+    @cached_yproperty_attr
+    def min_hours_since_last_boot(self):
+        """
+        Result must have occurred within hours from last boot.
```
pponnuvel, line 54: You mean: "[..] *after* hours [..]", right?

Maybe reword it to: "Result must have occurred after this many hours since last boot."?

pponnuvel (approved):
LGTM

hotsos/core/ycheck/engine/properties/search.py
dosaboy: yes sorry my docstring says the opposite of what i meant :/ ill fix that now

PR #440, opened 2022-10-13
Ingore lxcfs check when inside lxc container

dosaboy:
We dont need to check for lxcfs bug 1807628 if we are inside a lxc container.


tests/unit/test_system.py
```diff
@@ -162,11 +164,10 @@ def test_unattended_upgrades(self):
         issues = list(IssuesStore().load().values())[0]
         self.assertEqual([issue['desc'] for issue in issues], [msg])
 
+    @mock.patch('hotsos.core.plugins.system.system.SystemBase.'
+                'virtualisation_type', None)
```
KellenRenshaw, line 168: Should we also test kvm here as well?

KellenRenshaw (commented):
LGTM, asked about KVM testing. Otherwise, happy with it.

tests/unit/test_system.py
dosaboy: the check says ne 'lxc' so should be good, I used None in this test to make sure it would be a valid comparison.

PR #448, opened 2022-10-19
[kernel] add sockstat as a data source

### LINKED ISSUE #446: Add `/proc/net/sockstat` as a source
`/proc/net/sockstat` contains valuable information regarding per-protocol resource usage. Itemized list of obtainable information is as follows:

```
'TotalSocksInUse',
'TcpSocksInUse',
'TcpSocksOrphaned',
'TcpSocksInTimeWait',
'TcpSocksAllocated',
'TcpSocksMemPagesInUse',
'UdpSocksInUse',
'UdpSocksMemPagesInUse',
'RawSocksInUse',
'RawSocksMemPagesInUse',
'FragSocksInUse',
'FragMemPagesInUse'
```

It would be beneficial for us to have these values available at hand, especially the number of memory pages used by each protocol is quite useful in detecting some problematic scenarios. 

Related-to: #445

mustafakemalgilor:
Adds /proc/net/sockstat reading support
to the kernel plugin.

Fixes #446.

Signed-off-by: Mustafa Kemal Gilor <mustafa.gilor@canonical.com>

### DIFF
--- a/hotsos/core/plugins/kernel/net.py
+++ b/hotsos/core/plugins/kernel/net.py
@@ -220,3 +220,81 @@ class NetStatTCP(NetStatBase):
             'TCPReqQFullDoCookies',
             # skb retrans before original left host
             'TCPSpuriousRtxHostQueues']
+
+
+class SockStat(ProcNetBase):
+
+    """
+    Provides a common way to extract fields from /proc/net/sockstat.
+
+    Expected file format is:
+
+    hdr1: label1 value1 label2 value2...
+    hdr2: label1 value1 label2 value2...
+    hdr3: label1 value1 label2 value2...
+    """
+    def __init__(self):
+        super().__init__()
+        self._process_file(os.path.join(HotSOSConfig.DATA_ROOT,
+                                        'proc/net/sockstat'))
+
+    def _process_file(self, fname):
+        if not os.path.exists(fname):
+            log.info("file not found '%s' - skipping load")
+            return
+
+        log.debug("start processing %s", fname)
+        with open(fname) as fd:
+            for line in fd:
+                psl = line.split(':')
+                if len(psl) != 2:
+                    log.warning("badness: line does not match format %s", line)
+                    continue
+                (label, stats) = psl
+                if label not in self._data:
+                    self._data[label] = {}
+                try:
+                    stats = stats.strip().split(' ')
+                    stats = dict(stats[i:i+2] for i in range(0, len(stats), 2))
+                    # Convert string values to `int``
+                    stats = {k: int(v) for k, v in stats.items()}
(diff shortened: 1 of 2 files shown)


hotsos/core/plugins/kernel/net.py
```diff
@@ -220,3 +220,81 @@ def _fields(self):
             'TCPReqQFullDoCookies',
             # skb retrans before original left host
             'TCPSpuriousRtxHostQueues']
+
+
+class ProcSockStat():
```
dosaboy, line 225: can this be:

class SockStat(object):

hotsos/core/plugins/kernel/net.py
```diff
@@ -220,3 +220,81 @@ def _fields(self):
...
+    hdr1: label1 value1 label2 value2...
+    hdr2: label1 value1 label2 value2...
+    hdr3: label1 value1 label2 value2...
+    """
+    def __init__(self):
+        self._data = {}
+        self._process_file(os.path.join(HotSOSConfig.DATA_ROOT,
+                                        'proc/net/sockstat'))
+
+    def _process_file(self, fname):
```
dosaboy, line 241: I would be great if we could somehow adapt the baseclass to be able to cover both formats but i guess that can be done seperately.

tests/unit/test_kernel.py
```diff
@@ -285,6 +304,40 @@ def test_slab_major_consumers(self):
         self.assertEqual(top5, expected)
 
 
+class TestKernelNetworkInfo(TestKernelBase):
+    @utils.create_data_root({'proc/net/sockstat': PROC_SOCKSTAT})
+    def test_sockstat_parse(self):
+        uut = ProcSockStat()
+        self.assertEqual(uut.NsTotalSocksInUse, "908")
```
dosaboy, line 311: as mentioned above, integer values should be saved as int() so that we can compare them with integers (not strings).

hotsos/core/plugins/kernel/net.py
```diff
@@ -220,3 +220,81 @@ def _fields(self):
...
+        with open(fname) as fd:
+            for line in fd:
+                psl = line.split(':')
+                if len(psl) != 2:
+                    log.warning("badness: line does not match format %s", line)
+                    continue
+                (label, stats) = psl
+                if label not in self._data:
+                    self._data[label] = {}
+                try:
```
dosaboy, line 256: Need to make sure integers are saved as int() so that they can be compared with integers.

tests/unit/test_kernel.py
mustafakemalgilor: done

dosaboy (approved):
lgtm thanks mkg!

PR #453, opened 2022-10-25
Set tcp check counter limits to 500

dosaboy:
Warning when counters are non-zero is not useful since in general very small drop counts are not noticeable especiablly when a host has been up for a long time.


nicolasbock:
This sounds like it might be more useful to measure drops / uptime as opposed to simply the raw count. Would that be hard to change?

dosaboy:
These counters are showing drops since uptime i.e. just a snapshot and not necessarily reflecting current behaviour which makes it tricky to determine their usefulness or reliability. Some of the checks are done use a percentage of drops which while not perfect feels a better measure but at the very least I think a higher threshold is better than 0. I discussed with @jay-vosburgh and actually agrees to use a lower threshold like 500 but in the future we could certainly implement more complex logic.

nicolasbock:
Thanks for the clarification @dosaboy

nicolasbock (approved):

PR #454, opened 2022-10-25
More kern net check tweaks

dosaboy:
Dont warn on TCP_DEFER_ACK drops since they do not reflect actual drops. Also extend some checks to
consider either drops > 500 or > 1% total.


nicolasbock (approved):
That looks great!

jay-vosburgh (approved):

PR #460, opened 2022-11-04
Fix snap to support cryptography module

nicolasbock:

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -9,10 +9,11 @@ confinement: classic
 grade: stable
 base: core20
 environment:
-  REPO_INFO_PATH: $SNAP/repo-info
-  LC_ALL: C.UTF-8
   LANG: C.UTF-8
-  PYTHONPATH: $SNAP/lib/python3.8/site-packages
+  LC_ALL: C.UTF-8
+  PYTHONHOME: ${SNAP}/usr
+  PYTHONPATH: ${SNAP}/usr/lib/python3.8:${SNAP}/lib/python3.8/site-packages
+  REPO_INFO_PATH: ${SNAP}/repo-info
 
 # dpkg won't work without this
 layout:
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -24,40 +25,47 @@ layout:
...
+      - python3-pip
+      - python3-venv
+      - rsync
+    stage-packages:
+      - bsdmainutils
+      - coreutils
+      - dpkg
+      - libc6
+      - libpython3.8-minimal
+      - libpython3.8-stdlib
```
dosaboy, line 44: if we are going to carry a python version id like to carry the same one the core snap has. In the case of core20 that is 3.8 but then we recently started running all tests using 3.10 which is jammy's version so could we switch core20 -> core22 and bump 3.8 -> 3.10?

snap/snapcraft.yaml
```diff
@@ -24,40 +25,47 @@ layout:
...
+      - python3.8-minimal
+      - systemd
     build-environment:
       - "CRYPTOGRAPHY_DONT_BUILD_RUST": "1"
+    override-build: |
+      git rev-parse --short HEAD > repo-info
+      python3 -m venv ${SNAPCRAFT_PART_INSTALL}
+      ${SNAPCRAFT_PART_INSTALL}/bin/pip install -r requirements.txt
+      ${SNAPCRAFT_PART_INSTALL}/bin/python3 setup.py install
+      rsync -av defs ${SNAPCRAFT_PART_INSTALL}/etc/hotsos/
```
dosaboy, line 54: this bit (rsync) doesn't seem relevant to the crypto issue, is it needed?

snap/snapcraft.yaml
```diff
@@ -24,40 +25,47 @@ layout:
...
         - -lib/systemd/system/hwclock.service
         - -lib/systemd/system/rc.service
-        - -lib/systemd/system/x11-common.service
-        - -lib/systemd/system/cryptdisks-early.service
         - -lib/systemd/system/rcS.service
-        - -lib/systemd/system/cryptdisks.service
+        - -lib/systemd/system/x11-common.service
     stage:
       - $exclude-systemd-files
+      - -tests
```
dosaboy, line 71: what is this last line doing?

dosaboy (changes requested):
hi nicolas, this is great work. a couple of inline comments for you and also, could you create an issue and move the information in your commit msg to that issue (and ref the issue in the commit) in case anyone wants to comment on the findings and solution. thanks.

snap/snapcraft.yaml
```diff
@@ -24,40 +25,47 @@ layout:
...
 parts:
-  hotsos-cli:
-    plugin: dump
-    source: .
-    override-build: |
-      rm -rf tests
-      git rev-parse --short HEAD > repo-info
-      snapcraftctl build
-    organize:
-      defs: etc/hotsos/defs
```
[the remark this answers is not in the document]
nicolasbock: I realized that this wasn't working before because I hadn't copied the `snapcraftctl build` line. But when I re-add it the `organize` section works again as it did before.

snap/snapcraft.yaml
nicolasbock: I am running into issues with switching to core22. It's like linking hell all over :/ What do you think of merging this PR while I keep working on the core22 upgrade. Merging this change here would mean we finally have a working snap.

snap/snapcraft.yaml
nicolasbock: Fixed in latest iteration

dosaboy (approved):
lgtm. i want to move to core22/py3.10 asap but since we are already stuck with py38 anyway i see no reason not to land this.

Issue #240, opened 2021-12-07
support alternate TZ format from date command


dosaboy:
We currently support the string format of timezones e.g. UTC but the date command can also do numeric e.g.
Tue Nov 30 18:03:18 +05 2021

PR #462, opened 2022-11-08
Support date with numeric TZ

dosaboy:
The linux date command can specify a timezone
using a string or numeric offset. This ensures
we also support the latter. Also ensures that
original tz is included with --date input to
ensure correct conversion.

Resolves: #240


nicolasbock (approved):
LGTM

PR #467, opened 2022-11-11
Support helper factory objects with no attribute

dosaboy:
The current implementation of factory objects assumes that an attribute will be called on the generated object but this is not always desired so now we all generating an object that will not have an attr called on it. This allows us to simpligy the sysctl helper factory which is applied here.

Also cleans up the factory docs entry.


mustafakemalgilor (approved):
LGTM :shipit:

Issue #445, opened 2022-10-18
Check TCP/UDP used page count against allowed maximum (/net/ipv4/{udp,tcp}_mem)


mustafakemalgilor:
At any given time, an absolute maximum number of pages is allowed by the kernel for all TCP/UDP sockets. This value is kept in `net.ipv4.{tcp,udp}_mem` sysctl settings and automatically set at boot time (depending on available system memory).

The `*_mem` value is a three-value vector, consisting of `min, pressure`, and `max` values. If the number of pages allocated for {tcp,udp} reaches to the `max` value, the kernel will start to drop frames.

The number of allocated pages for TCP/UDP at any given time can be observed on the relevant line in `/proc/net/sockstat` file:

``` 
sockets: used 1488
TCP: inuse 36 orphan 0 tw 1 alloc 99 mem 7
UDP: inuse 53 mem 47
UDPLITE: inuse 0
RAW: inuse 0
FRAG: inuse 0 memory 0
```

The `mem` value in each line indicates the number of pages allocated for each protocol. So, in theory, we can compare the actual value against the maximum value to issue a warning (e.g. %85 of allowed pages are reached for UDP). 

This would allow us to notice some problematic scenarios such as: 

- An application that could not keep up with the load
- A resource leak in the kernel

Another take on this might be trying to correlate the actual memory used by all active sockets to the number of total pages used by TCP/UDP sockets. For TCP and UDP, all alive sockets are available under `/proc/net/{tcp,udp}`, together with their respective TX & RX queue sizes, which may help us to estimate how much space all the sockets are occupying at any given moment.

mustafakemalgilor:
Hi @sombrafam ;

Of course, let's look at a concrete example from a host we suspect that have some kind of memory leak:

```text
sockets: used 908
TCP: inuse 22 orphan 0 tw 2 alloc 58 mem 15
UDP: inuse 18 mem 762066
UDPLITE: inuse 0
RAW: inuse 0
FRAG: inuse 0 memory 0
```

Right of the bat, we could tell that something is not right with UDP memory usage, since there are 18 active sockets and `762066` pages of memory used for them. Let's look at the `sysctl` settings for UDP to make sense of it:

```
                   #min   #press  #max
net.ipv4.udp_mem = 381033 508045 762066 #unit: pages
net.ipv4.udp_rmem_min = 4096   #unit: bytes
net.ipv4.udp_wmem_min = 4096   #unit: bytes
[...]

PR #468, opened 2022-11-14
Add scenario checks for TCP and UDP mempages usage

mustafakemalgilor:
This MP introduces new TCP/UDP scenario warnings for high mempages usage. Also, fixes a small bug in systemd start_time parsing. 

Resolves: #445


dosaboy (approved):
lgtm. i want to move to core22/py3.10 asap but since we are already stuck with py38 anyway i see no reason not to land this.

dosaboy (approved):
lgtm!

PR #470, opened 2022-11-15
Add explicit `patchelf` part

nicolasbock:
This part is not necessary for `core20` but will be required for `core22`.


nicolasbock:
Second attempt. I don't see anything wrong in the build log for revision 600. Can we build this on launchpad so I can inspect the snap?

nicolasbock:
Test with snap from launchpad.

snap/snapcraft.yaml
```diff
@@ -24,43 +25,125 @@ layout:
...
+      # To allow for a classic snap that works across multiple operating system
+      # runtime environments, we need to ensure all shipped binaries consume
+      # libraries from the core or shipped snap.  We accomplish this by patching
+      # `RPATH`` or interpreter into dynamically linked binaries.
+      #
+      # /snap/${CORE}/current/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+      # /snap/${CORE}/current/lib/aarch64-linux-gnu/ld-linux-aarch64.so.1
+      # /snap/${CORE}/current/lib/powerpc64le-linux-gnu/ld64.so.2
+      # /snap/${CORE}/current/lib/s390x-linux-gnu/ld64.so.1
+      CORE=core20
```
mustafakemalgilor, line 124: TBH, this seemed a bit hacky at first glance to me, but even the [official blog post](https://snapcraft.io/blog/the-new-classic-confinement-in-snaps-even-the-classics-need-a-change) is recommending this approach. If I am understanding this correctly, the snap's `--classic` confinement causes application binaries to rely on system libraries instead of the shipped snap libraries. The default `--strict` confinement avoids this, but it's not suitable for hotsos's needs. I think `snapd` needs a middle ground between these two: `Use snap's root while dynamic loader loads the application dependencies, switch to system's root afterwards`. But this strategy may have some shortcomings if the application is loading its libraries at runtime (e.g. plugins loaded on-demand).

snap/snapcraft.yaml
```diff
@@ -24,43 +25,125 @@ layout:
...
+          fi
+      done
+      if [[ ${interpreter} == /dev/null ]]; then
+          echo "Unable to determine name of shared library interpreter."
+          exit 1
+      fi
+
+      for binary in \
+        $(find . -exec file {} \; | awk -F : '/ELF.*dynamic/ { print $1 }'); do
+          patchelf \
```
mustafakemalgilor, line 141: I think this loop could use some log output (e.g. setting rpath of file %s to %s).

snap/snapcraft.yaml
```diff
@@ -24,43 +25,125 @@ layout:
...
+      # touch the first line and absorb the whole line. Note that the matching
+      # expression cannot be more specific since the path to the Python
+      # interpreter depends on the build environment. Notably, local builds
+      # (`/root`) differ from builds on launchpad (`/build`).
+      sed --in-place \
+        --expression "1 s:^#\!/.*$:#\!/snap/hotsos/current/usr/bin/${PYTHON}:" \
+        ${SNAPCRAFT_PART_INSTALL}/bin/hotsos
+
+      # Add a missing link in snap to Python interpreter.
+      ln -sf ../usr/bin/${PYTHON} ${SNAPCRAFT_PART_INSTALL}/bin/python3
```
mustafakemalgilor, line 87: Is it okay for the `source` to be a relative path? Would that cause any problems if python3 is invoked from a different working directory?

nicolasbock: It's relative to `${SNAPCRAFT_PART_INSTALL}/bin`. In other words it will link to
`/snap/hotsos/current/usr/bin` in the snap.

snap/snapcraft.yaml
nicolasbock: I am not sure "rely on" is the right way of thinking about this. In a strictly
confined snap the filesystems are mounted into the snap namespace so that the
snap cannot "see" the host's filesystem. In a classically confined snap that's
not the case and the snaps are in `/snap/${SNAPNAME}/current`. This means that
if you, e.g., execute a shell script from your snap, the `bash` interpreter will
be the one from the host. The same is true for Python scripts. This behavior
might lead to unexpected consequences but will work most of the time.

In this case here the problem lies with what libraries the dynamic linker loads
when `hotsos` loads the `cryptography` module. There is a rust extension that
needs to be loaded, hence the involvement of the dynamic linker. At some point
in the library dependency chain, the loader will want to load `libphread.so` and
it will load the one from the host because of the aforementioned mounting
behavior of classic snaps. This causes a failure if the host is running a newer
version of `libc` than the one the rust extension in the snap was compiled with.
Using the hacky approach and patch the loader paths to include the
`/snap/${SNAPNAME}/....` prefix addresses this issue. Note that this also
requires that we install `libc` in the snap which is normally provided by the
`core??` snap itself. But since we can't patch anything in another snap we need
to install `libc` here ourselves.

I hope this clarifies the reason for patching elf files.

snap/snapcraft.yaml
nicolasbock: I added `set -x` to the script. Does that address your logging comment?

nicolasbock:
@dosaboy The test
```console
$ faketime '2024-01-01' hotsos --openstack --format yaml tests/unit/fake_data_root/ssl-cert
```
fails without this PR on Jammy with:
```
  potential-issues:
    HotSOSScenariosWarnings:
      - One or more scenarios failed to run (openstack_apache2_certificates) - run
        hotsos in debug mode (--debug) to get more detail (origin=openstack.auto_scenario_check)
```

lathiat:
Can we avoid importing another large sosreport into tests/unit/fake_data_root? The existing ones are already making the git repository quite large and clunky to check out.

3,000 files is better than the 13,000-85,000 the other ones have but would still be nice to avoid or slim down if possible.

sys/fs/cgroup makes up a fairly large portion of most of them. not sure we need it here?

mustafakemalgilor:
> Can we avoid importing another large sosreport into tests/unit/fake_data_root? The existing ones are already making the git repository quite large and clunky to check out.

I think it would be even better to compress them into an archive and extract them on demand (i.e., test setup stage).

dosaboy:
> Can we avoid importing another large sosreport into tests/unit/fake_data_root? The existing ones are already making the git repository quite large and clunky to check out.
> 
> 3,000 files is better than the 13,000-85,000 the other ones have but would still be nice to avoid or slim down if possible.
> 
> sys/fs/cgroup makes up a fairly large portion of most of them. not sure we need it here?

agreed

mustafakemalgilor (approved):

PR #473, opened 2022-11-16
Build snap in GitHub CI

nicolasbock:
This change builds and stores the snap on amd64 and offers the snap as a build
artifact for download and testing.

### DIFF
--- a/.github/workflows/snapfunctionaltest.yaml
+++ b/.github/workflows/snapfunctionaltest.yaml
@@ -1,0 +1,20 @@
+name: Build snap
+on:
+  - push
+  - pull_request
+
+jobs:
+  snap:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


dosaboy (commented):
We used to use github builds but moved away because we couldn't get multi-arch builds to work (in fact they do build but not as multi-arch). if we can get to work then I'm for moving back to github.

nicolasbock:
We would have to cross compile which might not work for all architecture combinations we care for. I meant this change more as a way to run things in CI using the (`amd64`) hotsos snap. Sort of as a functional test.

nicolasbock:
The issue we ran into with the shebang not being properly rewritten might have come up here.

dosaboy (commented):
i think im ok with doing this but can we name this file something more intuitive like snapbuildtest or similar

nicolasbock:
Done @hopem

Issue #477, opened 2022-11-25
keystone exceptions not being identified


dosaboy:
hotsos is not finding exceptions in keystone logs because they are not printed using the normal exception format using Traceback etc. Also since logs are always prepended with a module name this is causing the search constraints to fail to verify lines. Solution is to add a custome expr for keystone exceptions and skip constraints for keystone logs.

PR #478, opened 2022-11-25
Ensure keystone exceptions are identified

dosaboy:
Fixes the agent exceptions extension to ensure that exceptions in keystone logs are identified. Also
ensure that extra regex used for keystone and apache services are only added when searching logs from
those servies.

Resolves: #477


pponnuvel (approved):
LGTM

PR #481, opened 2022-12-07
Add check for neutron ovn metadata issue

zhhuabj:


dosaboy (commented):
Hi Joshua, thanks for this patch. The issue you are trying to detect requires more than one node to assert i.e. bug 1996594 is the result of ovs < 2.17.0 + leadership change + neutron-ovn-metadata-agent having an OVSDBError. The problem is that you need to look at logs from both ovn-central AND ovn-chassis to be able to detect this problem accurately. As it stands your scenario will look at ovn-central only and since a leadership change in itself is not a bug I dont think we can say that is enough to raise this issue. An alternative could be to keep the no_monitor_cond_since check but change has_1996594 to look at var/log/neutron/neutron-metadata-agent.log for OVSDBError. You can then say that it is possible the bug has been hit and that the user can go check the ovn-central logs to see if the leadership change occured around the same time (manually).

Oh and if you make ^^ change, this scenario should then be moved to defs/scenarios/openstack (since we'd be checking neutron)

zhhuabj:
Hi Ed, thanks for reviewing, the new code has been refactored according to your comment, but I will first hold on and then upload it after the customer replies and confirms that the upgrade can solve their problems. More exciting after the advertisement :)

zhhuabj:
The customer has confirmed they haven't faced the issue since they upgraded to Yoga.

dosaboy:
It's been several months since this patch had an update so ill go ahead and close the PR. Please feel free to re-open it if you have further updates.

PR #483, opened 2022-12-08
Add checks for lp#1996230

drencrom:
Check juju error log and give pointer to knowledge base article with the scripts used to solve this issue.


dosaboy:
now that tox is fixed on main ill rebase+merge

PR #487, opened 2022-12-10
Add testing with `mutmut`

nicolasbock:


dosaboy:
i'm going to go ahead and close this for now and we can revisit later if we want to add this. If we do decide to do it it will probably be enabled via https://github.com/canonical/se-tooling-ci-common

PR #491, opened 2022-12-13
Add unit tests for `cli.py`

nicolasbock:


tests/unit/test_cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+import unittest
+from unittest import mock
+
+import hotsos.cli
+
+
+class TestCLI(unittest.TestCase):
+
+    def test_get_hotsos_root(self) -> None:
```
dosaboy, line 13: does the method annotations have a purpose here?

nicolasbock: I mean it's just a habit of mine. We could eventually start checking using [mypy](http://mypy-lang.org/)

nicolasbock: If we merge canonical/hotsos#494 we can start statically type checking the code.

nicolasbock:
> Adding/using annotations feels like a policy/style change and if we are going to do that I'd like to understand the objective and merits so can I please ask that we evaluate that separately to this patchset. 

Done, I removed those annotations.

PR #493, opened 2022-12-13
Test in parallel

nicolasbock:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
 basepython = python3
-commands = stestr run --serial --test-path {[testenv]unit_tests} {posargs}
+commands = stestr run --test-path {[testenv]unit_tests} {posargs}
 
 [testenv:pep8]
 commands = flake8 -v --exclude=fake_data_root {posargs:{[testenv]pyfiles}}


dosaboy:
So I've thought about doing this before but i don't think its safe because, for one thing, of global config i.e. HotSOSConfig global to all threads and not thread safe so one test can modify the config of another test. Having said that I don't think it would be too much work to make this so we can look into that for sure.

nicolasbock:
I might misunderstand how `stestr` is running in parallel but I presumed that it's creating temporary directories and runs tests in parallel in those. But I will look into this more carefully.

dosaboy:
> I might misunderstand how `stestr` is running in parallel but I presumed that it's creating temporary directories and runs tests in parallel in those. But I will look into this more carefully.

the tmp dirs are fine, i think the issue is with import namespaces which will still be global to all tests

dosaboy:
Having run some tests locally (on admittedly not a super powerful machine) it seems that running the unit tests in parallel has no noticeable gain over running them serialised (which i suppose i am little curious about) so given that there is no clear benefit i propose we pass on this for now.

nicolasbock:
But then why use `stestr` at all? From the [documentation](https://github.com/mtreinish/stestr#overview):

> stestr is parallel Python test runner designed to execute [unittest](https://docs.python.org/3/library/unittest.html) test suites using **multiple processes** to split up execution of a test suite. It also will store a history of all test runs to help in debugging failures and optimizing the scheduler **to improve speed**. To accomplish this goal it uses the [subunit](https://github.com/testing-cabal/subunit) protocol to facilitate streaming and storing results from **multiple workers**.

If we don't care about or don't see any benefit from the parallel aspects of `stestr` we should just use `unittest` directly.

PR #494, opened 2022-12-13
Add `tox` environment for `mypy`

nicolasbock:
[Mypy](http://mypy-lang.org/) is a static type checker for Python.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -32,3 +32,6 @@ commands = bashate --verbose {toxinidir}/build.sh {toxinidir}/tools/test/functes
 
 [testenv:functional]
 commands = {toxinidir}/tools/test/functest.sh
+
+[testenv:mypy]
+commands = -mypy --install-types --non-interactive hotsos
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -32,3 +32,6 @@ commands = bashate --verbose {toxinidir}/build.sh {toxinidir}/tools/test/functes
 
 [testenv:functional]
 commands = {toxinidir}/tools/test/functest.sh
+
+[testenv:mypy]
+commands = -mypy hotsos
```
mustafakemalgilor, line 37: looking at the ci job output, mypy is complaining about `error: Library stubs not installed for "yaml"  [import]`. The official [documentation](https://mypy.readthedocs.io/en/stable/running_mypy.html#library-stubs-not-installed) is suggesting to run mypy with `--install-types --non-interactive` flags to automatically install the stubs as needed.

nicolasbock: Thanks for the hint @mustafakemalgilor ! Fixed.

dosaboy:
i'm going to go ahead and close this for now and we can revisit later if we want to add this. If we do decide to do it (and I am in favour of doing so but need to look at the pros/cons and how to enable it the right way) it will probably be enabled via https://github.com/canonical/se-tooling-ci-common

PR #495, opened 2022-12-13
Add manual workflow trigger

nicolasbock:

### DIFF
--- a/.github/workflows/snapfunctionaltest.yaml
+++ b/.github/workflows/snapfunctionaltest.yaml
@@ -2,6 +2,7 @@ name: Build snap
 on:
   - push
   - pull_request
+  - workflow_dispatch
 
 jobs:
   snap:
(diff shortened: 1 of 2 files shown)


dosaboy:
Any chance you can add some info to the commit about why you are adding this? As I understand it this allows you to run a workflow manually but since they are run on every push why would we need this?

dosaboy:
It is possible to allow people to re-run workflows/jobs. I am able to do this already but that may not be the case for all users but I can take a look if it can be enabled. Presumably if that is possible this would not be needed?

nicolasbock:
Interesting, I was assuming that since I can't re-run jobs it's because this trigger is missing. But maybe that's not the problem after all.

nicolasbock:
The reason I was even looking at this was because for some reason CI didn't run for the last change set of canonical/hotsos#491 . I wanted to trigger a run but couldn't.

dosaboy:
Ok i found the answer - https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs - so ill land this patch and we'll see how it goes.

PR #496, opened 2022-12-13
Add check for LP 1960319

dosaboy:


defs/scenarios/openstack/neutron/bugs.yaml
```diff
@@ -59,6 +59,13 @@ checks:
         min-results: 10
         search-period-hours: 1
         search-result-age-hours: 48
+  has_1960319_log:
+    input: var/log/neutron/neutron-server.log
+    expr: '(\S+\s+[\d:]+).\S+ .+ ovsdbapp.event FileNotFoundError: \[Errno 2\] No such file or directory: ''ovsdb-client'''
```
mustafakemalgilor, line 64: We can prepend a `^` to `expr` to make `expr` only match from the beginning of a new line if the regex search is being made with the `multiline` flag.

mustafakemalgilor: Otherwise, LGTM :shipit:

dosaboy: @mustafakemalgilor thanks, my understanding is that since we are using re.match() (and not re.search()) this will effectively start at the beginning of the line and anything else will be rejected (e.g. if the line starts with whitespace). We are also not using multiline search (for reasons). I tested with the following which makes the test fail:

--- a/defs/tests/scenarios/openstack/neutron/bugs_lp1960319.yaml
+++ b/defs/tests/scenarios/openstack/neutron/bugs_lp1960319.yaml
@@ -18,7 +18,7 @@ data-root:
       2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event     self._execute_child(args, executable, preexec_fn, close_fds,
       2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event   File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child
       2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event     raise child_exception_type(errno_num, err_msg, err_filename)
-      2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event FileNotFoundError: [Errno 2] No such file or directory: 'ovsdb-client'
+       2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event FileNotFoundError: [Errno 2] No such file or directory: 'ovsdb-client'
       2022-02-05 15:08:46.198 75085 ERROR ovsdbapp.event
   copy-from-original:
     - sos_commands/date/date

Issue #498, opened 2023-01-06
Add option to disable search constraints


dosaboy:
The current implementation for search constraints uses a binary search that sometimes needs to search backwards to find a last known good line i.e. in log files that contain lines that don't start with a timestamp we treat those as unverifiable and so have to find the most recent verifiable line which can be ahead or behind the current position. Seeking backwards seems to force a SEEK_SET in the kernel which for very large files becomes very slow as it will start from the start if file each time. Until we find a better approach I am going to add a config option to disable search constraints for files that contain unverifiable lines. The will not have much impact other than some search results might contain more data than expected but it should speed things up for hosts that have large log files with lots of unverifiable data/lines in them.

PR #499, opened 2023-01-06
Add new allow_constraints_for_unverifiable_logs config

dosaboy:
Disables constraints for files that contain unverifiable lines and adds a config option to re-enable them. This is until we have a way to apply constraints for files containing lines that dont have timestamps without it being really slow for large files.

Resolves: #498

### DIFF
--- a/hotsos/core/config.py
+++ b/hotsos/core/config.py
@@ -112,6 +112,28 @@ class SearchtoolsConfigOpts(ConfigOptGroupBase):
                                         'log files, this is used to limit the '
                                         'logrotate history in days.'),
                            default_value=7))
+        self.add(ConfigOpt(name='allow_constraints_for_unverifiable_logs',
+                           description=('Search constraints use a binary '
+                                        'search that sometimes needs to '
+                                        'seek backwards to find a last '
+                                        'known good line i.e. in log files '
+                                        "that contain lines that don't start "
+                                        'with a timestamp we treat those as '
+                                        'unverifiable and so have to find the '
+                                        'most recent verifiable line which '
+                                        'can be ahead or behind the current '
+                                        'position. Seeking backwards seems to '
+                                        'force a SEEK_SET in the kernel '
+                                        'regardless of what whence is set to '
+                                        'which for large files becomes very '
+                                        'slow as it will start from 0 each '
+                                        'time. Backwards seeking is now not '
+                                        'supported by default and requires '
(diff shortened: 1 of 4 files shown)


hotsos/core/config.py
```diff
@@ -112,6 +112,25 @@ def __init__(self):
                                         'log files, this is used to limit the '
                                         'logrotate history in days.'),
                            default_value=7))
+        self.add(ConfigOpt(name='allow_constraints_for_unverifiable_logs',
+                           description=('Search constraints use a binary '
+                                        'search that sometimes needs to '
+                                        'search backwards to find a last '
+                                        'known good line i.e. in log files '
+                                        "that contain lines that don't start "
```
brianphaley, line 120: My nitpicking eyes were drawn to the "" here :-p

brianphaley (approved):

hotsos/core/config.py
```diff
@@ -112,6 +112,25 @@ def __init__(self):
...
+                                        'with a timestamp we treat those as '
+                                        'unverifiable and so have to find the '
+                                        'most recent verifiable line which '
+                                        'can be ahead or behind the current '
+                                        'position. Seeking backwards seems to '
+                                        'force a SEEK_SET in the kernel which '
```
pponnuvel, line 126: I understand the issue - but I doubt it's the *kernel* setting SEEK_SET (I suspect it's somewhere in the Python lib).

pponnuvel (approved):
LGTM

hotsos/core/config.py
dosaboy: thank brian, yeah that's cause of the ' in the string

hotsos/core/config.py
dosaboy: yeah I can reword this as its not entirely correct, basically what I see is that if i am at pos 100002 and want to seek(SEEK_SET, 100001) it will seek from the start of the file which may be intentional but if i do seek(SEEK_CUR, -1) it also forces to SEEK_SET and starts from 0 (which i can see in strace).

hotsos/core/config.py
brianphaley: Oh yeah, didn't (or did not) even notice :)

hotsos/core/config.py
dosaboy: ive updated the description fwiw

Issue #480, opened 2022-11-29
OSD is not able to cache onodes


taodd:
Details are at [1]
one is to check below osd log:

"2022-10-25T00:47:26.562+0000 7f424f78e700 30 bluestore.MempoolThread(0x564a9dae2a68) _resize_shards max_shard_onodes: 0 max_shard_buffer: 8388608"

If the max_shard_onodes is 0, this needs to be flagged and strongly indicating it might causing the bug above. 

Another way is checking (if the dump_mempool is available) if the onode number is very small
Like below:

"bluestore_cache_onode": {
"items": 3,
"bytes": 1848
},

We should flag when the onode's number is less than 100

[1] https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1996010

pponnuvel:
An example dump_mempool at `/customers/bloomberg/00345802/ceph-mempools-after.tar.xz`

PR #503, opened 2023-01-10
Add scenario for LP 1996010

dosaboy:
Resolves: #480


defs/scenarios/storage/ceph/ceph-osd/bugs.yaml
```diff
@@ -10,6 +10,10 @@ checks:
     - property:
         path: hotsos.core.plugins.storage.ceph.CephDaemonConfigShowAllOSDs.bluestore_volume_selection_policy
         ops: [[ne, []]]
+  has_1996010:
+    # NOTE: this needs quite a high debug level to appear - debug_bluestore=20/20
```
taodd, line 14: It needs debug_bluestore=30/30

taodd (commented):
The log check is enough to detect the issue, but it needs debug_bluestore=20.

Another place to check is the bluestore_cache_onode.items from the output of dump_mempools, we can flag the issue when bluestore_cache_onode.items < 10.

dosaboy:
@taodd ive hopefully addressed all points you raised

Issue #363, opened 2022-06-20
Check if juju osd config ceph-cluster-network and ceph-public-network contain multiple ip addresses


taodd:
Pay attention that if the options ceph-cluster-network and ceph-public-network contains multiple IP addresses
Report a warning if any of the options contains multiple IP addresses.

E.g suppose 10.55.136.0/24  is the public network and 10.55.134.0/24 is the replication network. 
So, we should expect ceph-cluster-network to be '10.55.134.0/24' and ceph-public-network to be '10.55.136.0/24'
if we see a string value like '10.55.136.0/24 10.55.134.0/24' at any of the two options, raise a warning about this that the cluster might have a connectivity issue.

dosaboy:
@taodd we can only observer this if it is in ceph.conf, is that going to be possible?

dosaboy:
ok @taodd confirmed that (a) yes it is observable in ceph.conf and also (b) ceph does actually support using more than one address for each of these configs BUT in this error case the same addresses were used for both cluster/public and we also typically would not expect that to be used. So based on that I think we can define a scenario.

PR #504, opened 2023-01-10
Check for overlapping addresses in ceph.conf

dosaboy:
Resolves: #363

### DIFF
--- a/hotsos/core/plugins/storage/ceph.py
+++ b/hotsos/core/plugins/storage/ceph.py
@@ -71,10 +71,59 @@ CEPH_REL_INFO = {
 CEPH_POOL_TYPE = {1: 'replicated', 3: 'erasure-coded'}
 
 
+def csv_to_set(f):
+    """
+    Decorator used to convert a csv string to a set().
+    """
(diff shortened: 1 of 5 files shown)


taodd (approved):
LGTM

hotsos/core/plugins/storage/ceph.py
```diff
@@ -71,11 +71,60 @@
...
+        if val is not None:
+            return val
+
+        if hasattr(self, orig_key):
+            return getattr(self, orig_key)
+
+    @property
+    @csv_to_set
+    def cluster_addr_set(self):
+        return self.get('cluster addr')
```
pponnuvel, line 121: I think you want to check `cluster network` and `public network`. `cluster addr` and `public addr` are single value each anyway.

Also: the value can be specified as `cluster network` or `cluster_network` (similarly `public network` or `public network`) in the ceph.conf file. Will this work for both?

dosaboy: bah you are right, it even says that in the issue :0 i'll get that fixed

Issue #437, opened 2022-10-11
Easy to confuse charm version with unit


lathiat:
In the current output, it is easy to confuse the charm version with the unit number. Partly because they are both in the same format with a dash, and, also because the charm version number is listed first in the output.

When quickly scanning through I mistakenly assumed 'ovn-central/3' instead of 'ovn-central/0' in the following output:

```
juju:   
  services:
    systemd:
      enabled:
        - jujud-machine-3-lxd-8
    ps: 
      - jujud (1)
  version: 2.9.11
  machine: 3-lxd-8
  charms:
    - filebeat-24
    - landscape-client-33
    - logrotate-charm-5
    - nrpe-70
    - ovn-central-3
    - telegraf-40
  units:
    local:
      - filebeat-39
      - landscape-client-38
      - logrotate-24
      - nrpe-container-52
      - ovn-central-0
      - telegraf-26
  charm-repo-info:
    ovn-central-0: '7792718'
```

I think we could improve this in three ways:

1. The OVN units and their number is critical info I am always using in cases. It would be nice to have this list higher up in the output somehow. It's not clear hwo the current output is sorted, it's not alphabetical.
2. Modify the local charm unit output to use the canonical 'unit/N' format, e.g. 'ovn-central/0' to line up with the output from 'juju status' and all other juju CLI commands
3. Optional and maybe harder: The charm revision number is somewhat unhelpful at this point in the realm of charmhub - while it does map back to a version on charmhub we can retrieve, if we can improve this output to show something more detailed that may also help - e.g. 'filebeat (v24, channel usurri/stable)' or something to that effect. This would also improve the visual distinguishment.

PR #508, opened 2023-01-12
Rework how Juju unit info is displayed

dosaboy:
Resolves: #437


lathiat (approved):
LGTM

dosaboy:
as discussed with trent i will make a final change to remove the concept of lxd vs. local units since lxd units are no longer easy to see from juju with recent versions and in any case it was always somewhat of a hack.

PR #514, opened 2023-02-01
Fix keystone ldap check for LP 1896125

dosaboy:
python3-keystone is installed as a dep for all openstack services so we need to check for package "keystone" to know this is a keystone host and we also now check that ldap is actually in use.

### DIFF
--- a/defs/scenarios/openstack/keystone/bugs.yaml
+++ b/defs/scenarios/openstack/keystone/bugs.yaml
@@ -1,7 +1,7 @@
 checks:
   has_1896125:
     apt:
-      python3-keystone:
+      keystone:
         # Train
         - min: 2:16.0.0
           max: 2:16.0.1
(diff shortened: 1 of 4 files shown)


defs/scenarios/openstack/keystone/bugs.yaml
```diff
@@ -11,6 +11,14 @@ checks:
         # Victoria
         - min: 2:18.0.0
           max: 2:18.0.0-99
+    config:
+      handler: hotsos.core.plugins.openstack.OpenstackConfig
+      path: etc/keystone/keystone.conf
```
mustafakemalgilor, line 16: The [ldap] option can be turned on per-domain basis [1], so if we want this to work properly we also need to check all .conf files under `/etc/keystone/domains/` as well.

[1] See `Additional LDAP integration settings` section at https://docs.openstack.org/keystone/zed/admin/configuration.html#identity-ldap-server-set-up

mustafakemalgilor: The domain names can be arbitrary so I'm unsure if doing this is possible right now, we would need to allow wildcards in the path.

dosaboy: Good catch! I have now added support to do this and updated the patch.

dosaboy: ugh just realised my domain config path is not valid, ill fix that now

dosaboy: done.

mustafakemalgilor (approved):

PR #523, opened 2023-02-10
ycheck: Add PackageCheckItems class

joalif:
PackageCheckItems inherits CheckItemBase. From now on APTCheckItems SnapCheckItems inherit from PackageCheckItems.


hotsos/core/ycheck/engine/properties/requires/common.py
```diff
@@ -49,6 +50,33 @@ def __iter__(self):
             yield item
 
 
+class PackageCheckItems(CheckItemsBase):
```
dosaboy, line 53: I would name this PackageCheckItemsBase

hotsos/core/ycheck/engine/properties/requires/common.py
```diff
@@ -49,6 +50,33 @@ def __iter__(self):
...
+
+    @cached_property
+    def packages_to_check(self):
+        return [item[0] for item in self]
+
+    @cached_property
```
dosaboy, line 59: this should be an abc.abstractproperty but to do this we'll need to change:

class CheckItemsBase(object):

to be 

class CheckItemsBase(abc.ABC):

dosaboy: can we also rename info() to packaging_helper()

dosaboy (changes requested):

dosaboy:
lgtm but can you please squash the last 3 commits into one so that we have one commit for "Add PackageCheckItemsBase ..." and one for the lxd check.

dosaboy:
ah we also need to update defs/README.md to reflect that fact that apt and snap checks support the same features. Im happy to do that as a separate commit if you want.

dosaboy:
the snap build error is not related to this branch so i will ignore here

PR #528, opened 2023-02-24
Tell git to ignore .swp files

joalif:


.gitignore
```diff
@@ -5,6 +5,7 @@ __pycache__
 .vscode
 *.pyc
 *.snap
+*.swp
```
brianphaley, line 8: Should probably be *.sw? to cover all possible extensions, for example if you edit the same file twice it will be .swp and .swo

brianphaley (changes requested):

dosaboy:
ill land this and we can adjust the expression if we need to cover further types, seems that .swp will be good for most cases tho

PR #534, opened 2023-03-02
Ignore comment part from sysctl values

pponnuvel:


dosaboy:
any chance of a unit test?

pponnuvel:
> any chance of a unit test?

Added a comment in one of the existing fake data roots. Without this fix, the tests would fail with that comment. Is that sufficient?

Issue #539, opened 2023-03-09
juju charm versions are incorrect


dosaboy:
If a charm has an install history of more than one version, the most recent version detected is not always correct. This is because the versions are not being sorted properly e.g.

$ ls /var/lib/juju/agents/unit-ovn-chassis-7/state/bundles/
cs_3a_ovn-chassis-12  cs_3a_ovn-chassis-14  cs_3a_ovn-chassis-17  cs_3a_ovn-chassis-21  cs_3a_ovn-chassis-7

should always return version=21 but does not. instead it is dependent on order in which the files are returned by glob which seems random.

PR #540, opened 2023-03-09
Fix juju charm version

dosaboy:
Resolves: #539

### DIFF
--- a/tests/unit/test_juju.py
+++ b/tests/unit/test_juju.py
@@ -1,8 +1,12 @@
+import os
+import shutil
+
 from unittest import mock
 
 from . import utils
 
 from hotsos.core.config import HotSOSConfig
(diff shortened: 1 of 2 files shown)


tests/unit/test_juju.py
```diff
@@ -13,6 +17,34 @@ def setUp(self):
         HotSOSConfig.plugin_name = 'juju'
 
 
+class TestJujuResources(JujuTestsBase):
+
+    def test_charm(self):
+        charms = JujuBase().charms
+        self.assertEqual(list(charms.keys()),
+                         ['nova-compute', 'neutron-openvswitch', 'ceph-osd'])
```
brianphaley, line 25: Could use any_order=True or sort I guess.

Issue #541, opened 2023-03-11
cache cli commands


dosaboy:
hotsos is 100% read-only so there is no need to execute cli commands more than once. we should therefore cache cli command output to avoid repeated executions which can also lead to unnecessary system or cluster load e.g. ceph commands.

PR #543, opened 2023-03-13
Enable cli binary command caching

dosaboy:
Repeated execution of binary commands can lead to unecessary load of the host or cluster so we now cache their output and read from a file copy for successive calls to the same command. We currently only cache commands that are executed without args.

Resolves: #541


mustafakemalgilor (approved):
LGTM :shipit:

Issue #544, opened 2023-03-14
support pypi builds


dosaboy:
As an alternative to using snaps we can also use pypi to pip install hotsos.

PR #546, opened 2023-03-14
Add pypi build

dosaboy:
Resolves: #544


nicolasbock:
@dosaboy I think I have a fix, but I am not allowed to push to this PR.

hotsos/cli.py
```diff
@@ -32,6 +33,11 @@ def get_repo_info():
         with open(repo_info) as fd:
             return fd.read().strip()
 
+    # pypi
+    with resources.path('hotsos', '.repo-info') as repo_info:
+        with open(repo_info) as fd:
+            return fd.read().strip()
```
nicolasbock, line 39: When we run the `functional` test the `.repo-info` file does not exist.

```suggestion
    try:
        with resources.path('hotsos', '.repo-info') as repo_info:
            with open(repo_info) as fd:
                return fd.read().strip()
    except FileNotFoundError:
        return "unknown"
```

nicolasbock (changes requested):

nicolasbock (approved):
This looks good :smile:

PR #549, opened 2023-03-17
Add GitHub workflow for Python package

nicolasbock:


.github/workflows/python-package.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - main
+  pull_request:
+    - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.9", "3.10"]
```
dosaboy, line 14: i don't think 39 is standard in ubuntu so maybe we can skip it to save time?

nicolasbock: Ah sure

.github/workflows/python-package.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Install dependencies
+        run: |
+          git rev-parse --short HEAD > hotsos/.repo-info
+          python -m pip install --upgrade pip
+          python -m pip install build
+          python -m build
+
+      - name: Publish a Python distribution to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        if: github.event_name == 'push' && github.ref_name == 'main' && matrix.python-version == '3.8'
```
nicolasbock, line 34: @dosaboy I added a few conditionals here so that this step should only be used at the right moment. A PR is not going to trigger this and I can't test it.

Whenever you have some time, please add the PyPi API token as `PYPI_API_TOKEN` in the repository secrets. That should allow this step here to work.

Issue #451, opened 2022-10-21
check /proc/net/netlink for drops and report them


aserdean:
bonus points for correlating the inodes with the process. 
usually it is expected the highest one to be ovs but it could happen for other subsytems i.e. libvirt

PR #550, opened 2023-03-22
net: implemented `lsof` and `/proc/net/netlink` parsing:

mustafakemalgilor:
this feature is a prerequisite of #451.

- added `lsof_bMnlP` to CLIHelper command catalog
- implemented a generic, regex-based structured table of values parser (STOVParser)
- unit tests for `Lsof` and `NetLink`

### DIFF
--- a/hotsos/core/plugins/kernel/net.py
+++ b/hotsos/core/plugins/kernel/net.py
@@ -1,9 +1,11 @@
 import abc
 import os
-
+import re
+import itertools
+from typing import Iterator
 from hotsos.core.log import log
 from hotsos.core.config import HotSOSConfig
-from hotsos.core.host_helpers import SYSCtlFactory
+from hotsos.core.host_helpers import SYSCtlFactory, CLIHelper
 
 
 class ProcNetBase(abc.ABC):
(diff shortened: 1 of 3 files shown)


nicolasbock (approved):

hotsos/core/plugins/kernel/net.py
```diff
@@ -394,3 +396,239 @@ def __getattr__(self, fld):
 
         raise AttributeError("{} not found in {}".
                              format(fld, self.__class__.__name__))
+
+
+class STOVParser(abc.ABC):
```
dosaboy, line 401: I really like what this is doing and thing that it is some that could be generally useful. I am currently exploring how we could have FileSearcher perform these mappings natively so that e.g. when you go to retrieve your search results you can just retrieve them by (column) name rather than get(1) etc. I will report back asap on whether ive managed to get this to work and if it makes sense to just land this and adjust later we can do that too.

dosaboy: searchkit now supports naming results fields but I'll go ahead and land this with the aim of proposing a switch to searchkit as a separate patch.

PR #552, opened 2023-03-23
Implement build versions

dosaboy:
Implements build versions. Both snap and pypi builds will use a version generated from git tag. Since the snap is typically built on each push, the version includes a count of the number of commits since the last tagged release.

### DIFF
--- a/.github/workflows/snapbuild.py
+++ b/.github/workflows/snapbuild.py
@@ -1,0 +1,20 @@
+on:
+  workflow_run:
+    workflows: [Test]
+    branches: [main]
+    types: [completed]
+name: Build and publish Snap
+jobs:
+  build:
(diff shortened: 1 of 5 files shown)


.github/workflows/snapbuild.py
```diff
@@ -0,0 +1,20 @@
...
+    workflows: [Test]
+    branches: [main]
+    types: [completed]
+name: Build and publish Snap
+jobs:
+  build:
+    if: ${{ github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
nicolasbock, line 12: You can use `checkout@v3`

nicolasbock (approved):

PR #554, opened 2023-03-23
Cleanup of snapcraft.yaml

nicolasbock:


dosaboy:
this looks like it is more than just a cleanup?

nicolasbock:
It's so much cleaner now 😉 

The main reason for those changes is that the PR switches from using the `python` plugin to the `dump` plugin (so that we can `rsync` the `defs` and `templates`). Once we do that we need to build the Python package ourselves in the override.

dosaboy:
> It's so much cleaner now wink
> 
> The main reason for those changes is that the PR switches from using the `python` plugin to the `dump` plugin (so that we can `rsync` the `defs` and `templates`). Once we do that we need to build the Python package ourselves in the override.

any chance this could be reflected in the commit message?

nicolasbock:
I cleaned this up even more, going back to the original `python` plugin. But now the combination of `MANIFEST.in` and `organize` will install the `defs` and `templates` properly.

dosaboy:
aiui now that we have viable alternative methods of install we wont proceed with this. i'll close for now, please reopen if you would like continue.

PR #556, opened 2023-03-27
Fix Invalid cross-device link error moving logfile

lathiat:
If $TMPDIR and the target output directory are on a different
filesystem, os.rename() may fail with:
[Errno 18] Invalid cross-device link

Use shutil.move instead to support this scenario.


dosaboy:
lgtm

Issue #557, opened 2023-03-27
issue with ubuntu pro status output


dosaboy:
I'm getting the following:

    services:
      cis:
        entitled: 'yes'
        status: enabled
      esm-infra:
        entitled: 'yes'
        status: enabled
      fips-updates:
        entitled: 'yes'
        status: enabled
      A:
        entitled: change
        status: has

i.e. the last entry is scrambled

PR #558, opened 2023-03-27
system/ubuntu-pro: more robust `ua/pro status` parsing:

mustafakemalgilor:
Apparently, ubuntu pro status may contain `NOTICES` section, which disrupts the current parsing algorithm. This section is in between the service/account status sections. The new revised algorithm avoids reading the middle section by limiting the parser to first and last sections only.

Fixes: #557


lathiat:
I am thinking we could use "pro status --format json" instead?

It does say it's experimental and subject to change, but realisticlaly so is the CLI output. We could always reach out to the team about the json and it's likely change path.

mustafakemalgilor:
> I am thinking we could use "pro status --format json" instead?
> 
> It does say it's experimental and subject to change, but realisticlaly so is the CLI output. We could always reach out to the team about the json and it's likely change path.

We could, it would be more reliable but unfortunately "sos" does not capture it at the moment. We'd need to submit a patch upstream and wait until it lands everywhere. What I'm thinking is maybe we can add JSON handling as well and fall back to the present method if JSON output is not available.

dosaboy (commented):
Hi Mustafa, i've got a few tweaks to propose:

diff --git a/hotsos/core/plugins/system/system.py b/hotsos/core/plugins/system/system.py
index 96f53d2e..62f87d67 100644
--- a/hotsos/core/plugins/system/system.py
+++ b/hotsos/core/plugins/system/system.py
@@ -160,24 +160,22 @@ class SystemBase(object):
         # human-readable output for now. This function should ideally
         # rely on the json output when the upstream sos starts to
         # include it.
-        pro_status = CLIHelper().pro_status()
         s = FileSearcher()
         service_status_seqdef = SequenceSearchDef(
             start=SearchDef(r"^SERVICE +ENTITLED +STATUS +DESCRIPTION"),
-            body=SearchDef(r"^([a-zA-Z0-9-]+) +([a-zA-Z0-9-/]+)"
-                           r" +([a-zA-Z0-9-/]+) +(.+)\n"),
+            body=SearchDef(r"^(\S+) +(\S+) +(\S+) +(.+)\n"),
             end=SearchDef(r"\n"),
             tag="service-status")
         account_status_seqdef = SequenceSearchDef(
             start=SearchDef(r" *?(Account): (.*)\n"),
-            body=SearchDef(r" *?([A-Za-z0-9 ]+): (.*)\n"),
+            body=SearchDef(r" *?([\S ]+): (.*)\n"),
             end=SearchDef(r" *?(Technical support level): (.*)\n"),
             tag="account-status")
         not_attached_def = SearchDef(
             r".*not attached to.*(Ubuntu (Pro|Advantage)|UA).*",
             tag="not-attached")
 
-        f = mktemp_dump("".join(pro_status))
+        f = mktemp_dump("".join(CLIHelper().pro_status()))
         s.add(not_attached_def, f)
         s.add(service_status_seqdef, f)
         s.add(account_status_seqdef, f)
@@ -188,8 +186,8 @@ class SystemBase(object):
 
         ssects = results.find_sequence_sections(service_status_seqdef)
         asects = results.find_sequence_sections(account_status_seqdef)
-        if (not ssects) or (not asects):
-            log.debug("badness: `pro status` does not match"
+        if not all([ssects, asects]):
+            log.debug("badness: `pro status` does not match "
                       "the expected format")
             return {"status": "error"}
 
diff --git a/tests/unit/test_system.py b/tests/unit/test_system.py
index 51c5067c..abff29c5 100644
--- a/tests/unit/test_system.py
+++ b/tests/unit/test_system.py
@@ -217,7 +217,6 @@ class TestUbuntuPro(SystemTestsBase):
         mh.return_value.pro_status.return_value = UBUNTU_PRO_ATTACHED
         result = SystemBase().ubuntu_pro_status
         self.assertNotEqual(result, None)
-        print(result)
         self.assertNotEqual(result, False)
 
         expected_result = {

mustafakemalgilor:
> Hi Mustafa, i've got a few tweaks to propose:

Thanks, I've applied them :heavy_check_mark:

dosaboy:
lgtm, the snap build error is unrelated to this patch so i think we're good to land.

Issue #562, opened 2023-04-06
hotsos throws an exception if sos_commands/date/date does not exist


brianphaley:
If I run hotsos (by accident) using a directory that is not the root of sos-report output it throws an exception:

$ hotsos ./
INFO: analysing sosreport ./ /
Traceback (most recent call last):
  File "/usr/local/bin/hotsos", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/hotsos/cli.py", line 363, in main
    cli(prog_name='hotsos')
  File "/usr/lib/python3/dist-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/hotsos/cli.py", line 349, in cli
    client.run()
  File "/usr/local/lib/python3.10/dist-packages/hotsos/client.py", line 411, in run
    content = self._run(plugin)
  File "/usr/local/lib/python3.10/dist-packages/hotsos/client.py", line 394, in _run
    return plugintools.PluginRunner(parts).run()
  File "/usr/local/lib/python3.10/dist-packages/hotsos/core/plugintools.py", line 386, in run
    inst = cls()
  File "/usr/local/lib/python3.10/dist-packages/hotsos/plugin_extensions/openstack/vm_info.py", line 151, in __init__
    c = SearchConstraintSearchSince(exprs=[OPENSTACK_LOGS_TS_EXPR])
  File "/usr/local/lib/python3.10/dist-packages/hotsos/core/search.py", line 39, in __init__
    super().__init__(*args,
  File "/usr/local/lib/python3.10/dist-packages/searchkit/constraints.py", line 554, in __init__
    self.current_date = datetime.strptime(current_date, self.date_format)
TypeError: strptime() argument 1 must be str, not list

Under the covers it's looking for date output, but didn't find it.

It looks like there is simply a missing return statement in this case, I'll send a PR.

PR #563, opened 2023-04-06
Fix exception when incorrect date is encountered

brianphaley:
Add a return SearchConstraintSearchSince when an invalid date is encountered, else it will throw an exception. Closes issue #562.


brianphaley:
I don't know why I couldn't add issue 562 would be closed so it happens automatically (?)

PR #565, opened 2023-04-12
ceph-mon: Detect empty cluster log

lathiat:
Detect and report when the ceph cluster log /var/log/ceph/ceph.log is
empty, likely due to Ceph log rotation Bug #55383

https://tracker.ceph.com/issues/55383

### DIFF
--- a/hotsos/core/host_helpers/filestat.py
+++ b/hotsos/core/host_helpers/filestat.py
@@ -27,6 +27,16 @@ class FileObj(object):
         log.debug("mtime %s=%s", self.filename, mt)
         return mt
 
+    @property
+    def size(self):
+        if not os.path.exists(self.filename):
+            log.debug("size %s - file not found", self.filename)
+            return -1
+
+        size = os.path.getsize(self.filename)
+        log.debug("size %s=%d", self.filename, size)
+        return size
+
 
 class FileFactory(FactoryBase):
     """
--- a/hotsos/defs/scenarios/storage/ceph/ceph-mon/empty_clog.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-mon/empty_clog.yaml
@@ -1,0 +1,24 @@
+vars:
+  clog_size: '@hotsos.core.host_helpers.filestat.FileFactory.size:var/log/ceph/ceph.log'
+
+checks:
+  empty_clog_size:
+    varops: [[$clog_size], [eq, 0]]
+
+conclusions:
+  empty_clog:
+    decision:
+      - empty_clog_size
+    raises:
+      type: CephTrackerBug
+      bug-id: 55383
+      message: >-
+        The ceph.log file is empty, this is likely because the version of
+        ceph-mon on this node does not re-open
+        /var/log/ceph/{ceph.log,ceph.audit.log} after log rotation. This leaves
+        ceph.log empty and the logs are written to a deleted file. This will prevent
+        you from seeing the cluster log and hotsos checks for information from these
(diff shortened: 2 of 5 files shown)


lathiat:
I thought about an apt version check as well but decided against it, since this log has constant output it should never be empty. So there is likely something wrong if it is. Also packages can be upgraded but services are not restarted by default, so a version check may silence this warning even when it still applies in practice.

hotsos/core/plugins/storage/ceph.py
```diff
@@ -726,6 +726,14 @@ def osds_pgs_suboptimal(self):
 
         return sorted_dict(_osds_pgs, key=lambda e: e[1], reverse=True)
 
+    @cached_property
```
dosaboy, line 729: You could use https://github.com/canonical/hotsos/blob/main/hotsos/core/host_helpers/filestat.py to do this. You would need to extend FileObj() with a size property to get the same functionality but then it is nicely re-usable. For an example on how to call it from the scenario yaml see https://github.com/canonical/hotsos/blob/a7f8b1cb31b2bbaf80f73d2305be40a059cf5a26/hotsos/defs/scenarios/openvswitch/ovn/ovn_chassis_ssl_certs.yaml#L2

hotsos/core/host_helpers/filestat.py
```diff
@@ -20,6 +20,16 @@ def mtime(self):
         log.debug("mtime %s=%s", self.filename, mt)
         return mt
 
+    @property
+    def size(self):
+        if not os.path.exists(self.filename):
+            log.debug("mtime %s - file not found", self.filename)
```
dosaboy, line 26: log msg should say "size" not "mtime"

hotsos/core/host_helpers/filestat.py
```diff
@@ -20,6 +20,16 @@ def mtime(self):
...
+            return -1
+
+        size = os.path.getsize(self.filename)
+        log.debug("mtime %s=%d", self.filename, size)
```
dosaboy, line 30: same as above

hotsos/defs/tests/scenarios/storage/ceph/ceph-mon/empty_clog.yaml
```diff
@@ -0,0 +1,46 @@
+target-name: empty_clog.yaml
```
dosaboy, line 1: target-name isnt needed if the filename is identical. doesn't hurt tho of course.

hotsos/defs/scenarios/storage/ceph/ceph-mon/empty_clog.yaml
```diff
@@ -0,0 +1,24 @@
...
+    raises:
+      type: CephTrackerBug
+      bug-id: 55383
+      message: >-
+        The ceph.log file is empty, this is likely because the version of
+        ceph-mon on this node does not re-open
+        /var/log/ceph/{ceph.log,ceph.audit.log} after log rotation. This leaves
+        ceph.log empty and the logs are written to a deleted file. This will prevent
+        you from seeing the cluster log and hotsos checks for information from these
+        logs will not trigger. This issue is fixed in 17.2.5. You can work around
```
dosaboy, line 21: is this a 17.2.x only bug? Should the version referenced here be tuned depending on what version/release is actually in use?

lathiat: Yeah it's only in 17.2.x, wasn't in 16.x so should be OK to just say 17.2.5

PR #566, opened 2023-04-12
Parse datetimes as UTC

lathiat:
Parse datetimes as UTC so that they are read consistently on different
systems. Without this, in some timezones (e.g. GMT+8 after 2PM) running
the test suite results in a failure because ceph is EOL in 2999 days
instead of 3000 days.


dosaboy:
hmm im not sure why it isn't already UTC since the date command is doing --utc - https://github.com/canonical/hotsos/blob/main/hotsos/core/host_helpers/cli.py#L414

dosaboy:
or is it that datetime.fromtimestamp() is converting to local TZ?

dosaboy:
confirmed, the issue is with datetime.fromtimestamp so this patch is good

PR #570, opened 2023-04-17
[kernel] Detect hung tasks in kernel logs

### LINKED ISSUE #198: warn if any task has been blocked for more than 120 seconds in kern.log
Warn in hotsos if any task has been blocked for longer than hung_task_timeout.

Entire call traces might be too long or bulky to print in hotsos, but hotsos could output something like "XXX process has been blocked for more than 120 seconds" if seen in the latest boot? (we should ignore these messages in older boots, I guess?)

nkshirsagar:
Fixes #198

### DIFF
--- a/hotsos/defs/tests/scenarios/kernel/kernlog_calltrace_hungtasks.yaml
+++ b/hotsos/defs/tests/scenarios/kernel/kernlog_calltrace_hungtasks.yaml
@@ -1,0 +1,37 @@
+target-name: kernlog_calltrace.yaml
+data-root:
+  files:
+    var/log/kern.log: |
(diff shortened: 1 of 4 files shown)


hotsos/defs/tests/scenarios/kernel/kernlog_calltrace_hungtasks.yaml
```diff
@@ -0,0 +1,37 @@
+target-name: kernlog_calltrace_hungtasks.yaml
```
dosaboy, line 1: this should be:

target-name: kernlog_calltrace.yaml

nkshirsagar: thank you!

dosaboy:
lgtm thanks nikhil!

PR #574, opened 2023-04-20
Convert STOV to searchkit

dosaboy:
The STOV implementations in plugins.kernel.net can be achieved by using searckit directly with named fields. This approach allows us to use existing code and removes the need for a custom library.


mustafakemalgilor (approved):
LGTM :shipit:

PR #595, opened 2023-05-05
os-exceptions: perform ost warn/exceptions search with hyperscan

xmkg:
this is a proof-of-concept work. using "hyperscan" improves the runtime performance of hotsos by 5.32x(*) whilst keeping the functionality intact. 

before: 36m24,760s
after: 6m50,791s

(*): in a known large sosreport file


mustafakemalgilor:
see https://github.com/dosaboy/searchkit/pull/6

PR #598, opened 2023-05-08
Add microk8s support to the kubernetes plugin.

dnegreira:
This patch aims to add support for microk8s to the kubernetes plugin. Hotsos will now display the containers/pods running in the system as well as which microk8s snap version is installed.

Closes: #582

### DIFF
--- a/hotsos/core/plugins/kubernetes.py
+++ b/hotsos/core/plugins/kubernetes.py
@@ -18,6 +18,7 @@ SERVICES = [r"etcd\S*",
             r"dockerd\S*",
             r"kubelet\S*",
             r"kube-\S*",
+            r"microk8s\S*",
             ]
 
 # Packages that only exist in a K8s deployment
@@ -29,6 +30,7 @@ K8S_PACKAGES = [r'cdk-addons',
                 r'kubelet',
                 r'kubeadm',
                 r'kubefed',
+                r'microk8s'
                 ]
 # Packages that are used in a K8s deployment
 K8S_PACKAGE_DEPS = [r'charm[\S]+',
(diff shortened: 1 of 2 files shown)


hotsos/core/plugins/kubernetes.py
```diff
@@ -110,10 +113,15 @@ def __init__(self, *args, **kwargs):
...
+        self.microk8s = SnapPackageHelper(core_snaps=MICROK8S_SNAP)
+        if self.microk8s.core:
+            microk8s_deps = snap_deps + MICROK8S_SNAP
+            self.snaps = SnapPackageHelper(core_snaps=K8S_PACKAGES,
+                                           other_snaps=microk8s_deps)
 
     @property
     def plugin_runnable(self):
-        if self.apt.core or self.snaps.core:
+        if self.apt.core or self.snaps.core or self.microk8s.core:
```
dosaboy, line 124: is there a benefit to treating the microk8s snap separately to the other snaps? this will result in two lookups and if it has not intended benefit i would prefer we included microk8s with the other snaps.

dosaboy:
im also wondering if we should open a ticket on sosreport to start collecting microk8s information in a similar way to the existing [kubernetes](https://github.com/sosreport/sos/blob/main/sos/report/plugins/kubernetes.py) plugin so that we can start analysing it here and have parity a between host-run and sosreport-run

hotsos/core/plugins/kubernetes.py
dnegreira: My only reasoning was that in the future if we want to separate checks between microk8s / k8s we would have a way to have it done.
I agree it becomes cleaner/easier to just add the microk8s with the other snaps, and making the code change in the future when necessary is also easy, I will fix this.

dnegreira: @dosaboy fixed on https://github.com/canonical/hotsos/commit/9dcfaef97f9a1735242377ffc426e091986d47b1

PR #599, opened 2023-05-10
[ceph] Warn on LP#2013960

### LINKED ISSUE #591: Warn on LP#2013960
All Quincy releases upto 17.2.5 are affected by this bug. Affects only Quincy (17.2.x).

The impact is that whenever any recovery happens, it could affect client I/O. More the recovery greater the impact.

https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/2013960

Fix is in 17.2.6. Recommend to use 'wpq' (osd_op_queue) in the meantime.

pponnuvel:
Fixes #591.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/ceph-osd/bugs/lp2013960.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-osd/bugs/lp2013960.yaml
@@ -1,0 +1,17 @@
+checks:
+  node_is_ceph_osd_and_has_version:
+    apt:
+      ceph-osd:
+        - min: 17.2.0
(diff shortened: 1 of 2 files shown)


dosaboy:
@pponnuvel the reason this is failing is because your lp check says 17.2.5 is max and the test has 17.2.5-0ubuntu0.22.04.2 which is > 17.2.5. I would suggest using the actual last package version that has the bug as the max version

pponnuvel:
> @pponnuvel the reason this is failing is because your lp check says 17.2.5 is max and the test has 17.2.5-0ubuntu0.22.04.2 which is > 17.2.5. I would suggest using the actual last package version that has the bug as the max version

Thanks.

PR #600, opened 2023-05-10
[CLI helper] switch filesys/lsof to process/lsof

pponnuvel:
filesys/lsof isn't always collected, so using process/lsof instead.

### DIFF
--- a/hotsos/core/host_helpers/cli.py
+++ b/hotsos/core/host_helpers/cli.py
@@ -778,9 +778,9 @@ class CLIHelper(HostHelpersBase):
             'lscpu':
                 [BinCmd('lscpu'),
                  FileCmd('sos_commands/processor/lscpu')],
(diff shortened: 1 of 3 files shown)


pponnuvel:
@mustafakemalgilor @dosaboy 

`filesys/lsof` isn't typically collected as part of sosreports; rather `process/lsof` is always there.
So I've modified to use the latter. The difference is that the latter doesn't have TID and TASKCMD fields.
I am not sure if these fields are needed to implement something in hotsos (lsof isn't actually used anywhere yet, so I couldn't confirm). If those two fields are needed, then we probably will have work with upstream to get those.
Otherwise please check if this PR is OK.

mustafakemalgilor:
`lsof` parsing is chiefly implemented for issue #451. For now, we don't rely on TID/TASKCMD so it should be fine.

Issue #596, opened 2023-05-07
Detect repeated ovn leader elections


nicolasbock:
We have seen a few cases now where `ovn` decides to frequently run leader
elections. The log files will include lines such as

```
[sosreport-juju-23df31-39-lxd-25-00359661-2023-05-06-owhfskk]/ovsdb-server-nb.log     |2023-05-06T17:28:09.803Z|21486|raft|INFO|term 3090: 1083 ms timeout expired, starting election
[sosreport-juju-23df31-38-lxd-26-00359661-2023-05-06-vsebghl]/ovsdb-server-nb.log     |2023-05-06T17:28:19.657Z|21393|raft|INFO|term 3091: 1024 ms timeout expired, starting election
[sosreport-juju-23df31-39-lxd-25-00359661-2023-05-06-owhfskk]/ovsdb-server-nb.log     |2023-05-06T17:29:00.454Z|21547|raft|INFO|term 3097: 1132 ms timeout expired, starting election
[sosreport-juju-23df31-40-lxd-25-00359661-2023-05-06-tbilkzx]/ovsdb-server-nb.log     |2023-05-06T17:29:14.288Z|17791|raft|INFO|term 3098: 1436 ms timeout expired, starting election
[sosreport-juju-23df31-38-lxd-26-00359661-2023-05-06-vsebghl]/ovsdb-server-nb.log     |2023-05-06T17:29:23.328Z|21476|raft|INFO|term 3099: 1091 ms timeout expired, starting election
[sosreport-juju-23df31-40-lxd-25-00359661-2023-05-06-tbilkzx]/ovsdb-server-nb.log     |2023-05-06T17:29:54.265Z|17840|raft|INFO|term 3103: 1312 ms timeout expired, starting election
[sosreport-juju-23df31-39-lxd-25-00359661-2023-05-06-owhfskk]/ovsdb-server-sb.log     |2023-05-06T17:29:57.105Z|03617|raft|INFO|term 185: 1685 ms timeout expired, starting election
```

`hotsos` should flag this issue and suggest remedies:

```
sosreport:
  potential-issues:
    multiple ovsdb elections detected. This can be caused by insufficient polling timeout settings. Verify that....
openvswitch:
  ovn-checks:
    ovsdb-server-nb:
      started-election:
        '2023-05-06': 12345
```

PR #603, opened 2023-05-16
Add support for ovn election messages

nicolasbock:
Closes canonical/hotsos#596

### DIFF
--- a/hotsos/defs/scenarios/openvswitch/ovn/ovn_elections.yaml
+++ b/hotsos/defs/scenarios/openvswitch/ovn/ovn_elections.yaml
@@ -1,0 +1,51 @@
+checks:
+  sb_ovn_elections:
+    input:
+      path:
+        - var/log/ovn/ovsdb-server-sb.log
+    search:
+      expr: '([0-9-]+)T([0-9:\.]+)Z.+\|raft\|INFO\|term ([0-9]+):\s(([0-9]+) ms timeout expired, )?starting election$'
+      constraints:
+        min-results: 5
+        search-period-hours: 1
+        search-result-age-hours: 48
+  nb_ovn_elections:
+    input:
+      path:
+        - var/log/ovn/ovsdb-server-nb.log
+    search:
+      expr: '([0-9-]+)T([0-9:\.]+)Z.+\|raft\|INFO\|term ([0-9]+):\s(([0-9]+) ms timeout expired, )?starting election$'
+      constraints:
+        min-results: 5
+        search-period-hours: 1
+        search-result-age-hours: 48
+conclusions:
+  ovn_frequent_elections_sb:
+    priority: 1
+    decision: sb_ovn_elections
+    raises:
+      type: OVNWarning
+      message: >-
+        The OVN SB database is experiencing frequent raft elections. This can
+        have an impact on the performance of the cloud and might indicate an
+        issue with the ovn-central cluster. Please investigate.
+  ovn_frequent_elections_nb:
+    priority: 1
+    decision: nb_ovn_elections
+    raises:
+      type: OVNWarning
+      message: >-
+        The OVN NB database is experiencing frequent raft elections. This can
+        have an impact on the performance of the cloud and might indicate an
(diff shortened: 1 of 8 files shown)


dosaboy:
i think a scenario would be a better fit than an event and this is for two main reasons; first you can raise issues from scenarios (which is what we need to do in this patch) and also you can write your tests in yaml which means you don't need to clobber the fake_data_root files like you have here. A good example to follow is https://github.com/canonical/hotsos/blob/main/hotsos/defs/scenarios/storage/ceph/ceph-mon/mon_elections_flapping.yaml. There are also some changes in this patch that dont relate to the issue at hand, perhaps they could be submitted seperately.

hotsos/defs/tests/scenarios/openvswitch/ovn/ovn_elections.yaml
```diff
@@ -0,0 +1,51 @@
...
+      2023-05-06T16:59:59.892Z|16372|raft|INFO|term 2917: 1274 ms timeout expired, starting election
+      2023-05-06T17:00:25.937Z|19934|raft|INFO|term 2919: 1143 ms timeout expired, starting election
+      2023-05-06T17:00:55.262Z|16396|raft|INFO|term 2921: 1050 ms timeout expired, starting election
+      2023-05-06T17:01:05.290Z|20107|raft|INFO|term 2922: 1266 ms timeout expired, starting election
+      2023-05-06T17:01:21.685Z|19956|raft|INFO|term 2923: 1338 ms timeout expired, starting election
+      2023-05-06T17:01:58.065Z|20142|raft|INFO|term 2926: 1154 ms timeout expired, starting election
+      2023-05-06T17:02:14.646Z|19988|raft|INFO|term 2927: 1204 ms timeout expired, starting election
+      2023-05-06T17:02:40.497Z|16430|raft|INFO|term 2929: 1258 ms timeout expired, starting election
+      2023-05-06T17:03:11.530Z|20024|raft|INFO|term 2931: 1175 ms timeout expired, starting election
+      2023-05-06T17:03:24.486Z|16460|raft|INFO|term 2932: 1226 ms timeout expired, starting election
```
dosaboy, line 48: as it is this will fail because under-the-hood the search constraints you have a applied are saying logs must be 48 hours old and that will use the CLIHelper.date output which will be None since your data_root does not have a sos_commands/date/date. So you have two options; either do

````
data-root:
....
  copy-from-original:
    sos_commands/date/date
```

which will give you "Thu Feb 10 16:19:17 UTC 2022" and therefore you will need to adjust the timestamps in your logs to be within 48 hours of this or you can create a fake one e.g.:

```
data-root:
  files:
    sos_commands/date/date: |
      Thu May 07 16:19:17 UTC 2023
```

brianphaley (approved):

brianphaley (approved):

hotsos/defs/scenarios/openvswitch/ovn/ovn_elections.yaml
```diff
@@ -0,0 +1,21 @@
+checks:
+  ovn_elections:
```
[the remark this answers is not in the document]
nicolasbock: done

hotsos/defs/scenarios/openvswitch/ovn/ovn_elections.yaml
```diff
@@ -0,0 +1,51 @@
...
+    search:
+      expr: '([0-9-]+)T([0-9:\.]+)Z.+\|raft\|INFO\|term ([0-9]+):\s(([0-9]+) ms timeout expired, )?starting election$'
+      constraints:
+        min-results: 5
+        search-period-hours: 1
+        search-result-age-hours: 48
+  nb_ovn_elections:
+    input:
+      path:
+        - var/log/ovn/ovsdb-server-nb.log
```
dosaboy, line 15: nit: dont use list for one item

dosaboy (commented):
lgtm! couple tiny nits to fix

PR #604, opened 2023-05-17
Add OVN cert expiry check

dosaboy:
Resolves: #325


dosaboy:
needs a couple of adjustments so will have a new patch up in a min

dosaboy:
should be good to go now

nicolasbock (approved):

PR #610, opened 2023-05-22
[ceph] Warn if mon/osd is using leveldb.

### LINKED ISSUE #343: Detect ceph-[mon|osd] backend
Report whether ceph-mon is using leveldb or rocksdb.

leveldb is the older one used by filestore deployments. Since filestore is deprecated in Quincy, we probably need to "suggest" that the migration of to rocksdb too.

pponnuvel:
Fixes #343.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/ceph-osd/meta_backend_osd.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-osd/meta_backend_osd.yaml
@@ -1,0 +1,21 @@
+checks:
+  ceph_is_luminous_or_newer:
+    apt:
+      ceph-osd:
+        - min: 12.2.0
+  backend_type:
+    input:
+      path: var/lib/ceph/osd/*/kv_backend
+    expr: 'leveldb'
+conclusions:
+  deprecated_type:
+    decision:
+      - ceph_is_luminous_or_newer
+      - backend_type
+    raises:
+      type: CephMonWarning
+      message: >-
+        One or more OSDs are using 'leveldb' backend. 'rocksdb' has been
+        the default backend since Luminous release for performance reasons and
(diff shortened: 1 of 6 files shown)


hotsos/defs/scenarios/storage/ceph/ceph-osd/meta_backend_osd.yaml
```diff
@@ -0,0 +1,21 @@
...
+    decision:
+      - ceph_is_luminous_or_newer
+      - backend_type
+    raises:
+      type: CephMonWarning
+      message: >-
+        One more more OSDs are using 'leveldb' backend. 'rocksdb' has been
+        the default backend since Luminous release for performance reasons and
+        'leveldb' support has been removed since Quincy release. Recommended
+        converting the Ceph OSDs to use 'rocksdb' backend.
```
dosaboy, line 21: Can we tighten the wording a little like:

        One more more OSDs are using 'leveldb' backend. 'rocksdb' has been
        the default backend since Luminous release for performance reasons and
        'leveldb' support has been removed since Quincy release. Recommendation
        is to convert the Ceph OSDs to use 'rocksdb' backend.

And what would be ideal is if we could link to a KB article on how to do this. If one does not exist we can come back to it later.

pponnuvel: We do have a KB. So I simply add the SF link in the message here?

pponnuvel: I added a KB link for ceph mon conversion. We don't have one for OSD conversion.

pponnuvel: Replaced with "ua" link (KB).

PR #611, opened 2023-05-22
[ceph] Warn if SSD OSDs are using bcache

### LINKED ISSUE #573: Report if OSD backed by SSDs are using bcache
It's almost always not recommended to use bcache if the OSDs are using SSDs. 

Refer https://warthogs.atlassian.net/browse/ACT-2

So if SSD OSDs are using bcache, we should issue a warning.

This can be inferred from `ceph report` & `ceph osd df tree`. Both already available in hotsos' cached output.

pponnuvel:
It's nearly always the case that the SSDs can provide better performance on their own compared to a bcache device which is typically shared between several OSDs. Bcache usage is mainly meant when using OSDs are HHDs. So this is likely to be a misconfiguration and recommend removing the bcache device(s).

Fixes #573.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/ceph-mon/ssds_using_bcache.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-mon/ssds_using_bcache.yaml
@@ -1,0 +1,20 @@
+checks:
+  ssds_using_bcache:
+    property:
+      path: hotsos.core.plugins.storage.ceph.CephCluster.ssds_using_bcache
+      ops: [[length_hint], [gt, 0]]
+conclusions:
+  ssd_osds_using_bcache:
+    decision: ssds_using_bcache
+    raises:
+      type: CephOSDWarning
+      message: >-
+        Found the osd(s) {osds} using bcache and their underlying block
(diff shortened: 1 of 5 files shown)


hotsos/defs/scenarios/storage/ceph/ceph-mon/ssds_using_bcache.yaml
```diff
@@ -0,0 +1,20 @@
...
+      message: >-
+        Found the osd(s) {osds_list} using bcache and their underlying block
+        device of these OSDS appear to be SSDs. It's nearly always the case
+        that the benefit of using bcache for SSD OSDs is nil and can even
+        adversely affect performance in some cases. So this is likely to be a
+        misconfiguration and it's probably better to remove the bcache and use
+        the OSDs directly instead. Please compare the IOPs of the SSD (OSDs)
+        vs. the bcache device (SSD/NVMe) to ascertain.
+      format-dict:
+        osds: hotsos.core.plugins.storage.ceph.CephCluster.ssds_using_bcache
```
dosaboy, line 20: key should be "osds_list"

Issue #609, opened 2023-05-19
apply search constraints to kernel.kernlog.calltrace


dosaboy:
For example, kern.log might span several days and if we use the default period of 24 hours and there was an OOM 5 days ago it will still get reported. We should therefore apply the same constraints that we do for scenarios etc.

PR #616, opened 2023-05-24
Kernlog search constraints

dosaboy:
Implement search constraints for kern.log checks. This patch makes use of the new searchkit.constraints.TimestampMatcherBase class to allow timestamps of any type/format to be matched and updates all users of SearchConstraintSearchSince to use this approach for consistency.

Resolves: #609

### DIFF
--- a/hotsos/core/plugins/kernel/kernlog/common.py
+++ b/hotsos/core/plugins/kernel/kernlog/common.py
@@ -1,13 +1,50 @@
 import abc
 import os
 
+from searchkit.constraints import TimestampMatcherBase
+
+from hotsos.core.log import log
 from hotsos.core.config import HotSOSConfig
 from hotsos.core.host_helpers import CLIHelper, HostNetworkingHelper
-from hotsos.core.search import FileSearcher
+from hotsos.core.search import FileSearcher, SearchConstraintSearchSince
 
 KERNLOG_TS = r'\[\s*\d+\.\d+\]'
 KERNLOG_PREFIX = (r'(?:\S+\s+\d+\s+[\d:]+\s+\S+\s+\S+:\s+)?{}'.
                   format(KERNLOG_TS))
+
+
+class KernLogTimestampMatcher(TimestampMatcherBase):
+    """
+    kern.log has a slightly esoteric timestamp format so we have to do a little
+    juggling to get it into a standard format that can be used with search
+    constraints.
+
+    NOTE: remember to update
+          hotsos.core.ycheck.engine.properties.search.CommonTimestampMatcher
+          if necessary.
+    """
+    MONTH_MAP = {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5,
+                 'jun': 6, 'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10,
+                 'nov': 11, 'dec': 12}
+
+    @property
+    def year(self):
+        """ Use current year as it is not normally included in the logs """
+        return CLIHelper().date(format='+%Y')
+
+    @property
+    def month(self):
+        _month = self.result.group('month').lower()
+        for sm, im in self.MONTH_MAP.items():
+            if _month.startswith(sm):
+                return im
+
+        log.error("could not establish month integer from '%s'", _month)
+
+    @property
+    def patterns(self):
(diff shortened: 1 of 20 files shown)


hotsos/core/plugins/storage/ceph.py
```diff
@@ -91,6 +91,14 @@ def csv_to_set_inner(*args, **kwargs):
     return csv_to_set_inner
 
 
+class CephTimestampMatcher(TimestampMatcherBase):
+
+    @property
+    def patterns(self):
+        return [r'^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})\s?T'
```
pponnuvel, line 98: Ceph logs include the milliseconds too. E.g. `2023-04-18T00:00:51.995+0000 7fdd55879700  4 rocksdb:`. But I think this pattern should work in any case.

IIRC, older log formato allowed `' '` (space) in place of `T`. So you need to change to cover either: `[ |T]`.

hotsos/core/ycheck/engine/properties/search.py
```diff
@@ -16,9 +18,31 @@
...
+              import them due to circular dependency issues.
+        """
+        # should match plugins.openstack.openstack.OpenstackDateTimeMatcher
+        openstack = (r'^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})+\s+'
+                     r'(?P<hours>\d{2}):(?P<minutes>\d{2}):(?P<seconds>\d+)')
+        # since they are identical we wont add but leaving in case we want to.
+        # edit later.
+        # juju = openstack
+        # should match plugins.storage.ceph.CephDateTimeMatcher
+        ceph = (r'^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})\s?T'
```
pponnuvel, line 40: As above.

hotsos/core/plugins/kernel/kernlog/common.py
```diff
@@ -1,15 +1,47 @@
...
+    """
+    kern.log has a slightly esoteric timetamp format so we have to do a little
+    juggling to get it into a standard format that can be used with search
+    constraints.
+    """
+
+    @property
+    def year(self):
+        """ Use current year as it is not normally included in the logs """
+        return CLIHelper().date(format='+%Y')
```
pponnuvel, line 26: This could be problematic towards the year ends (sosreport collected in Dec and hotsos ran in Jan) ;-)

Maybe, get the year from the sosreport itself, from `sos_commands/date/date` for example?

hotsos/core/plugins/kernel/kernlog/common.py
```diff
@@ -1,15 +1,47 @@
...
+
+    @property
+    def month(self):
+        _month = self.result.group('month').lower()
+        for sm, im in {'jan': 1, 'feb': 2, 'mar': 3, 'apr': 4, 'may': 5,
+                       'jun': 6, 'jul': 7, 'aug': 8, 'sep': 9, 'oct': 10,
+                       'nov': 11, 'dec': 12}.items():
+            if _month.startswith(sm):
+                return im
```
brianphaley, line 35: Just a question on this code. I know there are many ways to do this, and this works fine, but I was wondering of this was called a lot, and are we going to be iterating through this list a lot?  If so it could be tweaked to avoid that, something like:

months = {'jan': 1, ... } (define at top of file)
_month = self.result.group('month').lower()[:3]
try:
    return months[_month]
except KeyError:
    log.error(...)

hotsos/core/plugins/storage/ceph.py
dosaboy: ack have fixed

hotsos/core/plugins/kernel/kernlog/common.py
dosaboy: if running against a sosreporet CLIHelper().date(format='+%Y') is going to return the year from sos_commands/date/date so should be doing what you ask already.

hotsos/core/plugins/kernel/kernlog/common.py
dosaboy: in reality it will only be executed once per instance but i have tidied up a bit and moved to be a class attribute

brianphaley (approved):

PR #619, opened 2023-05-24
Do not use lists with only one element

nicolasbock:


hotsos/defs/tests/scenarios/openvswitch/ovn/ovn_elections_all.yaml
```diff
@@ -1,7 +1,6 @@
 target-name: ovn_elections.yaml
 data-root:
-  copy-from-original:
-    - sos_commands/date/date
+  copy-from-original: sos_commands/date/date
```
dosaboy, line 3: This does actually have to be a list so single element is fine

Issue #623, opened 2023-05-30
raise error if snap run on < focal


dosaboy:
The hotsos snap will execute on any version of ubuntu that it can be installed on but it will not work properly if the host OS version of python is < 3.8 because classic snaps will occasionally call outside of the snap even if they are built with a specific version of python. For example on Ubuntu Bionic this results in an attempt to use some py 3.6 dependencies which are not compatible e.g.

```
root@b-test:~# hotsos
Traceback (most recent call last):
  File "/snap/hotsos/737/bin/hotsos", line 5, in <module>
    from hotsos.cli import main
  File "/snap/hotsos/737/lib/python3.8/site-packages/hotsos/cli.py", line 9, in <module>
    from importlib import metadata, resources
ImportError: cannot import name 'metadata'
```

This is failing because it is trying to use importlib from py 3.6 i.e.

```
root@b-test:~# python3 --version
Python 3.6.9
root@b-test:~# python3
Python 3.6.9 (default, Mar 10 2023, 16:46:00) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from importlib import metadata, resources
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'metadata'
```

So we should throw an error saying unsupported rather than allowing this exception to happen.

PR #624, opened 2023-05-30
Check for OS if running snap

nicolasbock:
The snap does not run on anything not core20 (Focal).

Closes: https://github.com/canonical/hotsos/issues/623

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import click
 import contextlib
+import distro
 import os
 import sys
 import subprocess
@@ -17,6 +18,18 @@ from hotsos.client import (
     OutputManager,
     PLUGIN_CATALOG,
 )
+
+
+def is_snap():
+    return os.environ.get('SNAP_NAME', '') == 'hotsos'
+
+
+def get_os_id():
+    return distro.id()
+
+
+def get_os_version():
+    return float(distro.version())
 
 
 def get_hotsos_root():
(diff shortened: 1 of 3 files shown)


hotsos/cli.py
```diff
@@ -19,6 +20,18 @@
 )
 
 
+def is_snap():
+    return os.environ.get('SNAP_NAME', '') == 'hotsos'
+
+
+def get_os_id():
+    return distro.id() == 'ubuntu'
```
brianphaley, line 28: This should just return distro.id() as the caller is checking that value, not true/false.

brianphaley (changes requested):

hotsos/cli.py
nicolasbock: Oh right, good catch.

hotsos/cli.py
```diff
@@ -369,4 +382,11 @@ def cli(data_root, version, defs_path, templates_path, all_logs, quiet,
 
 
 if __name__ == '__main__':
+    if is_snap():
+        if get_os_id() != 'ubuntu':
+            print("This snap has only been verified to run on Ubuntu")
+            sys.exit(1)
+        if get_os_codename() not in ["focal"]:
+            print("This snap has only been verified to run on Focal")
+            sys.exit(2)
```
brianphaley, line 391: I thought of having --force override this, but probably best to just exit, or use pip :-)

brianphaley (approved):

hotsos/cli.py
dosaboy: we need to allow it work on >= focal. we test py3.8, py3.10 and py3.11 currently so it should be good for anything above and including focal.

nicolasbock: Fixed

nicolasbock:
@brianphaley and @dosaboy I reworked the code a bit and added unit tests. Please have another look.

brianphaley (approved):

PR #626, opened 2023-05-31
kernel/net: introduce new netlink scenario

mustafakemalgilor:
the new scenario checks for the dropped netlink packets and reports them with the associated processes, if any.

Closes: #451


pponnuvel (approved):
LGTM

hotsos/defs/scenarios/kernel/network/netlink.yaml
```diff
@@ -0,0 +1,16 @@
...
+  netlink-socks-with-drops:
+    decision: has_socks_with_drops
+    raises:
+      type: NetworkWarning
+      message: >-
+        This host is reporting dropped packets for the following netlink sockets:
+        
+        {socks_with_drops}
+        
+        This maybe a symptom for the problems in the associated process(es).
```
brianphaley, line 14: nit: "... symptom of problems with the following ..."

dosaboy:
hi @mustafakemalgilor im finally revisiting this and would like to get it landed. Unfortunately it now fails pylint due to new rules that are in effect:

```
************* Module hotsos.core.plugins.kernel.net
hotsos/core/plugins/kernel/net.py:623:8: W0622: Redefining built-in 'list' (redefined-builtin)
```

I propose applying the following to your patch (rebased) to resolve this and other comments - 
[diff.txt](https://github.com/canonical/hotsos/files/12144363/diff.txt)

PR #633, opened 2023-06-09
Create nova libvirt interface class

dosaboy:
Move code used to extract nova guest information from libvirt info the core openstack plugin so that it is easier to consume from other places e.g. yaml checks.

### DIFF
--- a/hotsos/core/plugins/openstack/nova.py
+++ b/hotsos/core/plugins/openstack/nova.py
@@ -1,6 +1,7 @@
 import os
 import re
 
+from hotsos.core.plugins.kernel.sysfs import CPU
 from hotsos.core.config import HotSOSConfig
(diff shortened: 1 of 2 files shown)


hotsos/core/plugins/openstack/nova.py
```diff
@@ -1,6 +1,7 @@
 import os
 import re
 
+from hotsos.core.plugins.kernel.sysfs import CPU
```
brianphaley, line 4: I guess we don't enforce alphabetical ordering, or some other things, with pep8, it's tree-wide so not just here. I had been looking at some pep8/pylint cleanups so can add some of this there.

dosaboy: we don't but i'm all for it though so welcome pylint tweaks to enforce it

PR #639, opened 2023-06-13
Correctly cleanup temp directories

brianphaley:
The BaseTestCase class, when called from the
test_config.TestHotSOSConfig.test_restore_defaults() test, was not cleaning up all the temp directories it had created. This is because it was over-writing the instance variable self.global_tmp_dir.

Just re-use the directory if it already exists.


dosaboy:
The issue is that test_restore_defaults is calling setUp() from within the test itself so this will fix that. It is crucial that each test starts with a fresh global_tmp_dir but i checked and they still do with this patch so all good.

brianphaley:
Right - my first comment saved a list of all the temp dirs and then removed them all on exit. Since it was just this one test calling setUp() twice this seemed the easier fix.

Issue #643, opened 2023-06-14
add check for excessive libvirtd memory usage


dosaboy:
there have been some reports from the field about libvirtd consuming more memory than usual, sometimes steadily growing over time into the 100s of GBs which would clearly indicate a memory leak. We can easily check how memory is being used in hotsos so lets add a check and warn if usage goes above e.g. 5G which should be plenty of memory.

dosaboy:
see https://github.com/canonical/hotsos/blob/main/hotsos/defs/scenarios/openvswitch/ovn/service_mem_usage.yaml for an example

PR #644, opened 2023-06-14
Check libvirtd memory usage

tpsilva:
Added a check that raises an OpenstackWarning if libvirtd memory usage is > 5G.

Resolves: #643


brianphaley (approved):
lgtm

PR #647, opened 2023-06-15
Fix pep8 I100, I201 and I202 warnings

brianphaley:
This fixes basic import ordering and grouping issues and starts enforcing them.


dosaboy:
patch looks mostly good. i went and did a final sweep to ensure consistency with groups and here is my proposed change:
[patch.diff.txt](https://github.com/canonical/hotsos/files/11801057/patch.diff.txt)

brianphaley:
Hey Ed - thanks for the diff. The only problem is that breaks the flake8 H306 check :(  I actually have a follow-on that fixes things and enables that, which will get one step closer to correct ordering.

Issue #642, opened 2023-06-14
make snap confined and supporting sosreport only


dosaboy:
The snap is currently built in classic mode so that it can be run on both a host or against a sosreport but this has led to problems e.g. when using the python3-cryptography lib. Rather than have an error prone snap that sort of works for host/sosreport it would be preferable to have work well for one at the expense of the other. Therefore going forward we will make the snap strictly confined in order to alleviate python lib issues and therefore only support using the snap with sosreports. If a user attempts to use the snap on a host (i.e. not against a sosreport) it will print a message suggesting they install the deb or pypi packages that have full support for host and sosreport.

dosaboy:
to kick this off i tried simply changing classic -=> strict and it built fine but when i ran it got the following which I now remember was one of the problems we had before we went classic:

```
Traceback (most recent call last):
  File "/snap/hotsos/x1/bin/hotsos", line 5, in <module>
    from hotsos.cli import main
  File "/snap/hotsos/x1/lib/python3.8/site-packages/hotsos/cli.py", line 15, in <module>
    from hotsos.core.host_helpers import CLIHelper
  File "/snap/hotsos/x1/lib/python3.8/site-packages/hotsos/core/host_helpers/__init__.py", line 1, in <module>
    from .cli import CLIHelper  # noqa: F403,F401
  File "/snap/hotsos/x1/lib/python3.8/site-packages/hotsos/core/host_helpers/cli.py", line 12, in <module>
    from hotsos.core.host_helpers.common import HostHelpersBase
  File "/snap/hotsos/x1/lib/python3.8/site-packages/hotsos/core/host_helpers/common.py", line 4, in <module>
    from searchkit.utils import MPCache
  File "/snap/hotsos/x1/lib/python3.8/site-packages/searchkit/__init__.py", line 1, in <module>
    from .search import (   # noqa: F403,F401
  File "/snap/hotsos/x1/lib/python3.8/site-packages/searchkit/search.py", line 22, in <module>
    RS_LOCK = multiprocessing.Lock()
  File "/usr/lib/python3.8/multiprocessing/context.py", line 68, in Lock
    return Lock(ctx=self.get_context())
  File "/usr/lib/python3.8/multiprocessing/synchronize.py", line 162, in __init__
    SemLock.__init__(self, SEMAPHORE, 1, 1, ctx=ctx)
  File "/usr/lib/python3.8/multiprocessing/synchronize.py", line 57, in __init__
[...]

PR #648, opened 2023-06-16
Make snap confined and change base to core22

tpsilva:
Fixes: #642


tpsilva:
@dosaboy not really sure why those systemd files needed to be excluded, but core22 was not happy about it[1], so I left it commented out.

Tested on a sosreport and it's working fine. If you're ok with it, I'll add the check on main to only allow sosreport exec.


[1] https://snapcraft.io/docs/snapcraft-filesets

nicolasbock (approved):
LGTM

The snapcraft.yaml can be simplified a bit. I put the simplifications into a separate PR, canonical/hotsos#651, but I can also add those changes here as code suggestions.

dosaboy:
lgtm thanks tiago!

PR #659, opened 2023-06-23
Fix confined snap message

dosaboy:
Now that docs have been updated, msg need to point to new location.

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -314,12 +314,14 @@ def main():
         data_root = fix_data_root(data_root)
 
         if is_snap() and data_root == '/':
-            print("WARNING: hotsos is installed as a snap which only "
+            print("ERROR: hotsos is installed as a snap which only "
                   "supports running against a sosreport due to access "
-                  "restrictions. If you want to analyse a host you need to "
+                  "restrictions.\n\n"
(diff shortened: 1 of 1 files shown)


hotsos/cli.py
```diff
@@ -314,11 +314,12 @@ def cli(data_root, version, defs_path, templates_path, all_logs, quiet,
...
         if is_snap() and data_root == '/':
-            print("WARNING: hotsos is installed as a snap which only "
+            print("ERROR: hotsos is installed as a snap which only "
                   "supports running against a sosreport due to access "
-                  "restrictions. If you want to analyse a host you need to "
+                  "restrictions.\n\n"
+                  "If you want to analyse a host you need to "
                   "use an alternative installation method e.g. debian "
-                  "package - see https://github.com/canonical/hotsos#install "
+                  "package - see docs at https://github.com/canonical/hotsos "
```
nicolasbock, line 322: Should we rather point to the readthedocs docs?

dosaboy: i mean we could i just thought is more future proof in case the url ever changed

dosaboy: you are right, ill change it

nicolasbock (approved):

PR #662, opened 2023-06-24
Run functest.sh under bash

brianphaley:
Trying to run 'tox -e functional' locally yields:

functional: failed with /home/bhaley/hotsos/tools/test/functest.sh is not allowed, use allowlist_externals to allow it
  functional: FAIL code 1 (10.80 seconds)
  evaluation failed :( (10.86 seconds)

With the newer version of tox we are using, we need to add it to allowlist_externals.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -68,7 +68,8 @@ allowlist_externals = bashate
 commands = bashate --verbose {toxinidir}/build.sh {toxinidir}/tools/test/functest.sh
 
 [testenv:functional]
-commands = {toxinidir}/tools/test/functest.sh
+allowlist_externals = bash
+commands = bash {toxinidir}/tools/test/functest.sh
 
 [testenv:docs]
 commands = sphinx-build -j auto -d {toxinidir}/doc/build/doctrees -b html {toxinidir}/doc/source doc/build/html


tox.ini
```diff
@@ -68,7 +68,8 @@ allowlist_externals = bashate
 commands = bashate --verbose {toxinidir}/build.sh {toxinidir}/tools/test/functest.sh
 
 [testenv:functional]
-commands = {toxinidir}/tools/test/functest.sh
+allowlist_externals = bash
```
nicolasbock, line 71: With `tox-4.4.6-1` locally I need to set

```suggestion
allowlist_externals = {toxinidir}/tools/test/functest.sh
```

Allowing `bash` does not work.

brianphaley: Hmm, I'm running tox 4.6.3 and it works, and most of the charm-* repos have bash in allowlist_externals. Did you have the updated 'commands' line?

brianphaley: I think either will work, just copied what I saw in neutron and nova.

nicolasbock: Ah ok. I suppose the advantage of your change is that it is now explicit that the script is really bash.

tox.ini
```diff
@@ -68,7 +68,8 @@ allowlist_externals = bashate
...
+commands = bash {toxinidir}/tools/test/functest.sh
 
 [testenv:docs]
 commands = sphinx-build -j auto -d {toxinidir}/doc/build/doctrees -b html {toxinidir}/doc/source doc/build/html
```
nicolasbock, line 75: Interesting that this env does not require an allowlist.

nicolasbock (approved):

tox.ini
brianphaley: I think there are some commands that are implicitly allowed, like stestr and this one, and you only find out when you run it using tox4 :)

nicolasbock: Haha, always great to find those obscure exceptions 😉

PR #669, opened 2023-06-30
Add types to config opt defs

dosaboy:
When config options are registered they must now provide a type and the value they are set to will be cast to this type.

### DIFF
--- a/hotsos/core/config.py
+++ b/hotsos/core/config.py
@@ -9,16 +9,17 @@ class ConfigException(Exception):
 
 class ConfigOpt(object):
 
-    def __init__(self, name, description, default_value):
+    def __init__(self, name, description, default_value, type):
         self.name = name
         self.description = description
         self.default_value = default_value
+        self.type = type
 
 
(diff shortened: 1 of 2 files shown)


hotsos/core/config.py
```diff
@@ -159,9 +160,30 @@ def name(self):
...
     def __init__(self, *optgroups):
+        self.optsgroups = []
         self.data = {}
-        for ogroup in optgroups:
-            self.data.update(ogroup())
+        for optgroup in optgroups:
+            optgroup = optgroup()
+            self.optsgroups.append(optgroup)
+            a = [name.lower() for name in self.data.keys()]
+            b = [name.lower() for name in optgroup.keys()]
```
brianphaley, line 169: Could add a test for groups/names being duplicated based on case, but wouldn't hold it over that.

brianphaley (approved):

hotsos/core/config.py
dosaboy: done.

brianphaley (approved):

Issue #673, opened 2023-07-10
ovn cert checks should abort if start time not available


dosaboy:
Currently if the service e.g. ovn-controller start time is not available in the systemd journal we return 0 for the start time which results in false positives in checks like https://github.com/canonical/hotsos/blob/4b8e036490572a63d692f67fd100f1518734389a/hotsos/defs/scenarios/openvswitch/ovn/ovn_chassis_certs_logs.yaml#L8

We need to add a guard against and also consider looking for start time elsewhere e.g. in service status.

PR #674, opened 2023-07-10
Ensure start time non-zero when checking certs

dosaboy:
When we check that certs are modified before last service start we must ensure that the start time is non-zero for cases where it is not available.

Adds support for getting service start time from status if not found in journal.

Resolves: #673


mustafakemalgilor (approved):
LGTM :shipit:

Issue #675, opened 2023-07-11
ovn-northd memory usage isn't accurately reported by hotsos


nishant-dash:
When I run hotsos with the openvswitch plugin on an ovn-central node on focal, I see
```console
    OpenvSwitchWarnings:
      - The ovn-northd service is consuming more than 5G memory (current=903585792).
        This is not normal and could indicate a memory leak. Please check. (origin=openvswitch.auto_scenario_check
```

However,  when checking
```console
root     1940327  3.6  0.0 5653052 850368 ?      Ssl  Apr13 4679:30 ovn-northd -vconsole:emer -vsyslog:err -vfile:info --ovnnb-db=ssl:10.130.11.131:6641,ssl:10.130.12.66:6641,ssl:10.130.13.120:6641 --ovnsb-db=ssl:10.130.11.131:16642,ssl:10.130.12.66:16642,ssl:10.130.13.120:16642 -c /etc/ovn/cert_host -C /etc/ovn/ovn-central.crt -p /etc/ovn/key_host --no-chdir --log-file=/var/log/ovn/ovn-northd.log --pidfile=/var/run/ovn/ovn-northd.pid --detac
```

that is only a RSS of 850 MB. 

Also It would be nice if hotsos directly reported the units beside `current=XXX <units>`

Env:
- Ubuntu 20.04.4 LTS
- in a lxd
- ovn-central unit, 22.03 on openstack focal ussuri
- hotsos(deb): 1.15.10.g4b8e03649057+20.04.20230707194707

mustafakemalgilor:
It seems that the value is currently retrieved from the respective systemd services' memory cgroup:

https://github.com/canonical/hotsos/blob/4b8e036490572a63d692f67fd100f1518734389a/hotsos/core/host_helpers/systemd.py#L56-L61

The memory consumption is being pulled from the `memory.usage_in_bytes` (or `memory.current` if >=jammy). The [kernel cgroup documentation](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt) mentions that `memory.usage_in_bytes` does not reflect the exact memory usage and it's better to rely on (RSS + CACHE + SWAP) instead:

```text
5.5 usage_in_bytes

For efficiency, as other kernel components, memory cgroup uses some optimization
to avoid unnecessary cacheline false sharing. usage_in_bytes is affected by the
method and doesn't show 'exact' value of memory (and swap) usage, it's a fuzz
[...]

PR #676, opened 2023-07-11
Fix service mem usage checks

dosaboy:
The systemd helper now checks memory.stat instead of memory.usage_in_bytes since the latter is just an
approximation.

Resolves: #675

### DIFF
--- a/hotsos/core/host_helpers/systemd.py
+++ b/hotsos/core/host_helpers/systemd.py
@@ -86,25 +86,47 @@ class SystemdService(object):
         return t.timestamp()
 
     @property
-    def memory_current(self):
-        """ Returns service current memory usage in bytes. """
-        path = os.path.join(HotSOSConfig.data_root,
-                            'sys/fs/cgroup/memory/system.slice',
-                            "{}.service".format(self.name),
-                            'memory.usage_in_bytes')
-        if not os.path.exists(path):
-            # path changed between Ubuntu Focal and Jammy releases.
-            path = os.path.join(HotSOSConfig.data_root,
-                                'sys/fs/cgroup/system.slice',
-                                "{}.service".format(self.name),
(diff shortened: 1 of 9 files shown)


hotsos/core/host_helpers/systemd.py
```diff
@@ -86,25 +86,40 @@ def start_time_secs(self):
...
+        fs = FileSearcher()
+        fs.add(SearchDef(r'(cache|rss|swap) (\d+)'), path=path)
+        for result in fs.run().get(path, {}):
+            total_usage[result.get(1)] = int(result.get(2))
+
+        if len(total_usage) == 0:
+            log.warning("failed to identify mem usage info for %s in %s",
+                        "{}.service".format(self.name), path)
+
+        if len(total_usage) < 3:
```
dosaboy, line 114: swap counter will only exist if swap is enabled so we should not print this warning

hotsos/core/host_helpers/systemd.py
```diff
@@ -86,25 +86,36 @@ def start_time_secs(self):
...
                                 "{}.service".format(self.name),
-                                'memory.current')
+                                'memory.stat')
 
-        if not os.path.exists(path):
-            log.warning("service memory info not found at %s", path)
-            return 0
+        total_usage = {}
+        fs = FileSearcher()
+        fs.add(SearchDef(r'(cache|rss|swap) (\d+)'), path=path)
```
mustafakemalgilor, line 106: memory.stat does not contain cache/rss/swap values in [cgroup v2](https://www.kernel.org/doc/Documentation/cgroup-v2.txt) (jammy and above).

dosaboy: going to stick with memory.current for cgroup-v2 unless we think that is also not reliable

Issue #678, opened 2023-07-12
protect against systemctl_status_all non-unicode chars


dosaboy:
When a service start time cannot be obtained from the systemd journal we turn to systemctl status which apparently can have non-unicode chars so we need to account for that:

Traceback (most recent call last):
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/ycheck/engine/properties/common.py", line 482, in get_property
    _obj = getattr(_obj, _prop)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/utils.py", line 22, in _cached_property
    val = f(self)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/systemd.py", line 82, in start_time_secs
    t = self.start_time
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/utils.py", line 22, in _cached_property
    val = f(self)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/systemd.py", line 65, in start_time
    fs.add(seqdef, path=mktemp_dump(''.join(cli.systemctl_status_all())))
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 506, in __call__
    return fsource(*args, **kwargs)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 33, in catch_exceptions_inner2
    return f(*args, **kwargs)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 106, in reset_command_inner
    out = f(self, *args, **kwargs)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 90, in run_post_exec_hooks_inner
    out = f(self, *args, **kwargs)
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 80, in run_pre_exec_hooks_inner
    return f(self, *args, **kwargs)
[...]

PR #679, opened 2023-07-12
Fix CLIHelper().systemctl_status_all

dosaboy:
The output of this command can sometimes contain non-unicode
chars or special encodings e.g. in sosreports so we need to
handle UnicodeDecodeErrors correctly. We now process each line
individually and store an empty line for those that could not
be decoded. We also use subprocess.run as opposed to
subprocess.check_output since some commands like this return
non-zero exit codes by default which does not work with
check_output.

Resolves: #678

### DIFF
--- a/hotsos/core/host_helpers/cli.py
+++ b/hotsos/core/host_helpers/cli.py
@@ -175,35 +175,47 @@ class BinCmd(CmdBase):
             cmd = cmd.format(**kwargs)
 
         _cmd = cmd.split() + self.original_cmd_extras
-        output = subprocess.check_output(_cmd,
-                                         timeout=HotSOSConfig.command_timeout,
-                                         stderr=subprocess.STDOUT)
+        out = subprocess.run(_cmd, timeout=HotSOSConfig.command_timeout,
+                             capture_output=True, check=False)
+        output = out.stdout
+        if out.stderr:
+            output += out.stderr
+
+        if out.returncode != 0:
+            log.info("command '%s' exited with non-zero code '%s'", cmd,
+                     out.returncode)
+
+        try:
+            output = output.decode('UTF-8')
+        except UnicodeDecodeError:
+            log.exception("failed to decode command output for '%s'", cmd)
+            output = ''
 
         if self.json_decode:
-            return json.loads(output.decode('UTF-8'))
+            return json.loads(output)
 
         if self.singleline:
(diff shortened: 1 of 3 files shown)


hotsos/core/host_helpers/cli.py
```diff
@@ -228,7 +234,15 @@ def __call__(self, *args, **kwargs):
         elif self.json_decode:
             output = json.load(open(self.path))
         else:
-            output = open(self.path, 'r').readlines()
+            output = []
+            ln = 0
+            for line in open(self.path, 'rb'):
```
brianphaley, line 239: I know pylint will usually complain about not using a "with open()" since that causes the file to get closed on leaving the branch. Could probably fix this with something like:

with open(self.path, 'rb') as f:
    for line in f.readlines():
        ...

dosaboy: +1 will fix

dosaboy:
To test the bincmd manually you do:

```
from hotsos.core.host_helpers.cli import CLIHelper
HotSOSConfig.data_root='/'
HotSOSConfig.global_tmp_dir='/tmp/foo'
HotSOSConfig.plugin_tmp_dir='/tmp/foo'
out=CLIHelper().systemctl_status_all()
open('/tmp/foo/out.txt', 'w').write(''.join(out))
```

pponnuvel (approved):
LGTM

hotsos/core/host_helpers/cli.py
```diff
@@ -219,16 +231,36 @@ def __call__(self, *args, **kwargs):
...
-            output = self.safe_readlines(self.path)
-        elif self.json_decode:
+        if self.json_decode:
             output = json.load(open(self.path))
         else:
-            output = open(self.path, 'r').readlines()
+            output = []
+            ln = 0
+            with open(self.path, 'rb') as fd:
+                for line in fd:
```
brianphaley, line 242: Shouldn't this be fd.readlines() ?

dosaboy: no because we want to do one line at a time, if you do readlines() its all or nothing.

dosaboy: i.e. if you do fd.readlines() and any line is not decodable, all will fail

brianphaley (approved):

PR #681, opened 2023-07-17
CLI context manager to support returning path

dosaboy:
Adds a new CLIHelperFile class that will return a path to a file containing the output of the command. This is so that we can stop using mktemp_dump() to dump contents of command output to a file which is redundant if the output came from a file in the first place e.g. with a sosreport.


mustafakemalgilor (approved):
LGTM :shipit:

nicolasbock (approved):
LGTM

PR #684, opened 2023-07-21
Add bash completion for snap

nicolasbock:
This change adds bash completions for the snap version of hotsos.


dosaboy:
nice!

PR #690, opened 2023-07-24
host_helpers/systemd: fix start_time does not respect timezone when r…

### LINKED ISSUE #689: systemd helper start_time does not honour tz from service status
When https://github.com/canonical/hotsos/blob/main/hotsos/core/host_helpers/systemd.py:SystemdService gets start_time from systemctl status it does not honour that tz - it should be converting to UTC since the rest of hotsos works on that basis.

mustafakemalgilor:
…eading from systemd status

the start_time method first tries to pull the start time info from system journal which already accounts for the timestamp's timezone. if it fails, it falls back to parsing `systemctl status --all` output and it ignores the specified timezone in the timestamp, which causes it to treat time in the local timezone of the running system. this patch fixes that.

Fixes #689


hotsos/core/host_helpers/systemd.py
```diff
@@ -58,11 +58,12 @@ def start_time(self):
         fs = FileSearcher()
         # The following expressions need to take account of control characters
         # that might exist in the output e.g. line can start with '*' or U+25CF
+        # Active: active (running) since Wed 2022-02-09 22:38:17 UTC; 17h ago
         seqdef = SequenceSearchDef(
                     start=SearchDef(r'\S+ ({}.service) -'.format(self.name)),
                     body=SearchDef(r"\s+Active: active \(?\S*\)?\s*since "
                                    r"\S{3} (\d{4}-\d{2}-\d{2} "
-                                   r"\d{2}:\d{2}:\d{2})"),
+                                   r"\d{2}:\d{2}:\d{2} .*);"),
```
dosaboy, line 66: instead of .* can we use \w+ instead so that (a) it stops at the first non-unicode char and (b) we guarantee it exists (since the followup code will fail if it doesnt).

mustafakemalgilor: > instead of .* can we use \w+ instead so that (a) it stops at the first non-unicode char and (b) we guarantee it exists (since the followup code will fail if it doesnt).

Replaced `.*` with `[\w\+:]+` since timezone strings can contain `+` and `:` on top of word characters.

mustafakemalgilor:
> just leaving a note say that (as discussed) this needs a unit test

fixed some tz quirks, now ready to review.

hotsos/core/host_helpers/systemd.py
```diff
@@ -26,6 +29,25 @@ def __init__(self, name, state, has_instances=False):
         self.state = state
         self.has_instances = has_instances
 
+    @cached_property
+    def __tzinfos(self):
```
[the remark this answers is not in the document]
dosaboy: also could do with a docstring

Issue #691, opened 2023-07-24
run tests utc and non-utc timezone


dosaboy:
timezones are important in hotsos because it needs to run the same whatever the timezone e.g. when applying search constraints. we should therefore run the tests using a utc and non-utc timezone to ensure we get the same results.

PR #692, opened 2023-07-24
github-actions&tox: add non-utc timezone test run

mustafakemalgilor:
the tests were only being run in the UTC timezone which hides most of the timezone problems the code potentially has. this patch extends the existing `py3` tox environment to pass a non-UTC timezone to the test environment.

Fixes: #691

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ setenv = VIRTUAL_ENV={envdir}
          HOTSOS_ROOT={toxinidir}/hotsos
          TESTS_DIR={[testenv]unit_tests}
          PYFILES={toxinidir}/setup.py {toxinidir}/hotsos/ {[testenv]unit_tests}
+         non-utc-tz: TZ=EST+5
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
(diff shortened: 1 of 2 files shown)


mustafakemalgilor:
in order to pipeline to be green, this MP depends on https://github.com/canonical/hotsos/pull/690 to be merged first.

dosaboy:
@mustafakemalgilor 690 is merged so you can rebase

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = True
-envlist = py3,pep8,pylint,bashate,functional
+envlist = py3-{non-utc-tz},pep8,pylint,bashate,functional
```
dosaboy, line 3: is the curly brackets needed here?

dosaboy: so as it is now, if i run ```tox``` it does:

  py3-non-utc-tz: commands succeeded
  pep8: commands succeeded
  pylint: commands succeeded
  bashate: commands succeeded
  functional: commands succeeded
  congratulations :)

i.e. it does not run -epy3. Personally i think it would be better to run -epy3-non-utc-tz in the github workflow (along with -epy3) and leave the default envlist as it was i.e. ```envlist = py3,pep8,pylint,bashate,functional```

mustafakemalgilor: I see, I'll fix it in a few.

mustafakemalgilor: ... done

Issue #694, opened 2023-07-24
check if nodes impacted by fix for cve-2023-2088


dosaboy:
The fix for https://discourse.ubuntu.com/t/cve-2023-2088-for-charmed-openstack/37051 is going out tomorrow and users that have not upgraded their charms to have support for service tokens may be impacted if they upgrade their packages first. We can add a check to detect those impacted and raise an issue accordingly.

PR #695, opened 2023-07-24
Add checks for CVE-2023-2088 and fixes

dosaboy:
resolves: #694

### DIFF
--- a/hotsos/core/ycheck/engine/properties/requires/types/apt.py
+++ b/hotsos/core/ycheck/engine/properties/requires/types/apt.py
@@ -30,7 +30,7 @@ class APTCheckItems(PackageCheckItemsBase):
 
     def package_version_within_ranges(self, pkg, versions):
         result = False
-        versions = sorted(versions, key=lambda i: i['min'], reverse=True)
+        versions = sorted(versions, key=lambda i: str(i['min']), reverse=True)
         pkg_version = self.packaging_helper.get_version(pkg)
         for item in versions:
             v_min = str(item['min'])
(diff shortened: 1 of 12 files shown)


nicolasbock (approved):

sombrafam (changes requested):
So, as we talked in our meeting, we should have a way to check if the customer has service tokens enable so we avoid detecting false positives. 

Also, we should check for all the packages/services affected:

nova >= 3:27.0.0-0ubuntu1.3
cinder >= 2:22.0.0-0ubuntu1.3
python-os-brick >= 6.2.0-0ubuntu2.3
ironic >= 1:21.4.0-0ubuntu1.1
python-glance-store >= 4.3.0-0ubuntu1.3

dosaboy:
> So, as we talked in our meeting, we should have a way to check if the customer has service tokens enable so we avoid detecting false positives.
> 
> Also, we should check for all the packages/services affected:
> 
> nova >= 3:27.0.0-0ubuntu1.3 cinder >= 2:22.0.0-0ubuntu1.3 python-os-brick >= 6.2.0-0ubuntu2.3 ironic >= 1:21.4.0-0ubuntu1.1 python-glance-store >= 4.3.0-0ubuntu1.3

+1 i'll add that in, should not be too hard

sombrafam:
II dont know if we support configurations where glance and ironic could also be affected be we would better to add checks for them as well.

sombrafam (changes requested):
II dont know if we support configurations where glance and ironic could also be affected be we would better to add checks for them as well.

dosaboy:
> II dont know if we support configurations where glance and ironic could also be affected be we would better to add checks for them as well.

ive done cinder-api, nova-api-os-compute and nova-compute for now (since they are the api involved in an attach/detach) and we can add more later if needs be

PR #698, opened 2023-08-02
Cleanup some cached properties

dosaboy:
Cleans up properties that are using cached_property but also using a variable to cache the value which is now redundant.

### DIFF
--- a/hotsos/core/plugins/kubernetes.py
+++ b/hotsos/core/plugins/kubernetes.py
@@ -46,16 +46,11 @@ K8S_PACKAGE_DEPS_SNAP = [r'core[0-9]*']
 
 class KubernetesBase(object):
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


nicolasbock (approved):

hotsos/core/plugins/kubernetes.py
```diff
@@ -70,24 +65,17 @@ def bind_interfaces(self):
 
     @cached_property
     def pods(self):
-        if self._pods:
-            return self._pods
-
         _pods = []
```
brianphaley, line 68: nit: could probably just be 'pods' and below 'containers', guessing used _ to match private name?

brianphaley (approved):

hotsos/core/plugins/kubernetes.py
dosaboy: wont that trigger a pylint error as it overrides a global name?

dosaboy: hmm no didnt

PR #699, opened 2023-08-02
Remove `stestr` dependency

nicolasbock:
We are not using `stestr` in parallel and can simplify the test
requirements by using `unittest` instead.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,7 @@ setenv = VIRTUAL_ENV={envdir}
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
-commands = stestr run --serial --test-path {[testenv]unit_tests} {posargs}
+commands = python -m unittest --verbose {posargs}
 
 [testenv:pep8]
 allowlist_externals = flake8
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,7 +1,6 @@
 bashate
 flake8==6.0.0
 flake8-import-order==0.18.2
-stestr
 pylint==2.17.4
 rpdb
 testtools


tox.ini
```diff
@@ -20,7 +20,7 @@ setenv = VIRTUAL_ENV={envdir}
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
-commands = stestr run --serial --test-path {[testenv]unit_tests} {posargs}
```
brianphaley, line 23: Guess you could have dropped the --serial too but same result

tox.ini
```diff
@@ -20,7 +20,7 @@ setenv = VIRTUAL_ENV={envdir}
...
+commands = python3 -m unittest --verbose {posargs}
```
brianphaley, line 23: Should this just be 'python' ? It will default to 3 anyways

tox.ini
nicolasbock: I figured that we could drop that dependency. I don't like having too many dependencies, the more you have the more things can break 😉

tox.ini
dosaboy: I agree that removing another dep is a good thing but i also checked and it seems that running in parrallel is also safe now. im fine either way i guess.

nicolasbock: On my laptop with 4 cores and hyperthreading enabled I get
```console
  py3: OK (55.01=setup[19.25]+cmd[35.76] seconds)
  congratulations :) (55.07 seconds)
```
for `stestr` parallel execution and
```console
  py3: OK (95.11=setup[19.67]+cmd[75.44] seconds)
  congratulations :) (95.17 seconds)
```
for `stestr --serial` execution. For comparison, `unittest` is
```console
  py3: OK (96.50=setup[15.58]+cmd[80.92] seconds)
  congratulations :) (96.57 seconds)
```

nicolasbock: Basically parallel testing is about twice as fast.

mustafakemalgilor: AFAIK unittest does not support parallel test execution by default. If we want parallel testing we should either keep stestr or switch to something else, like pytest. IMO, stestr is a niche thing and not widely adopted so I would be in favor of switching to pytest.

mustafakemalgilor: I assumed pytest would support parallel execution, but to my surprise, it doesn't. So sticking with stestr doesn't look too bad if we want to run the tests in parallel.

nicolasbock: To be honest, I don't think that we really need to care about test performance. Even when run in serial the tests complete within minutes which I think is completely reasonable. We should focus on the dependency question only.

mustafakemalgilor: > To be honest, I don't think that we really need to care about test performance. Even when run in serial the tests complete within minutes which I think is completely reasonable. We should focus on the dependency question only.

As I mentioned in the other MP, speed is not the only benefit here, and we should think of concurrency as well. hotsos is a multi-process application, so running the test concurrently may reveal bugs that are otherwise not possible running in serial.  I don't expect unit tests to reveal any kind of concurrency issues but functional or integration tests might.

nicolasbock: There are two kinds of parallel here: (1) running multiple tests in parallel and (2) running one test in parallel. `stestr` does the former and would not be helpful in uncovering race conditions within hotsos, which would require the latter.

`stestr` might be helpful in uncovering unintended dependencies between unit tests, but I don't really see a benefit here. The tests are not supposed to depend on each other.

mustafakemalgilor: > There are two kinds of parallel here: (1) running multiple tests in parallel and (2) running one test in parallel. `stestr` does the former and would not be helpful in uncovering race conditions within hotsos, which would require the latter

OK, let me clarify what I mean here. 

I know that `stestr` runs multiple tests in parallel, and that's the point. Running test cases in parallel (1) is kind of similar to what hotsos does in runtime -- *multiple execution paths calling different functions concurrently*. So, it *may* reveal a concurrency issue, especially in the long run. The (2) is also desirable, but it's not the *concurrency* I was talking about before.

Assume that there are two different functions A and B, accessing the same system resource C, but the resource C is not thread/process safe. In such a scenario, we *may* discover the concurrency problem if we run the tests in parallel, but we will not never, ever discover that if we run the tests serially.

So, as a best practice, I always tend to:

- Shuffle the test cases
- Always run all tests concurrently (... given that the code must be concurrent-safe)
- Repeat them N times (N depends on how costly the test suite is)

... so the CI pipeline has a better chance of finding a concurrency issue while executing the tests.

> The tests are not supposed to depend on each other

*yes*,  randomizing the tests' execution order is a way to confirm that is indeed true. Isn't the whole point of testing things is *confirming that things are doing what they're supposed to do*?. Let me explain with one example.

Given that we have 3 test cases, X, Y, and Z, the results must be EXACTLY the same independent of the run order (e.g. [Z, X, Y], [X, Z, Y])... Let's say we have a hidden dependency between X and Y, (e.g. X sets a global variable that Y also depends on and that causes Y to be green), changing the run order from XY to YX *may* reveal that hidden dependency. 

I hope this is clear enough :)

nicolasbock: > Running test cases in parallel (1) is kind of similar to what hotsos does in runtime -- multiple execution paths calling different functions concurrently.

These two cases are very different. In case (1) you are simply running the same thing multiple times concurrently just as if multiple users ran the same command at the same time for example. There is no shared memory between these processes and their interaction is very limited. In case 2 you are using some form of parallel processing and you likely use some form of interprocess communication. Interactions between parallel execution units are a desired design goal here.

> Assume that there are two different functions A and B, accessing the same system resource C, but the resource C is not thread/process safe.

Within the test suite, the only system resource `hotsos` uses is a read on files which is hopefully thread/process safe :wink:

mustafakemalgilor: > These two cases are very different. In case (1) you are simply running the same thing multiple times concurrently just as if multiple users ran the same command at the same time for example.

That's not the scenario when you run tests with "stestr --parallel". The test runner aggregates the test cases and divides them into chunks, each to be executed by a different process. So, if you randomize the test cases, the test runner would be running different functions concurrently each time -- which IMO is a nice thing to have. 

> Within the test suite, the only system resource `hotsos` uses is a read on files which is hopefully thread/process safe 😉

For now. The test cases should ideally be abstract from the implementation details and never should assume anything. Hotsos may consume other data sources in the future, which may or may not be read-only. Tests should not be designed to test the implementation. Instead, they should be in place for validating the contract that the implementation must be fulfilling. 

Anyway, I'm just expressing my *opinion* here to provoke some brainstorming -- it's your call after all :)

nicolasbock: > That's not the scenario when you run tests with "stestr --parallel".

Let me clarify what I was trying to say:

Each test runner runs a separate interpreter and a set of tests inside that interpreter. The tests within each runner are executed one after the other but potentially in a parallel context, e.g. using threads.

In the context of the tests, `hotsos` is only reading from the filesystem and is not using any other system information. Anything not contained in test artifacts would be useless in unit tests because that information would not be reproducible as it depends on the CI system.

I fail to see how tests in separate test runners could possibly interact which leads me to conclude that the two things we gain from using `stestr` is potentially faster execution of the test suite and another dependency.

In the end, either way of running the test suite is a minor detail because both are really testing the same thing.

mustafakemalgilor: What I'm trying to tell you is the claims like 'hotsos is strictly read-only', or 'it only uses Filesystem read-only so it's inherently thread/process safe' are nothing but assumptions, which conflicts with the black-box testing. Implementation can and will do whatever it desires, and if the tests do not validate these claims, the implementation might break them without a notice. I'm not saying that running these tests with 'stestr --parallel' would validate these claims, but it might be a start. Projects are moving targets, implementations can and will change, so can these assumptions. 

Have a good one 👋

PR #700, opened 2023-08-02
Swith to parallel testing with `stestr`

nicolasbock:


nicolasbock:
For comparison I added this alternative to #699

brianphaley (approved):
I guess since it's faster than unittest I'd go with this.

nicolasbock:
From GitHub CI:

`stestr --parallel` is 65 s

<https://github.com/canonical/hotsos/actions/runs/5742796873/job/15565675485#step:5:845>

vs. `unittest` 152 s

<https://github.com/canonical/hotsos/actions/runs/5742796873/job/15565675485#step:5:845>

mustafakemalgilor:
Speed is not the only benefit of test parallelization. It also allows revealing weird race conditions and concurrency issues. I'd also recommend shuffling the test execution order as well (via the `--random` option). This will ensure that the test cases have no implicit dependencies between them.

mustafakemalgilor:
> This is only the case if you have dependencies between tests which we don't as far as I am aware of.

Test randomization is a way to *ensure* that the test cases are independent of each other. Changing the execution order of tests should not affect the test results in any way. After all, you shouldn't have any dependencies between any unit tests. If you do, then it's not a unit test -- unit tests, by definition, are individual.

nicolasbock:
Sure, in theory that's correct but I don't see any reason to assume that there are any interactions between the tests here. No test consumes anything another produces. `hotsos` is strictly read-only.

mustafakemalgilor:
> Sure, in theory that's correct but I don't see any reason to assume that there are any interactions between the tests here

Yes, my thoughts as well. It's most probably the case, but it's more about validating a *contract* here. I'm not implying there are test cases that depend on each other, but I want to make sure that it stays that way for the future, and if it doesn't, we find out easily. Hopefully, we'll have more contributors as the time passes and I want to make sure that we can discover as many mistakes as possible without having to look at the code that people are sending to review.

> `hotsos` is strictly read-only.

This is also a contract that needs to be validated -- if not being done already ;)

dosaboy:
I'm going to ahead and close this pull request since I do not believe we have a strong need (nor desire?) at the moment to parallelise tests. Lets re-evaluate.

PR #701, opened 2023-08-03
plugin/system: added `uptime` to summary

### LINKED ISSUE #680: Add uptime to system plugin summary

mustafakemalgilor:
Closes #680

### DIFF
--- a/hotsos/core/plugins/system/system.py
+++ b/hotsos/core/plugins/system/system.py
@@ -3,7 +3,10 @@ import re
 from functools import cached_property
 
 from hotsos.core.config import HotSOSConfig
-from hotsos.core.host_helpers import CLIHelper, CLIHelperFile, SYSCtlFactory
+from hotsos.core.host_helpers import (
+    CLIHelper, CLIHelperFile,
+    SYSCtlFactory, UptimeHelper
+)
 from hotsos.core.log import log
 from hotsos.core.search import (
     FileSearcher, SearchDef,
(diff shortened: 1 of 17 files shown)


hotsos/core/plugins/system/system.py
```diff
@@ -81,6 +81,10 @@ def date(self):
     def hostname(self):
         return CLIHelper().hostname()
 
+    @cached_property
```
dosaboy, line 84: I implemented https://github.com/canonical/hotsos/blob/main/hotsos/core/host_helpers/uptime.py some time back partly for this purpose but never got around to adding it to the summary. Personally I think it would be preferable if the uptime was just the elapsed time and not the loadavg etc since that is already in the system summary under ```load:```

mustafakemalgilor: ack, working on it

mustafakemalgilor: .... done.

dosaboy: you've made me realise how unclear my UptimeHelper is so ive gone ahead and proposed a cleanup in https://github.com/canonical/hotsos/pull/703 which is similar to what you have done here but provides a way to repr(helper) to get the string format. Let me know what you think.

mustafakemalgilor: #703 is way clearer than what I did here, I approved it and I'll revise this MP after it's merged.

mustafakemalgilor: @dosaboy, rebased and ready for review

PR #703, opened 2023-08-03
Make UptimeHelper clearer

dosaboy:
Makes it clearer that the timing properties are reflections of total time by prefixin them with in_* e.g. in_minutes. Also implements a __repr__ to return a string of format <int>d:<int>h:<int>m:<int>s to represent total uptime.

### DIFF
--- a/hotsos/core/host_helpers/uptime.py
+++ b/hotsos/core/host_helpers/uptime.py
@@ -7,8 +7,13 @@ from hotsos.core.log import log
 
 class UptimeHelper(object):
     def __init__(self):
-        # unfortunately sosreports dont have proc/uptime otherwise we would use
-        # that.
+        """
+        uptime is either expressed in different combinations of days, hours
+        and minutes so we determine which format is used to then extract
+        the time to present it in a common way.
+        """
+        # unfortunately sosreports don't have proc/uptime otherwise we would
+        # use that.
         self.uptime = CLIHelper().uptime() or ""
         # this needs to take into account the different formats supported by
         # https://gitlab.com/procps-ng/procps/-/blob/newlib/library/uptime.c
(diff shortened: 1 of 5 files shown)


mustafakemalgilor:
regarding the failing jobs, there are references to `.hours` in scenarios as well so they should be updated:

- hotsos/defs/tests/scenarios/storage/ceph/ceph-mon/mon_elections_flapping.yaml
- hotsos/defs/tests/scenarios/storage/ceph/ceph-mon/osd_slow_heartbeats.yaml

hotsos/core/host_helpers/uptime.py
```diff
@@ -58,14 +64,22 @@ def minutes(self):
...
-            return int(self.minutes / 60)
+    def in_hours(self):
+        """ Total uptime in hours. """
+        if self.in_minutes:
+            return int(self.in_minutes / 60)
+
+    def __repr__(self):
+        days = int(self.in_hours / 24)
+        hours = self.in_hours - days * 24
+        minutes = self.in_minutes - (self.in_hours * 60)
```
brianphaley, line 81: It looks like all in_* methods can return None if their respective in_* values are zero. Should they have a 'return 0' as last line to force a valid value? Else some of these calculations will throw a TypeError I think.

hotsos/core/host_helpers/uptime.py
```diff
@@ -58,14 +64,22 @@ def minutes(self):
...
-    def hours(self):
-        if self.minutes:
-            return int(self.minutes / 60)
+    def in_hours(self):
+        """ Total uptime in hours. """
+        if self.in_minutes:
+            return int(self.in_minutes / 60)
+
+    def __repr__(self):
+        days = int(self.in_hours / 24)
```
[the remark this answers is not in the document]
dosaboy: ive fixed them to return 0 instead of None

hotsos/core/host_helpers/uptime.py
dosaboy: fwict this is ok now

Issue #682, opened 2023-07-17
Would be nice to summarize some statistics


brianphaley:
I was just looking at some hotsos output, and while sometimes it's easy to count lines in a block, in this case there were too many:

ovs-vswitchd:
      bridge-no-such-device:
        '2023-07-17':
          tapc6ecd39c-ca: 1
          tapfcb60ee9-a2: 11
          tap57f591b3-f3: 8
          tapc879300d-c5: 10
          tap53d9a607-68: 1
          tapcd3b6c07-08: 9
          ... hundreds more ...

Perhaps if there was a summary at the top of each one, something like:

ovs-vswitchd:
      bridge-no-such-device:
        '2023-07-17' (965):
          tapc6ecd39c-ca: 1
          tapfcb60ee9-a2: 11
          tap57f591b3-f3: 8
          tapc879300d-c5: 10
          tap53d9a607-68: 1
          tapcd3b6c07-08: 9

I don't think there were 965 just had to pick some number.

There are other areas like 'vm-info' that could have a similar change.

And looking at this list wonder if it's useful to sort by occurrence? Could make a new bug for that too.

PR #706, opened 2023-08-05
Shorten some summary entries

dosaboy:
Some summary entries can have long lists of subentries which can make them hard to read and with no real value. This adds support to events.EventProcessingUtils.categorise_events to pick top N results for events saved by date.

Also summarises neutron l3ha router counts and nova vm counts.

Resolves: #682

### DIFF
--- a/examples/hotsos-example-openstack.summary.yaml
+++ b/examples/hotsos-example-openstack.summary.yaml
@@ -109,8 +109,7 @@ openstack:
     - qemu-kvm 1:4.2-3ubuntu6.19
     - radvd 1:2.17-2
   neutron-l3ha:
-    backup:
-      - 984c22fd-64b3-4fa1-8ddd-87090f401ce5
+    backup: 1
   os-server-external-events:
     network-changed:
       succeeded:
@@ -122,7 +121,9 @@ openstack:
           instance: 359150c9-6f40-416e-b381-185bff09e974
   vm-info:
     running:
-      - d1d75e2f-ada4-49bc-a963-528d89dfda25
+      count: 1
+      uuids:
+        - d1d75e2f-ada4-49bc-a963-528d89dfda25
     cpu-models:
       Skylake-Client-IBRS: 1
     vcpu-info:
--- a/hotsos/core/ycheck/events.py
+++ b/hotsos/core/ycheck/events.py
@@ -75,7 +75,8 @@ class EventProcessingUtils(object):
 
     @classmethod
     def categorise_events(cls, event, results=None, key_by_date=True,
-                          include_time=False, squash_if_none_keys=False):
+                          include_time=False, squash_if_none_keys=False,
+                          max_results_per_date=None):
         """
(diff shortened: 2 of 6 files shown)


dosaboy:
some example output:

    ovs-vswitchd:
      bridge-no-such-device:
        '2023-07-21':
          total: 1143
          top5:
            tap5ec45597-c2: 71
            tapaa55ae19-21: 71
            tapb3d19e75-70: 71
            tap2636be37-0c: 71
            tap6e29df9f-c3: 71

hotsos/core/ycheck/events.py
```diff
@@ -187,7 +191,23 @@ def categorise_events(cls, event, results=None, key_by_date=True,
...
+            for date, entries in results.items():
+                if len(entries) <= max_results_per_date:
+                    shortened[date] = entries
+                    continue
+
+                top_n = dict(sorted(entries.items(),
+                                    key=lambda e: e[1],
+                                    reverse=True)[:max_results_per_date])
+                shortened[date] = {'total': len(entries),
+                                   "top{}".format(max_results_per_date): top_n}
```
brianphaley, line 208: Question on this, and maybe it doesn't matter, but will 'total' always be the first element printed? I thought it was indeterminate so could be 'top5' first. Although maybe this changed as of python 3.7 according to google :)

brianphaley (approved):

hotsos/core/ycheck/events.py
dosaboy: I think it is technically not possible to guarantee but seems to be fairly consistently in that order in py38

mustafakemalgilor: I believe Python 3.7+ guarantees that the dict class will [keep the insertion order](https://docs.python.org/3/whatsnew/3.7.html#:~:text=the%20insertion%2Dorder%20preservation%20nature%20of%20dict%20objects%20has%20been%20declared%20to%20be%20an%20official%20part%20of%20the%20Python%20language%20spec.). I'd expect yaml output to preserve the order if the *_dump function is being called with `sort_keys=False`, or alternatively we can prepend a `_` to the total if the `sort_keys=True`

examples/hotsos-example-openstack.summary.yaml
```diff
@@ -109,8 +109,7 @@ openstack:
     - qemu-kvm 1:4.2-3ubuntu6.19
     - radvd 1:2.17-2
   neutron-l3ha:
-    backup:
-      - 984c22fd-64b3-4fa1-8ddd-87090f401ce5
+    backup: 1
```
mustafakemalgilor, line 112: I have no context here but shouldn't it be better if `backup` also contain a count/uuids list like `running` for the sake of uniformity?

dosaboy: i guess we could argue that having a long list of uuids for vms is also superfluous but it does generally tend to be more manageable

Issue #707, opened 2023-08-08
some ceph commands not working in binary mode


dosaboy:
The following commands are not working:

```
>>> with CLIHelperFile() as cli:
...   cli.ceph_status_json_decoded()
... 
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/home/ubuntu/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/hotsos/core/host_helpers/cli.py", line 622, in __call__
    fd.write(out.value)
TypeError: write() argument must be str, not dict
```

2023-08-01 16:31:16,882.882 977956 INFO plugin.storage [-] command 'ceph mgr_module_ls' exited with non-zero code '22'

The first is because CLIHelperFile needs to skip decode and the second is a command name typo.

PR #708, opened 2023-08-08
Fix binary ceph commands

dosaboy:
When using CLIHelperFile with binary command we must not decode json and also need to store in a seperate cache key since output of commands can differ depending on if CLIHelperFile or CLIHelper is used.

Resolves: #707


pponnuvel (approved):
LGTM

Issue #710, opened 2023-08-21
Detect OVN maintenance task errors


brianphaley:
LP #1948466 shows an OVN maintenance task error that we've seen triggered downstream, detect since a package upgrade is required.

PR #711, opened 2023-08-21
Add check for LP #1948466

brianphaley:
OVN maintenance task can throw a KeyError trying to delete a subnet, leading to neutron-server slowness.

Closes-bug: #710

### DIFF
--- a/hotsos/defs/scenarios/openstack/neutron/bugs/lp1948466.yaml
+++ b/hotsos/defs/scenarios/openstack/neutron/bugs/lp1948466.yaml
@@ -1,0 +1,15 @@
+checks:
+  has_1948466:
+    input:
+      path: 'var/log/neutron/neutron-server.log'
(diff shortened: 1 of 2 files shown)


xtrusia:
Hello, Just glanced at this commit. I have a question.
You inserted "Failed to fix deleted resource" string to check lp bug but I can't find the same string there.
It would be better if we can recognize where we can find the string above as well. or any error log examples.
Thanks.

brianphaley:
Hi,

This message is in the neutron neutron/plugins/ml2/drivers/ovn/mech_driver/ovsdb/maintenance.py check_for_inconsistencies() code. That was merged in Ussuri so should be present in all branches afterwards.

This is one of the log entries from neutron-server.log where I copied it from:

2023-08-16 09:31:52.883 87890 ERROR neutron.plugins.ml2.drivers.ovn.mech_driver.ovsdb.maintenance [req-cef1fdf1-c249-4d8c-ad0d-23d3a81f60ec - - - - -] Maintenance task: Failed to fix deleted resource 43fad483-a7b2-4cac-9f98-6f7e0f0e75b5 (type: subnets): KeyError: 'uuid'

There is also a full trackback, but this line is easier to detect.

I believe the LP bug output was from a tempest run which might have a slightly different log as well.

xtrusia (approved):

dosaboy (approved):

Issue #718, opened 2023-09-05
non-unicode chars in sos output causing search to fail


dosaboy:
I am getting errors like this when analysing a sosreport:

```
2023-09-05 20:31:48,403.403 109393 DEBUG searchkit [-] starting search of dump/sosreport-hostx-2023-07-17-cetemjv/sos_commands/systemd/systemctl_status_--all (offset=0, pos=0)
2023-09-05 20:31:48,427.427 109393 DEBUG searchkit [-] sequence cb1fa446-bc73-4a5a-b055-1b90c6df2719 started section e97c24d2-9b15-4857-a472-3f54f33a6ae1 (completed=0)
2023-09-05 20:31:48,427.427 109393 DEBUG searchkit [-] sequence cb1fa446-bc73-4a5a-b055-1b90c6df2719 stopping section e97c24d2-9b15-4857-a472-3f54f33a6ae1 (completed=1)
2023-09-05 20:31:48,434.434 109393 ERROR searchkit [-] 
Traceback (most recent call last):
  File "/home/user1/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/searchkit/search.py", line 1039, in execute
    fd.read(1)
  File "/usr/lib/python3.10/gzip.py", line 301, in read
    return self._buffer.read(size)
  File "/usr/lib/python3.10/_compression.py", line 68, in readinto
    data = self.read(len(byte_view))
  File "/usr/lib/python3.10/gzip.py", line 488, in read
    if not self._read_gzip_header():
  File "/usr/lib/python3.10/gzip.py", line 436, in _read_gzip_header
    raise BadGzipFile('Not a gzipped file (%r)' % magic)
gzip.BadGzipFile: Not a gzipped file (b'\xe2\x97')

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user1/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/searchkit/search.py", line 1044, in execute
    stats = self._run_search(fd)
  File "/home/user1/.local/pipx/venvs/hotsos/lib/python3.10/site-packages/searchkit/search.py", line 990, in _run_search
    line = line.decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 13: invalid continuation byte
[...]

PR #722, opened 2023-09-07
Tolerate systemctl status non-unicode chars

dosaboy:
We now use decode('utf-8', errors="backslashreplace") in FileSearcher when searching sosreport systemctl status output that can contain non-unicode chars.

Resolves: #718

### DIFF
--- a/hotsos/core/host_helpers/systemd.py
+++ b/hotsos/core/host_helpers/systemd.py
@@ -83,7 +83,7 @@ class SystemdService(object):
         # NOTE: should consider getting service status directly rather than
         #       searching in all but currently do this to have parity with
         #       sosreport.
-        fs = FileSearcher()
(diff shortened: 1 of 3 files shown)


dosaboy:
> Nit that commit message doesn't match the code change:
> 
> FileSearcher(decode_errors='backslashreplace')

yeah i agree it is a bit confusing but the intention was to reflect the underlying change i.e. that using the above results in a decode('utf-8', errors='backslashreplace') i.e. https://github.com/dosaboy/searchkit/blob/b9e567433e5268b751e06c87181a891cadc702cd/searchkit/search.py#L992. I will ammend the commit msg to reflect this properly.

dosaboy:
@brianphaley ive updated the commit msg so hopefully makes more sense now

brianphaley (approved):

Issue #723, opened 2023-09-13
check for LP#2012284 - osinfo permission denied


rodrigogansobarbieri:
Add a check and print guidelines

PR #724, opened 2023-09-13
Add check for LP#2012284

rodrigogansobarbieri:
Error creating Nova VM with AppArmor set
to "enforce": osinfo permission denied

Closes: #723

### DIFF
--- a/hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
+++ b/hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
@@ -1,0 +1,18 @@
+checks:
+  has_2012284:
+    input:
+      path: 'var/log/nova/nova-compute.log'
+    expr: ".+Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory '/usr/share/osinfo': Permission denied"
+    hint: 'osinfo'
+conclusions:
+  lp2012284:
+    decision: has_2012284
+    raises:
+      type: LaunchpadBug
+      bug-id: 2012284
+      message: >-
+        known nova compute issue caused when package gir1.2-libosinfo-1.0 is installed
+        and apparmor is set to enforced. To fix the issue, the offending package needs to
+        be removed, or apparmor set to complain, or the rules would need to be added to the
+        apparmor profile (but this is not supported and can be overridden). Please refer
+        to the Launchpad #2012284 bug page more information.
--- a/hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
+++ b/hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
@@ -1,18 +1,31 @@
+vars:
+  nc_aa_mode: '@hotsos.core.host_helpers.apparmor.AAProfileFactory.mode:/usr/bin/nova-compute'
 checks:
   has_2012284:
     input:
       path: 'var/log/nova/nova-compute.log'
     expr: '([\d-]+) ([\d:]+).\d{3} .+Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory ./usr/share/osinfo.: Permission denied'
     hint: 'osinfo'
+  nc_in_enforce_mode:
+    varops: [[$nc_aa_mode], [eq, enforce]]
+  has_package:
+    or:
+      - apt: gir1.2-libosinfo-1.0
+      - apt: gir1.2-libosinfo-1.0:amd64
 conclusions:
   lp2012284:
-    decision: has_2012284
+    decision:
+        - has_package
+        - has_2012284
+        - nc_in_enforce_mode
     raises:
       type: LaunchpadBug
       bug-id: 2012284
       message: >-
-        Known nova compute issue caused when package gir1.2-libosinfo-1.0 is installed
+        Known nova compute issue caused when package {package} is installed
         and apparmor is set to enforce. To fix the issue, the offending package needs to
         be removed, or apparmor set to complain, or the rules would need to be added to the
         apparmor profile (but this is not supported and can be overridden). Please refer
         to the Launchpad #2012284 bug page more information.
+      format-dict:
+        package: '@checks.has_package.requires.package'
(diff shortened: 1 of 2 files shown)


hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
```diff
@@ -0,0 +1,18 @@
+checks:
+  has_2012284:
+    input:
+      path: 'var/log/nova/nova-compute.log'
+    expr: ".+Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory '/usr/share/osinfo': Permission denied"
```
brianphaley, line 5: I think this should have a timestamp match, at least all the existing nova ones have something like this:

([\d-]+) ([\d:]+)\.\d{3} .+Failure...

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
```diff
@@ -0,0 +1,18 @@
...
+    hint: 'osinfo'
+conclusions:
+  lp2012284:
+    decision: has_2012284
+    raises:
+      type: LaunchpadBug
+      bug-id: 2012284
+      message: >-
+        known nova compute issue caused when package gir1.2-libosinfo-1.0 is installed
```
brianphaley, line 14: Known (although most of the other scenario messages are lower case :-/ )

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
```diff
@@ -0,0 +1,18 @@
...
+        and apparmor is set to enforced. To fix the issue, the offending package needs to
```
brianphaley, line 15: I think it's actually "enforce" (without the d), at least according to the bug

brianphaley (changes requested):

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
rodrigogansobarbieri: done. However, I had to change the quotation marks and the ' in the message to a . because they cannot be escaped in yaml (as far as I could find)

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
rodrigogansobarbieri: done

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
rodrigogansobarbieri: done

hotsos/defs/scenarios/openstack/nova/bugs/lp2012284.yaml
```diff
@@ -0,0 +1,18 @@
+checks:
+  has_2012284:
+    input:
+      path: 'var/log/nova/nova-compute.log'
+    expr: '([\d-]+) ([\d:]+).\d{3} .+Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory ./usr/share/osinfo.: Permission denied'
```
brianphaley, line 5: So now I see I commented the opposite in your other patch - let's wait for Ed to confirm which is correct.

dosaboy:
I would like the check to also check that the package is installed and/or enforce mode is set. If the issue gets fixed but the logs still exist we dont want to the issue to be raised.

dosaboy:
could also check whether or not nova-compute is in enforce or complain mode using https://github.com/canonical/hotsos/pull/751

hotsos/defs/tests/scenarios/openstack/nova/bugs/lp2012284.yaml
```diff
@@ -0,0 +1,13 @@
+data-root:
+  files:
+    var/log/nova/nova-compute.log: |
+      2023-02-14 11:54:31.094 3490324 ERROR nova.compute.manager [req-d8c21949-4edb-4ae8-859e-bdc407402446 919173c1ba5b04004ac4c467c678e6e842b90f5206a224168ba0d3d83c398dfb fd686745c7724189bb02e5f62020a1b2 - 7078ee187a1c42c2a798707b9ca4cd68 7078ee187a1c42c2a798707b9ca4cd68] [instance: ae35c62d-3f91-4e76-8274-9a7893b9627d] Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory '/usr/share/osinfo': Permission denied (14)
+  copy-from-original:
+    - sos_commands/date/date
+raised-bugs:
+  https://bugs.launchpad.net/bugs/2012284: >-
+    Known nova compute issue caused when package gir1.2-libosinfo-1.0 is installed
+    and apparmor is set to enforce. To fix the issue, the offending package needs to
```
dosaboy, line 10: now that the apparmor helper is available you can actually check what mode nova-compute is in e.g.

```
vars:
  nc_aa_mode: '@hotsos.core.host_helpers.apparmor.AAProfileFactory.mode:nova-compute'
checks:
  nc_in_enforce_mode:
    varops: [[$nc_aa_mode], [eq, enforce]]
...
``

File: hotsos/defs/tests/scenarios/openstack/nova/bugs/lp2012284.yaml [hunk 6]
```diff
@@ -0,0 +1,13 @@
+data-root:
+  files:
+    var/log/nova/nova-compute.log: |
+      2023-02-14 11:54:31.094 3490324 ERROR nova.compute.manager [req-d8c21949-4edb-4ae8-859e-bdc407402446 919173c1ba5b04004ac4c467c678e6e842b90f5206a224168ba0d3d83c398dfb fd686745c7724189bb02e5f62020a1b2 - 7078ee187a1c42c2a798707b9ca4cd68 7078ee187a1c42c2a798707b9ca4cd68] [instance: ae35c62d-3f91-4e76-8274-9a7893b9627d] Failure prepping block device: gi.repository.GLib.GError: g-io-error-quark: Error opening directory '/usr/share/osinfo': Permission denied (14)
+  copy-from-original:
+    - sos_commands/date/date
+raised-bugs:
+  https://bugs.launchpad.net/bugs/2012284: >-
+    Known nova compute issue caused when package gir1.2-libosinfo-1.0 is installed
```
dosaboy, line 9: If this package was to be uninstalled we would want to stop raising this issue if we saw the log message since that would be a false positive. Can we add a check that the package is installed?

rodrigogansobarbieri:
rebased and addressed review feedback

dosaboy:
lgtm thanks @rodrigogansobarbieri !

Issue #725, opened 2023-09-13
add check for LP#1761062


rodrigogansobarbieri:
and point to fix in LP#2019141

PR #728, opened 2023-09-15
Add check for LP#1761062

rodrigogansobarbieri:
Error migrating instances because
destination VM directory already exists.

Closes: #725

### DIFF
--- a/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
+++ b/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
@@ -4,14 +4,27 @@ checks:
       path: 'var/log/nova/nova-compute.log'
     expr: '([\d-]+) ([\d:]+).\d{3} .+DestinationDiskExists: The supplied disk path .+ already exists, it is expected not to exist'
     hint: 'DestinationDiskExists'
+  libvirt_rbd_backend:
+    config:
+      handler: hotsos.core.plugins.openstack.OpenstackConfig
+      path: etc/nova/nova.conf
+      assertions:
+        - key: images_type
+          section: libvirt
+          ops: [[eq, rbd]]
+          allow-unset: False
 conclusions:
   lp1761062:
-    decision: has_1761062
+    decision:
+      - has_1761062
+      - libvirt_rbd_backend
     raises:
       type: LaunchpadBug
       bug-id: 1761062
       message: >-
-        reverting VM resizes or invoking evacuations will leave the instance folder
-        left behind when using rbd, and then migrating the instance back fails. The
-        fix is to enable the config \'ensure_libvirt_rbd_instance_dir_cleanup\' in nova.
-        The charm-nova-compute fix is available at LP#2019141.
+        Reverting VM resizes or invoking evacuations will leave the instance
+        folder left behind when using rbd (which nova-compute is configured
+        to do by default), and then migrating the instance back fails. The
+        fix is to enable the config 'ensure_libvirt_rbd_instance_dir_cleanup'
+        in Nova. The nova-compute charm fix for this is available in the
+        referenced bug.
--- a/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
+++ b/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
@@ -1,0 +1,17 @@
+checks:
+  has_1761062:
+    input:
+      path: 'var/log/nova/nova-compute.log'
+    expr: '([\d-]+) ([\d:]+).\d{3} .+DestinationDiskExists: The supplied disk path .+ already exists, it is expected not to exist'
+    hint: 'DestinationDiskExists'
+conclusions:
+  lp1761062:
+    decision: has_1761062
+    raises:
+      type: LaunchpadBug
+      bug-id: 1761062
+      message: >-
+        reverting VM resizes or invoking evacuations will leave the instance folder
+        left behind when using rbd, and then migrating the instance back fails. The
+        fix is to enable the config \'ensure_libvirt_rbd_instance_dir_cleanup\' in nova.
+        The charm-nova-compute fix is available at LP#2019141.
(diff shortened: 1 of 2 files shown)


hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
```diff
@@ -0,0 +1,17 @@
+checks:
+  has_1761062:
+    input:
+      path: 'var/log/nova/nova-compute.log'
+    expr: '([\d-]+) ([\d:]+).\d{3} .+DestinationDiskExists: The supplied disk path .+ already exists, it is expected not to exist'
```
brianphaley, line 5: So I think to match others there should only be a single parens in the regex:

([\d-]+) [\d:]+\.\d{3}

Of course lp1967956.yaml looks just like this :-/ so we need to ask the oracle (Ed) since now I'm seeing a mix in neutron/bugs yamls as well.

brianphaley (changes requested):

hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
dosaboy: I'll try explain this. The answer to whether or not result groups to match timestamps are required lies in understanding how search constraints are applied. When you perform a search using the "search" property your search will be subject to constraints applied by the search engine (searchkit) and that does not require any groups at all since it will search for a date/time in each line before executing the search expression. You then have the option to apply extra constraints using [search.constraints](https://hotsos.readthedocs.io/en/latest/contrib/language_ref/property_ref/shared_properties.html#search) and this is where you need to ensure result groups - see https://github.com/canonical/hotsos/blob/61b7e1b93f168fde079045fc3d4b21fb0281b766/hotsos/core/ycheck/engine/properties/search.py#L204 but basically you need 1 group that matches the date and an optional second that matches the time. Hope that makes sense.

hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
```diff
@@ -0,0 +1,17 @@
...
+conclusions:
+  lp1761062:
+    decision: has_1761062
+    raises:
+      type: LaunchpadBug
+      bug-id: 1761062
+      message: >-
+        reverting VM resizes or invoking evacuations will leave the instance folder
+        left behind when using rbd, and then migrating the instance back fails. The
+        fix is to enable the config \'ensure_libvirt_rbd_instance_dir_cleanup\' in nova.
```
dosaboy, line 16: I'd like to propose the following modifications:

```
diff --git a/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml b/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
index 48cba92a..73fa78d8 100644
--- a/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
+++ b/hotsos/defs/scenarios/openstack/nova/bugs/lp1761062.yaml
@@ -4,14 +4,27 @@ checks:
       path: 'var/log/nova/nova-compute.log'
     expr: '([\d-]+) ([\d:]+).\d{3} .+DestinationDiskExists: The supplied disk path .+ already exists, it is expected not to exist'
     hint: 'DestinationDiskExists'
+  libvirt_rbd_backend:
+    config:
+      handler: hotsos.core.plugins.openstack.OpenstackConfig
+      path: etc/nova/nova.conf
+      assertions:
+        - key: images_type
+          section: libvirt
+          ops: [[eq, rbd]]
+          allow-unset: False
 conclusions:
   lp1761062:
-    decision: has_1761062
+    decision:
+      - has_1761062
+      - libvirt_rbd_backend
     raises:
       type: LaunchpadBug
       bug-id: 1761062
       message: >-
-        reverting VM resizes or invoking evacuations will leave the instance folder
-        left behind when using rbd, and then migrating the instance back fails. The
-        fix is to enable the config \'ensure_libvirt_rbd_instance_dir_cleanup\' in nova.
-        The charm-nova-compute fix is available at LP#2019141.
+        Reverting VM resizes or invoking evacuations will leave the instance
+        folder left behind when using rbd (which nova-compute is currently
+        configured to do), and then migrating the instance back fails. The
+        fix is to enable the config 'ensure_libvirt_rbd_instance_dir_cleanup'
+        in Nova. The nova-compute charm fixfor this is available in the
+        referenced bug.
diff --git a/hotsos/defs/tests/scenarios/openstack/nova/bugs/lp1761062.yaml b/hotsos/defs/tests/scenarios/openstack/nova/bugs/lp1761062.yaml
index 2151983d..8f84725f 100644
--- a/hotsos/defs/tests/scenarios/openstack/nova/bugs/lp1761062.yaml
+++ b/hotsos/defs/tests/scenarios/openstack/nova/bugs/lp1761062.yaml
@@ -1,12 +1,16 @@
 data-root:
   files:
+    etc/nova/nova.conf: |
+      [libvirt]
+      images_type = rbd
     var/log/nova/nova-compute.log: |
       2023-03-22 10:05:40.657 20779 ERROR oslo_messaging.rpc.server [req-e20743e3-a683-41ba-b47b-ba92e97eff37 4f7cd8bf676d43bc9faf09b2eb41482f 2c3d8251c39545cbb6f77f331b7164f8 - default default] Exception during message handling: DestinationDiskExists: The supplied disk path (/var/lib/nova/instances/d8db3f2a-cd8f-48e1-9951-012d762664b2) already exists, it is expected not to exist.
   copy-from-original:
     - sos_commands/date/date
 raised-bugs:
   https://bugs.launchpad.net/bugs/1761062: >-
-    reverting VM resizes or invoking evacuations will leave the instance folder
-    left behind when using rbd, and then migrating the instance back fails. The
-    fix is to enable the config \'ensure_libvirt_rbd_instance_dir_cleanup\' in nova.
-    The charm-nova-compute fix is available at LP#2019141.
+    Reverting VM resizes or invoking evacuations will leave the instance folder
+    left behind when using rbd (which nova-compute is currently configured to
+    do), and then migrating the instance back fails. The fix is to enable the
+    config 'ensure_libvirt_rbd_instance_dir_cleanup' in Nova. The
+    nova-compute charm fixfor this is available in the referenced bug.
```

Issue #730, opened 2023-09-24
Hotsos leaves behind temp files


brianphaley:
Running hotsos on a live system it will leave behind both a temp directory and log file in /var/tmp

$ hotsos
...
$ ls -rt /var/tmp | tail -2
tmpybcph5nx
tmpounx_jrkhotsos.log

It doesn't seem to do it running on a sosreport.

Changing the way it creates and removes its global temp dir should resolve the issue.

PR #731, opened 2023-09-24
Cleanup temporary files better

brianphaley:
Changed hotsos to setup its global temp directory
before setting up logging so that all files can
live in one place. Removal is now done in the code in cli() as well.

Code in client.py was cleaned-up accordingly as it was no longer needed.

Resolves: #730

### DIFF
--- a/hotsos/client.py
+++ b/hotsos/client.py
@@ -284,6 +284,13 @@ class OutputManager(object):
 
         return filtered
 
+    def temp_log_filename(self):
+        """
+        Return the temporary log filename if present
+        """
+        if log.handlers and isinstance(log.handlers[0], logging.FileHandler):
+            return log.handlers[0].baseFilename
+
     def _save(self, path, fmt, html_escape=None, minimal_mode=None,
               plugin=None):
         content = self.get(fmt=fmt, html_escape=html_escape,
@@ -372,6 +379,9 @@ class HotSOSClient(object):
         log.debug("tearing down global env")
         if os.path.exists(HotSOSConfig.global_tmp_dir):
             shutil.rmtree(HotSOSConfig.global_tmp_dir)
+        # If we don't clear this, callers of run() could accidentally
+        # recreate the temp directories we just removed
+        HotSOSConfig.plugin_tmp_dir = None
 
     def setup_plugin_env(self, plugin):
         """ State saved here is specific to a plugin. """
@@ -379,6 +389,15 @@ class HotSOSClient(object):
         global_tmp = HotSOSConfig.global_tmp_dir
         HotSOSConfig.plugin_tmp_dir = tempfile.mkdtemp(prefix=plugin,
                                                        dir=global_tmp)
+
+    def remove_temp_log_file(self):
+        temp_log = self.summary.temp_log_filename()
+        if not temp_log:
+            return
+        log.debug("removing temporary log file %s", temp_log)
+        logfile = os.path.abspath(temp_log)
+        if os.path.exists(logfile):
+            os.remove(logfile)
 
     def _run(self, plugin):
         if plugin not in PLUGIN_CATALOG:
@@ -400,6 +419,7 @@ class HotSOSClient(object):
         well as any extensions.
         """
         log.name = 'hotsos.client'
+        temp_log = self.summary.temp_log_filename()
         try:
             self.setup_global_env()
             for plugin in PLUGIN_RUN_ORDER:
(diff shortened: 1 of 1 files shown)


dosaboy (commented):
I guess this calls into question what the purpose of the temporary log file is. With this change the logfile gets deleted when hotsos finished and the environment is torn down so does not give the possibility of post-analysis. Having said that, leaving the file after each run is not ideal. Also in this change the teardown is not within a try/except so if the main plugin runner fails the teardown will not happen. I'd like to propose the following:
  * make sure teardown_env is within a try/except like it was originally so that we always delete the plugin data (which can get quite large)
  * leave the logfile where it is BUT delete it at the same time as teardown_env() IFF there is no exception raised otherwise print a message saying e.g. "an error occurred while running hotsos see logfile at XXX for details"

hotsos/client.py
```diff
@@ -410,4 +403,3 @@ def run(self):
                         self.summary.update(plugin, content.get(plugin))
         finally:
             log.name = 'hotsos.client'
-            self.teardown_global_env()
```
brianphaley, line 413: So I just wanted to make it clear this call *mostly* removed the temp files, but for some reason these kernel plugin ones were always left behind:

$ find tmpzssaz5uq
tmpzssaz5uq
tmpzssaz5uq/kernelva2q73j_
tmpzssaz5uq/kernelva2q73j_/locks
tmpzssaz5uq/kernelva2q73j_/locks/cache_commands.lock
tmpzssaz5uq/kernelva2q73j_/locks/cache_all_global.lock
tmpzssaz5uq/kernelva2q73j_/caches
tmpzssaz5uq/kernelva2q73j_/caches/host_helpers_cli
tmpzssaz5uq/kernelva2q73j_/caches/host_helpers_cli/commands
tmpzssaz5uq/kernelva2q73j_/caches/host_helpers_cli/commands/hostname

And 'kernel' is the last one run based on PLUGIN_RUN_ORDER.

Which actually led me to change the order and see what happens... and it shows the last one run always leaves these same files behind.

So maybe there is some better way to fix this?

I actually found the real problem. Above in get() there is this call:
    CLIHelper().hostname()

Which is going to write to the hostname file in the temp dir. Since HotSOSConfig.plugin_tmp_dir is set it writes into the last plugins temp dir, which in all cases is the kernel one, re-creating some of the tree there. Clearing that variable solves the issue, at least for the directory. That was not obvious at all.

I'll look at the log file comment you left and see how to deal with that.

hotsos/client.py
```diff
@@ -380,6 +387,19 @@ def setup_plugin_env(self, plugin):
         HotSOSConfig.plugin_tmp_dir = tempfile.mkdtemp(prefix=plugin,
                                                        dir=global_tmp)
 
+    def clear_plugin_env(self):
+        log.debug("clearing plugin env")
+        HotSOSConfig.plugin_tmp_dir = None
```
dosaboy, line 392: this should go into teardown_global_env()

hotsos/client.py
```diff
@@ -408,6 +429,15 @@ def run(self):
...
+            if temp_log:
+                print('See temp log file for possible details:', temp_log)
+            raise
         finally:
             log.name = 'hotsos.client'
+            self.clear_plugin_env()
             self.teardown_global_env()
+
+        # This will only get called if no exceptions
+        self.remove_temp_log_file()
```
dosaboy, line 443: this can go inside the ```try:``` clause at the end since we are executing it if no exception occured

hotsos/client.py
```diff
@@ -371,7 +378,7 @@ def setup_global_env(self):
     def teardown_global_env(self):
         log.debug("tearing down global env")
         if os.path.exists(HotSOSConfig.global_tmp_dir):
-            shutil.rmtree(HotSOSConfig.global_tmp_dir)
+            shutil.rmtree(HotSOSConfig.global_tmp_dir, ignore_errors=True)
```
dosaboy, line 381: why would we need ignore_errors?

hotsos/client.py
brianphaley: Done

hotsos/client.py
brianphaley: Done

hotsos/client.py
```diff
@@ -284,6 +284,13 @@ def get(self, fmt='yaml', html_escape=False, minimal_mode=None,
 
         return filtered
 
+    def temp_log_filename(self):
```
dosaboy, line 287: nit: i feel like this should live in hotsos.core.log but can do as a separate patch

dosaboy (approved):

Issue #715, opened 2023-08-23
Trying to create a hotsos check for lp1940549 


106106:
Hi hotsos team,

I'm trying to create a check for lp1940549. Essentially just looking for the vault snap on version less than 1.7. Here's what I have so far:

```
checks:
  has_lp1940549:
    snap:
      vault:
        # Looking for version less than 1.7
        - min: 1967
          max: 2039
conclusions:
  lp1940549:
    decision:
      - has_lp1940549
    raises:
      type: LaunchpadBug
      bug-id: 1940549
      message: >-
        Vault may not renewed or reissued certificates. This may
        lead to other services being degraded due to TLS issues.
        Upgrade to channel 1.7 or later. See the knowledge base
        article Vault HA cluster upgrade with Juju for upgrade steps.
```

I'm trying to check it on a sosreport that I know has version 1.5.9 of vault with:
```
hotsos --defs-path ~/ sosreport-vault.tar.xz
hotsos --defs-path ~/test.yaml sosreport-vault.tar.xz
```
But no luck. Any pointer would be appreciated.
Thanks!

dosaboy:
hi @106106, your way does not work because your defs path does not have directories for the plugins you want to run. Try putting test.yaml under ~/scenarios/vault/ e.g.

```
mkdir -p ~/scenarios/vault
mv ~/test.yaml ~/scenarios/vault
hotsos --defs-path ~/ sosreport-vault.tar.xz
```

This is explained in https://hotsos.readthedocs.io/en/latest/contrib/scenarios.html#scenarios-overview and if it is unclear let me know and I can update the documentation.

PR #734, opened 2023-09-27
Detect LP1940549 and fix typo

106106:
Hey Team, 

please let me know if I can make any improvements/fixes in these commits or in how I'm contributing.

# Summary of changes
- add scenario for https://bugs.launchpad.net/vault-charm/+bug/1940549
- add test for the scenario
- add directories to match the repo structure and clean up placeholder test
- fix typo in the documentation for snaps

### DIFF
--- a/hotsos/defs/scenarios/vault/prevent_lp1940549.yaml
+++ b/hotsos/defs/scenarios/vault/prevent_lp1940549.yaml
@@ -1,0 +1,15 @@
+checks:
+  prevent_lp1940549:
+    snap:
+      vault:
+        # Looking for version less than 1.7
+        - min: 1967
+          max: 2039
+conclusions:
+  prevent_lp1940549:
+    decision:
+      - prevent_lp1940549
+    raises:
+      type: SystemWarning
+      message: >-
+        The installed Vault charm version suffers from a bug, see lp1940549, that impedes renewal or reissuing of certificates for services dependent on Vault. This can lead to other services being degraded due to TLS issues. It is strongly recommended to upgrade to channel 1.7 or later. See the knowledge base article 'Vault HA cluster upgrade with Juju' for upgrade steps.
--- a/hotsos/defs/tests/scenarios/vault/placeholder-not-a-test.yaml
+++ b/hotsos/defs/tests/scenarios/vault/placeholder-not-a-test.yaml
@@ -1,6 +1,0 @@
-# THIS IS A PLACEHOLDER FILE AND SHOULD BE DELETED ONCE A TEST IS ADDED TO THIS DIRECTORY
-data-root:
-  files:
-  copy-from-original:
-raised-bugs:
-raised-issues:
--- a/hotsos/defs/tests/scenarios/vault/prevent_lp1940549.yaml
+++ b/hotsos/defs/tests/scenarios/vault/prevent_lp1940549.yaml
@@ -1,0 +1,10 @@
+target-name: prevent_lp1940549.yaml
+data-root:
+  files:
+    sos_commands/snap/snap_list_--all: |
+        Name                     Version    Rev    Tracking       Publisher    Notes
+        vault                    1.5.9      2012   1.5/stable     canonical**  -
+  copy-from-original:
+raised-issues:
+  SystemWarning: >-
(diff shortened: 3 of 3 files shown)


106106:
Looks like my test failed. Is this because it have the scenario in a different sub-path /vault instead of /vault/bugs?

106106:
I did some manual testing of the scenario locally (ps. thanks for the [help](https://github.com/canonical/hotsos/issues/715)) but I'm not clear on how I can run a test locally. I read these [docs](https://hotsos.readthedocs.io/en/latest/contrib/testing.html) but I don't have the background to follow. Could you please provide some more pointers?

nicolasbock:
@106106 Please re-open the pull request.

Regarding the failed test, please have a look at [the failure](https://github.com/canonical/hotsos/actions/runs/6320196147/job/17162351601#step:10:38). it looks like your test is missing a key but looks good otherwise.

doc/source/contrib/language_ref/property_ref/requirement_types.rst
```diff
@@ -103,8 +104,8 @@ In the above example *mypackage* must have a revision between 0.0 and 1.0 or
 
 Cache keys:
 
-  * package - name of each installed package
-  * revision - revision of each installed package
+  * snap - name of each installed snap
```
dosaboy, line 107: the cache key is actually called "package" not "snap" so you can leave this as is.

doc/source/contrib/language_ref/property_ref/requirement_types.rst
```diff
@@ -79,19 +79,20 @@ Cache keys:
 Snap
 ----
 
-Takes an apt package name and optional list of revision ranges. Returns True if
-the package exists and if provided, revision is within ranges.
```
dosaboy, line 83: Thank you for pointing this out. I have gone ahead and fixed it in https://github.com/canonical/hotsos/commit/ef40c2ef8961239a493a8deeab79a0c5dd9715f7 since both the apt and snap property docs needed tweaking. You can go ahead and rebase your branch to get the fix (or just remove your change).

hotsos/defs/scenarios/vault/bugs/lp1940549.yaml
```diff
@@ -0,0 +1,17 @@
...
+          max: 2039
+conclusions:
+  lp1940549:
+    decision:
+      - has_lp1940549
+    raises:
+      type: LaunchpadBug
+      bug-id: 1940549
+      message: >-
+        Vault may not renewed or reissued certificates. This may lead to other services being degraded due to TLS issues.
```
dosaboy, line 16: The issue described in LP 1940549 is about the fact that the vault charm sometimes does not propagate certificates on charm relations correctly but is not a bug in vault itself. I do not feel that this message accurately depicts that and I'm also not sure that there is a way to accurate detect that the problem as occurred from a sosreport since the issue manifests on the remote side i.e. other charms that consume the certs.

dosaboy: Also fwiw, you would need to squash your commit into a single commit.

106106:
Ah, good point. My goal is to create a warning to proactively prevent this bug/issue from occurring. Not to detect its occurrence. We've noticed multiple high severity incidents from customers that have been exacerbated or triggered by this difficulty in reissuing certificates. Is this type of preventative warning in the scope of hotsos?

If so, perhaps it would be better to raise a warning and mention the bug in the description instead?
and re-word the line to something like?

```
The installed Vault charm version suffers from a bug, see lp1940549, that impedes renewal or     reissuing of certificates for services dependent on Vault. This can lead to other services being de    graded due to TLS issues. It is strongly recommended to upgrade to channel 1.7 or later. See the kno    wledge base article 'Vault HA cluster upgrade with Juju' for upgrade steps.
```

106106:
I adjusted the scenario and test. Also made the changes a single commit. I'm still a newbie with git so please let me know if there's anything that could be done better.

dosaboy:
It is absolutely in the scope of hotsos to provide preventative warnings but my point is that the version of vault snap installed is not an indication that the charm contains the bug or that the host is affected. To do that you would need to get the version of charm installed (which you **might** be able to get from https://github.com/canonical/hotsos/blob/main/hotsos/core/plugins/juju/resources.py#L120 but I know the presence and location of that information depends somewhat on (a) sosreport version and (b) juju version). I don't what every user running vault 1.7 to get this warning as that will generally be a false positive.

dosaboy:
This pull request has not received any updates in over a month so I will go ahead and close the pull request. If you would like to propose an update please re-open the pull request.

Issue #729, opened 2023-09-22
Spinner does not work when hotsos run on tar file


brianphaley:
Hotsos now has support for running directly on a tar file, but it's "spinner" code doesn't work, looks something like this in an 80-column window:

-> hotsos sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa.tar.xz
INFO: extracting sosreport sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa.tar.xz to /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-0
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-0
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-0
(repeated a gazillion times)
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-0
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-0
INFO: analysing sosreport /var/tmp/tmp1pboi7vt/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa |
hotsos:
  version: 1.16.1.post8
  repo-info: cd3edffd
...

I tried on another system to make sure it wasn't something in my environment.

Extracting and running on the resultant directory works fine for the same report.

dosaboy:
i get the same if my terminal size is smaller than the line length but that's because the spinner is at the end of the line

brianphaley:
Right, it's my old-fashioned 80-column terminal. I think we can change this to print the file/dir name on its own line, followed by a short "analysing" statement.

PR #736, opened 2023-09-27
Fix spinner message printing multiple times

brianphaley:
When analysing a tarfile, attempting to print the
"INFO: analysing $file" line in the spinner code
can overflow the window if it's narrow, resulting
in hundreds of lines of extra output.

Change to print what is being analysed on its own
line to keep the "analysing" spinner message short and prevent the extra output.

Resolves: #729

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -309,7 +309,10 @@ def main():
                     spinner_msg = ''
                 else:
                     show_spinner = not debug
-                    spinner_msg = 'INFO: analysing {} '.format(drm.name)
+                    if show_spinner:
+                        sys.stdout.write(
+                            'INFO: analysing {}\n'.format(drm.name))
+                    spinner_msg = 'INFO: analysing '
 
                 with progress_spinner(show_spinner, spinner_msg):
                     plugins = []
--- a/tools/test/functest.sh
+++ b/tools/test/functest.sh
@@ -47,12 +47,12 @@ test_plugin ()
         args+=( --short )
         label=".short"
     fi
-    # NOTE: we remove repo-info and date from hotsos and system plugin
+    # NOTE: we remove repo-info, date and INFO from hotsos and system plugin
     #       output since they are liable to change.
     ./scripts/hotsos --${plugin} ${args[@]} $data_root 2>/dev/null| \
-        egrep -v "^  repo-info:|date:" > $dtmp/$plugin$label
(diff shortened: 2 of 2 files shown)


nicolasbock (commented):
The change looks good to me, but for some reason the functional tests are all failing.

hotsos/cli.py
```diff
@@ -310,7 +310,8 @@ def cli(data_root, version, defs_path, templates_path, all_logs, debug,
                 spinner_msg = ''
             else:
                 show_spinner = not debug
-                spinner_msg = 'INFO: analysing {} '.format(drm.name)
+                sys.stdout.write('INFO: analysing {}\n'.format(drm.name))
```
nicolasbock, line 313: Since the spinner is only shown if `show_spinner == True` you could put this
line in an `if` statement so that we don't end up with an additional line when
usingn `--debug`.

```suggestion
                if show_spinner:
                    sys.stdout.write('INFO: analysing {}\n'.format(drm.name))
```

brianphaley: Hi Nick,

I was just going to make this change but just ran it to check output beforehand.

--> ./scripts/hotsos --debug ~/downloads/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa.tar.xz 
INFO: extracting sosreport /home/bhaley/downloads/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa.tar.xz to /var/tmp/tmpxprc8280/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa
INFO: analysing sosreport /var/tmp/tmpxprc8280/sosreport-juju-1cb930-51-lxd-18-00368438-2023-09-20-lowcoqa
...

I'm going to tweak this as you said, but if we wanted the other INFO on un-packing gone with --debug will require another change, probably just passing debug to DataRootManager would be enough. I don't have an opinion either way.

nicolasbock: I think what "irked" me (if that's the right way of saying this :smile:) was the fact that you had to modify the regular expression in the `functest.sh`. But after closer inspection I realized that the `Spinner` does not print if it doesn't detect a tty.

I like your change though. Just don't start using a 20 character wide terminal all of a sudden 😜

nicolasbock (approved):

PR #738, opened 2023-10-04
Allow tar extraction failures

dosaboy:
If one or more tar member fails to extract e.g. due to permissions errors, we continue extraction but
log a message to say it may be incomplete.

### DIFF
--- a/hotsos/core/root_manager.py
+++ b/hotsos/core/root_manager.py
@@ -72,7 +72,18 @@ class DataRootManager(object):
                 if not os.path.exists(target):
                     sys.stdout.write("INFO: extracting sosreport {} to {}\n".
                                      format(path, target))
-                    tar.extractall(self.tmpdir)
+                    try:
+                        tar.extractall(self.tmpdir)
+                    except Exception:
+                        # some members might fail to extract e.g. permission
+                        # denied but we dont want that to cause the whole run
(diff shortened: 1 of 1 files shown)


hotsos/core/root_manager.py
```diff
@@ -72,7 +72,18 @@ def data_root(self):
...
+                        tar.extractall(self.tmpdir)
+                    except Exception:
+                        # some members might fail to extract e.g. permission
+                        # denied but we dont want that to cause the whole run
+                        # to fail so we just log a message and ignore them.
+                        tar.errorlevel = 0
+                        sys.stdout.write("INFO: one or more members failed to "
+                                         "extract - disabling error mode and "
+                                         "continuing extraction (which will "
+                                         "be incomplete as a result)\n")
```
brianphaley, line 85: Don't know if it's worth it to print the exception (except Exception as exc, ...) but this probably doesn't happen often.

brianphaley (approved):

hotsos/core/root_manager.py
dosaboy: as things stand currently the logger is not yet setup by this point but its not a bad idea so ill aim to fix that in a separate patch

dosaboy: opened https://github.com/canonical/hotsos/issues/739 for ^^

PR #745, opened 2023-10-09
Add support for the Openstack Bobcat release

dosaboy:


hotsos/core/plugins/openstack/openstack.py
```diff
@@ -31,6 +31,7 @@
 # NOTE(tpsilva): when updating this, refer to the Charmed Openstack supported
 # versions page: https://ubuntu.com/openstack/docs/supported-versions
```
brianphaley, line 32: I was able to verify the versions here consulting this page:

https://openstack-ci-reports.ubuntu.com/reports/cloud-archive/bobcat_versions.html

Don't know if it's worth it to put a link to the parent page here as I only knew about that based on a ML post from Corey.

https://openstack-ci-reports.ubuntu.com/reports/cloud-archive/index.html

nicolasbock (approved):

hotsos/core/plugins/openstack/openstack.py
dosaboy: yeah good point ill add that thanks brian

Issue #747, opened 2023-10-16
Ensure neutron qr ports have mtu less than physical network


dosaboy:
Could also check that they are less than that allowed by tunnel (vxlan, gre etc) i.e. with overheads subtracted - https://github.com/openstack/neutron-lib/blob/bdebe1de3c9e4032a7a40785846065b676b45b64/neutron_lib/constants.py#L346

PR #748, opened 2023-10-16
Check neutron router port mtus

dosaboy:
Adds a check to ensure that ml2-ovs router ports have an mtu less than the physical network and raise an OpenstackWarning if any found not to match this. Also adds neutron port mtu info to summary.

Resolves: #747

### DIFF
--- a/hotsos/core/plugins/openstack/neutron.py
+++ b/hotsos/core/plugins/openstack/neutron.py
@@ -10,6 +10,11 @@ from hotsos.core.plugins.openstack.openstack import (
     OSTServiceBase,
 )
 
+# See https://github.com/openstack/neutron-lib/blob/master/neutron_lib/constants.py#L346  # noqa, pylint: disable=C0301
+IP_HEADER_BYTES = 20
+GRE_HEADER_BYTES = 22
+VXLAN_HEADER_BYTES = 30
+
 
 class NeutronBase(OSTServiceBase):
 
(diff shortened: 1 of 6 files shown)


hotsos/core/plugins/openstack/neutron.py
```diff
@@ -10,6 +10,11 @@
     OSTServiceBase,
 )
 
+# See https://github.com/openstack/neutron-lib/blob/master/neutron_lib/constants.py#L346  # noqa, pylint: disable=C0301
+IP_HEADER_BYTES = 20
+GRE_HEADER_BYTES = 28
```
brianphaley, line 15: I think this is 22

hotsos/plugin_extensions/openstack/service_network_checks.py
```diff
@@ -69,6 +79,53 @@ def get_phy_port_health_info(self):
...
+
+            for ns in self.cli.ip_netns():
+                # strip index
+                ns = ns.partition(" ")[0]
+                for nsprefix in ['qrouter', 'snat']:
+                    if not ns.startswith(nsprefix):
+                        continue
+
+                    for port in HostNetworkingHelper().get_ns_interfaces(ns):
+                        for pprefix in ['qr']:
```
brianphaley, line 100: In the snat- namespaces there will be sg-* interfaces with this reduced mtu as well.

hotsos/core/plugins/openstack/neutron.py
dosaboy: doh! thanks brian

hotsos/plugin_extensions/openstack/service_network_checks.py
dosaboy: good call. will fix

Issue #750, opened 2023-10-17
add apparmor helper


dosaboy:
this is so that we can have an easy way to check the mode of a service in checks/scenarios e.g. if service X is on complain or enforce mode.

PR #751, opened 2023-10-17
Add a new apparmor helper

dosaboy:
Resolves: #750

### DIFF
--- a/hotsos/core/host_helpers/apparmor.py
+++ b/hotsos/core/host_helpers/apparmor.py
@@ -1,0 +1,78 @@
+from functools import cached_property
+
+# NOTE: we import direct from searchkit rather than hotsos.core.search to
+#       avoid circular dependency issues.
+from searchkit import (
+    FileSearcher,
+    SearchDef,
+    SequenceSearchDef,
+)
+from hotsos.core.factory import FactoryBase
+from hotsos.core.host_helpers.cli import CLIHelperFile
+
+
+class AAProfile(object):
+
+    def __init__(self, name):
+        self.name = name
+        self.mode = ApparmorHelper().get_profile_mode(name)
+
+
+class ApparmorHelper(object):
+
+    @cached_property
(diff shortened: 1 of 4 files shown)


hotsos/core/host_helpers/apparmor.py
```diff
@@ -0,0 +1,78 @@
...
+            if name in profiles['profiles']:
+                return mode
+
+    @property
+    def profiles_enforce(self):
+        return self.profiles.get('enforce', {}).get('profiles', [])
+
+    @property
+    def profiles_complain(self):
+        return self.profiles.get('complain', {}).get('profiles', [])
```
brianphaley, line 69: I also see 'kill' and 'unconfined' profiles should we worry about those?

hotsos/core/host_helpers/apparmor.py
```diff
@@ -0,0 +1,78 @@
...
+)
+from hotsos.core.factory import FactoryBase
+from hotsos.core.host_helpers.cli import CLIHelperFile
+
+
+class AAProfile(object):
+
+    def __init__(self, name):
+        self.name = name
+        self.mode = ApparmorHelper().get_profile_mode(name)
```
brianphaley, line 18: There is a 'processes' section in apparmor_status should that be added?

dosaboy: i dont have a current need for it (was adding to help with https://github.com/canonical/hotsos/pull/724) but we can always extend later if needs be.

hotsos/core/host_helpers/apparmor.py
dosaboy: sure why not better to have them than not

dosaboy: actually i think unconfined is for processes not profiles so i think the only two statuses you can have are "complain" and "enforce"

PR #755, opened 2023-10-20
Add check for LP#2039444

### LINKED ISSUE #754: Add a check for LP#2039444

rodrigogansobarbieri:
Closes #754

### DIFF
--- a/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
+++ b/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
@@ -1,0 +1,16 @@
+checks:
+  has_2039444:
+    input:
+      path: 'var/log/juju/*mysql*.log'
+    start: '([\d-]+) ([\d:]+) .+ Failed adding instance .+ to cluster'
+    end: 'mysqlsh\.DBError: MySQL Error \(\d+\): Cluster\.add_instance: Could not open connection to .+: Host .+ is not allowed to connect to this MySQL server'
+conclusions:
+  lp2039444:
+    decision: has_2039444
+    raises:
+      type: LaunchpadBug
+      bug-id: 2039444
+      message: >-
+        The action run on the mysql leader failed to add a mysql instance.
+        This may be caused if the leader is the highest numbered juju
+        unit. See LP#2039444 for more information.
--- a/hotsos/defs/tests/scenarios/mysql/bugs/lp2039444.yaml
+++ b/hotsos/defs/tests/scenarios/mysql/bugs/lp2039444.yaml
@@ -1,0 +1,16 @@
+data-root:
+  files:
+    var/log/juju/unit-mysql-innodb-cluster-2.log: |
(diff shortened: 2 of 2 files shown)


rodrigogansobarbieri:
Pushed code as requested by Ed so he can take a look on the problem of having multi-line error messages where one does not have a timestamp

mustafakemalgilor (approved):
LGTM :shipit:

hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
```diff
@@ -0,0 +1,16 @@
+checks:
+  has_2039444:
+    input:
+      path: 'var/log/juju/*mysql*.log'
+    start: '([\d-]+) ([\d:]+) .+ Failed adding instance .+ to cluster'
+    end: '[\s\S]+mysqlsh\.DBError: MySQL Error \(\d+\): Cluster\.add_instance: Could not open connection to .+: Host .+ is not allowed to connect to this MySQL server'
```
mustafakemalgilor, line 6: I reckon the `[\s\S]+` is redundant, and that might be the reason why the CI fails. You can re-run the failing test locally by running `tox -e py3 -- test_bugs_lp2039444`

dosaboy: yes that's correct, the mysqlsh line in your test has nothing before "mysqlsh..." hence why that does not match. I have also landed support for sequence searches in scenario checks so please rebase your patch. Those two things should result in a a pass (i have tested it and it works).

```
diff --git a/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml b/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
index a59b7065..68bcdf12 100644
--- a/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
+++ b/hotsos/defs/scenarios/mysql/bugs/lp2039444.yaml
@@ -3,7 +3,7 @@ checks:
     input:
       path: 'var/log/juju/*mysql*.log'
     start: '([\d-]+) ([\d:]+) .+ Failed adding instance .+ to cluster'
-    end: '[\s\S]+mysqlsh\.DBError: MySQL Error \(\d+\): Cluster\.add_instance: Could not open connection to .+: Host .+ is not allowed to connect to this MySQL server'
+    end: 'mysqlsh\.DBError: MySQL Error \(\d+\): Cluster\.add_instance: Could not open connection to .+: Host .+ is not allowed to connect to this MySQL server'
 conclusions:
   lp2039444:
     decision: has_2039444
```

rodrigogansobarbieri: rebased and addressed Ed's review feedback. @mustafakemalgilor : Ed had to improve hotsos to support multi-line regex search. My attempt with [\s\S]+ would work with native python regex.

dosaboy:
lgtm, the message could be worded a bit better but i can propose a separate patch for that. Thanks @rodrigogansobarbieri !

PR #757, opened 2023-10-23
scenarios/ost/keystone/lp1896125: better detection of the LP#1896125

### LINKED ISSUE #756: Improve detection of LP#1896125
We had another instance of this bug happening, but the current detection method failed to recognize it. The [scenario](https://github.com/canonical/hotsos/blob/main/hotsos/defs/scenarios/openstack/keystone/bugs/lp1896125.yaml) needs to be reworked.

mustafakemalgilor:
the previous revision of the scenario was tied to the specific keystone versions which impedes the scenario's ability to detect the bug in other keystone versions. the new revision will detect the bug regardless of the keystone version.

Closes #756


dosaboy:
@mustafakemalgilor thanks for the patch. getting rid if the config check makes sense (for the reason you explained) but I wonder if we could somehow keep the version check. Once the installed package is upgraded we would not want to raise the issue if we saw the log message, or at least we would want to use a different message perhaps to say that package has version X which is supposed to be fixed but we are still seeing the problem.

dosaboy:
@mustafakemalgilor shall we revisit this?

mustafakemalgilor:
@dosaboy Yes, I'll submit a new revision using the latest version syntax.

mustafakemalgilor:
@dosaboy should be good to go.

Issue #752, opened 2023-10-19
detect nova cc charm failure to update compute known_hosts


dosaboy:
The nova-cloud-controller charm collects address/hostname info from compute hosts and distributes ssh known_hosts so that they can access each other for vm migrations etc. In order to succeed at this the charm needs to be able to access the compute hosts on the network for which it is adding an entry and if it cannot it will log an error such as:

2023-10-19 13:27:37 ERROR unit.nova-cloud-controller/0.juju-log server.go:316 Could not obtain SSH host key from 10.11.12.2: reason: Command '['ssh-keyscan', '-H', '-t', 'rsa', '10.11.12.2']' returned non-zero exit status 1.

We should alert if we see this since it will means that known_hosts entries failed to update and could be stale or incomplete.

PR #758, opened 2023-10-24
Add check for nova-cc charm ssh known_hosts fail

dosaboy:
Resolves: #752


brianphaley (approved):
Looks like it should catch IPv6 addresses as well :)

Issue #763, opened 2023-10-31
add check for ovn minor release upgrades with ovn-match-northd-version set to true


dosaboy:
The ovn charms default to setting ovn-match-northd-version=true due to bug https://bugs.launchpad.net/charm-ovn-chassis/+bug/1940043 but this is really to anticipate major version upgrades. Unfortunately it also applies with minor version upgrades i.e. going from 20.03.2 to 20.03.3 will give a version mismatch error in ovn-controller. There is a charm patch going out in the next release to make this configurable [1] but we should still check this scenario and advise accordingly since it seems to be popping up more and more.

[1] https://github.com/openstack-charmers/charm-layer-ovn/commit/f68ecbd10d2109014c67100ec7e7fe8e1823a7ec

PR #768, opened 2023-11-01
Check for ovn northd version mismatch

dosaboy:
If the versions are from the same major release
and ovn-match-northd-version='true' in the local
db then we can suggest setting
ovn-match-northd-version='false' to unblock the
controller. Also adds a scenario to warn if
ovn-match-northd-version is set to true since
this should only be the case during a major release upgrade.

Resolves: #763


dosaboy:
need to ref https://bugs.launchpad.net/charm-ovn-chassis/+bug/2030944 in the checks

brianphaley (changes requested):
Review not done, just partial

alanbach (commented):
I went over the changes and they look good to me. Once merged and we have a test build, I can try this out in my home-lab.

Thank you so much for implementing this!

hotsos/defs/scenarios/openvswitch/ovn/ovn_upgrades.yaml
```diff
@@ -0,0 +1,97 @@
...
+    decision:
+      - is_ovn_controller
+      - not: has_northd_version_mismatch
+      - dbflag_is_true
+    raises:
+      type: LaunchpadBug
+      bug-id: 2030944
+      message: >-
+        This node has '{dbkey}=true' set in the local ovsdb. This
+        will cause unnecessary errors if a minor release upgrade of ovn packages is
```
brianphaley, line 35: nit: s/ovn/OVN

hotsos/plugin_extensions/openvswitch/event_checks.py
```diff
@@ -263,6 +264,42 @@ def __call__(self, event):
...
+            if dbval and dbval.lower() == 'true':
+                msg = ("ovn-controller is reporting northd version mismatch "
+                       "errors and the versions it is reporting ({vfrom} and "
+                       "{vto}) are from the same major release. This is "
+                       "failing because you have '{dbkey}' set to "
+                       "'true' in the local ovsdb which is only required if "
+                       "performing a major release upgrade. You can safely do "
+                       "the following to workaround the problem: 'ovs-vsctl "
+                       "set Open_vSwitch . "
+                       "external-ids:{dbkey}=\"false\"".
```
brianphaley, line 298: I think there is a missing ' here to match the one from 'ovs-vsctl

hotsos/defs/scenarios/openvswitch/ovn/ovn_upgrades.yaml
```diff
@@ -0,0 +1,97 @@
...
+        performed. There is fix available in the OVN charms to unset this or
+        alternatively it can be done manually with: ovs-vsctl set
+        Open_vSwitch . external-ids:{dbkey}="false"
```
brianphaley, line 38: nit: should put the ovs-vsctl command inside ' ' if possible

tests/unit/test_openvswitch.py
```diff
@@ -459,6 +465,30 @@ def test_ovs_defferred_action_limit_reached(self):
...
+        issues = list(IssuesStore().load().values())[0]
+        msg = ("ovn-controller is reporting northd version mismatch errors "
+               "and the versions it is reporting (22.03.3-20.21.0-62.4 and "
+               "22.03.0-20.21.0-58.3) are from the same major release. This "
+               "is failing because you have 'ovn-match-northd-version' set to "
+               "'true' in the local ovsdb which is only required if "
+               "performing a major release upgrade. You can safely do the "
+               "following to workaround the problem: 'ovs-vsctl set "
+               "Open_vSwitch . "
+               "external-ids:ovn-match-northd-version=\"false\"")
```
brianphaley, line 489: Missing closing ' here too

hotsos/defs/tests/scenarios/openvswitch/ovn/ovn_upgrades_no_mismatch_dbflag_set.yaml
```diff
@@ -0,0 +1,27 @@
...
+  copy-from-original:
+    - uptime
+    - sos_commands/date/date
+raised-bugs:
+  https://bugs.launchpad.net/bugs/2030944: >-
+    This node has 'ovn-match-northd-version=true' set in the local ovsdb. This
+    will cause unnecessary errors if a minor release upgrade of ovn packages is
+    performed. There is fix available in the OVN charms to unset this or
+    alternatively it can be done manually with: ovs-vsctl set
+    Open_vSwitch . external-ids:ovn-match-northd-version="false"
```
brianphaley, line 27: Same comment here regarding putting this in ' '

brianphaley (changes requested):

jrodrigu-canonical:
For the newly deployed nova-compute on an existing OVN setup, you would get the following warning from hotsos:

`OVN controller is reporting a version mismatch between itself (22.03.3-20.21.0-62.4) and northd (22.03.0-20.21.0-58.3). This usually means ovn-central (ovn-northd etc) services have been upgraded before ovn-controller using a version of ovn that does not support this. To fix this issue you may need to upgrade ovn-controller on this host (current=22.03.3-0ubuntu0.22.04.1~cloud0).`

In this scenario, this would mean that ovn-controller was upgraded to a higher minor version than ovn-central. And the fix would be to upgrade ovn-central to 22.03.3 or set external-ids:ovn-match-northd-version="false"

As far as I can see, we can have 4 scenarios with OVN versions:

1. ovn-central has a greater major version than ovn-controller: the fix is to upgrade ovn-controller to the same major version
2. ovn-controller has a greater major version than ovn-central: the fix is to upgrade ovn-central to the same major version
3. ovn-central has a greater minor version than ovn-controller: the fix is to set external-ids:ovn-match-northd-version="false", or upgrade ovn-controller to the same version
4. ovn-controller has a greater minor version than ovn-central: the fix is to set external-ids:ovn-match-northd-version="false", or upgrade ovn-central to the same version

dosaboy:
hmm thats odd i was actually expecting it to match https://github.com/canonical/hotsos/blob/a9a33d906035e77124f6d9c57a8fb8ffade8a192/hotsos/defs/scenarios/openvswitch/ovn/ovn_upgrades.yaml#L41 in this case. I will look into why that is happening.

dosaboy:
hmm thats odd i was actually expecting it to match https://github.com/canonical/hotsos/blob/a9a33d906035e77124f6d9c57a8fb8ffade8a192/hotsos/defs/scenarios/openvswitch/ovn/ovn_upgrades.yaml#L41 in this case. I will look into why that is happening.

dosaboy:
@jrodrigu-canonical turns out turns out the output you got was from a previous version i.e. did not have this patch. if you update your hotsos snap/package to the latest (i just released the stable snap) you should get the new behaviour.

PR #769, opened 2023-11-01
host_helpers/packaging: support more operators in DPKGVersionCompare

mustafakemalgilor:
this patch adds the complete set of operators that "dpkg --compare-versions" support to the DPKGVersionCompare. also, DPKGVersionCompare will now throw DPKGBadVersionSyntax exception when an invalid version syntax is reported by the "dpkg --compare-versions".

the existing "min" and "max" are now aliases for "ge" and "le", respectively.

### DIFF
--- a/doc/source/contrib/language_ref/property_ref/requirement_types.rst
+++ b/doc/source/contrib/language_ref/property_ref/requirement_types.rst
@@ -72,6 +72,27 @@ Or with version ranges as follows:
 In the above example *mypackage* must have a version between 0.0 and 1.0 or
 4.0 and 5.0 inclusive.
 
+Another example:
+
+.. code-block:: yaml
+
+    apt:
+      mypackage:
+        - gt: 1.0
+          lt: 3.0
+        - eq: 5.0.3
+
+In the above example *mypackage* must have a version between 1.0 and 3.0
+(non-inclusive), or specifically 5.0.3.
+
+Supported operators:
+
+  * eq - equality comparison
+  * lt - less than (<)
+  * gt - greater than (>)
+  * le/max - less than or equal (<=)
+  * ge/min - greater than or equal (>=)
+
 Cache keys:
 
   * package - name of each installed package
--- a/hotsos/core/ycheck/engine/properties/requires/types/apt.py
+++ b/hotsos/core/ycheck/engine/properties/requires/types/apt.py
@@ -3,7 +3,7 @@ from functools import cached_property
 from hotsos.core.log import log
 from hotsos.core.host_helpers import (
     APTPackageHelper,
-    DPKGVersionCompare,
+    DPKGVersion,
 )
 from hotsos.core.ycheck.engine.properties.requires import (
     intercept_exception,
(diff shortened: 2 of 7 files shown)


hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,38 @@ def installed_versions(self):
 
         return _versions
 
-    def package_version_within_ranges(self, pkg, versions):
-        result = False
-        versions = sorted(versions, key=lambda i: str(i['min']), reverse=True)
+    def package_version_within_ranges(self, pkg, version_criterias):
```
nicolasbock, line 31: nit:

singular: criterion
plural: criteria

nicolasbock:
LGTM

hotsos/core/ycheck/engine/properties/requires/types/apt.py
mustafakemalgilor: I always assumed that the word 'criteria' was singular :) Thanks for the correction!

mustafakemalgilor: Done

dosaboy (commented):
lgtm thanks for implementing this, i will have one last pass before merging. i also think we should fully deprecate and replace all min/max in existing scenarios to make things consistent.

dosaboy:
I did a quick extra test and it seems it is not giving the desired result i.e. the below test fails:

```
+    @utils.create_data_root({'sos_commands/dpkg/dpkg_-l':
+                             'ii foo 2.2 amd64 blah'})
+    def test_apt_check_item_package_version_within_ranges_foo(self):
+        ci = apt.APTCheckItems('foo')
+        result = ci.package_version_within_ranges(
+            'foo',
+            [{'gt': '1'},
+             {'gt': '2.1',
+              'le': '2.1.1'},
+             {'gt': '3.2',
+              'lt': '3.5'}]
+        )
+        self.assertFalse(result)
+
```

This is because the current code is returning true for the first range that passes when it needs to consider that if a range has no upper bound and the next range has a lower bound, whether the version if within that range too

dosaboy:
If I ensure that an upper bound is provided in each range the logic works. I think that is actually how it works currently and it doesnt really make sense for ranges not to have an upper bound (max) so maybe we just make it mandatory.

mustafakemalgilor:
> I did a quick extra test and it seems it is not giving the desired result i.e. the below test fails:
> 
> ```
> +    @utils.create_data_root({'sos_commands/dpkg/dpkg_-l':
> +                             'ii foo 2.2 amd64 blah'})
> +    def test_apt_check_item_package_version_within_ranges_foo(self):
> +        ci = apt.APTCheckItems('foo')
> +        result = ci.package_version_within_ranges(
> +            'foo',
> +            [{'gt': '1'},
> +             {'gt': '2.1',
> +              'le': '2.1.1'},
> +             {'gt': '3.2',
> +              'lt': '3.5'}]
> +        )
> +        self.assertFalse(result)
> +
> ```
> 
> This is because the current code is returning true for the first range that passes when it needs to consider that if a range has no upper bound and the next range has a lower bound, whether the version if within that range too

restored the old behavior. the code should now behave as before.

dosaboy:
thanks! i will get on with reviewing post haste!

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,137 @@ def installed_versions(self):
...
-                v_max = str(item['max'])
-                lte_max = pkg_version <= DPKGVersionCompare(v_max)
-            else:
-                lte_max = True
+    def normalize_version_criteria(self, version_criteria):
+
+        # Step 0: Remove empty criteria
+        version_criteria = [x for x in version_criteria if len(x) > 0]
+
+        # Step 1: Replace legacy ops to new
```
mustafakemalgilor, line 36: nit: s/to/with/g

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,137 @@ def installed_versions(self):
...
+            "ge": lambda lhs, rhs: lhs >= DPKGVersionCompare(rhs),
+            "min": lambda lhs, rhs: ops["ge"](lhs, rhs),
+            "max": lambda lhs, rhs: ops["le"](lhs, rhs),
+        }
+
+        version_criteria = self.normalize_version_criteria(version_criteria)
+
+        for version_criterion in version_criteria:
+            # Each criterion is evaluated on its own
+            # so if any of the criteria is true, then
```
mustafakemalgilor, line 145: this statement is no longer true

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,137 @@ def installed_versions(self):
...
-        pkg_version = self.packaging_helper.get_version(pkg)
-        for item in versions:
-            v_min = str(item['min'])
-            if 'max' in item:
-                v_max = str(item['max'])
-                lte_max = pkg_version <= DPKGVersionCompare(v_max)
-            else:
-                lte_max = True
+    def normalize_version_criteria(self, version_criteria):
+
```
mustafakemalgilor, line 32: docstring

doc/source/contrib/language_ref/property_ref/requirement_types.rst
```diff
@@ -72,11 +72,32 @@ Or with version ranges as follows:
...
+
+In the above example *mypackage* must have a version between 1.0 and 3.0
+(non-inclusive), or specifically 5.0.3.
+
 Cache keys:
 
   * package - name of each installed package
   * version - version of each installed package
 
+Supported operators:
```
dosaboy, line 93: This section should move to the towards the top.

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,147 @@ def installed_versions(self):
...
+    def normalize_version_criteria(self, version_criteria):
+        """Normalize all the criterions in a criteria.
+
+        Normalization does the following:
+        - removes empty criteria
+        - replaces old ops with the new ones
+        - sorts each criterion(ascending) and criteria(descending)
+        - adds upper/lower bounds to criteria, where needed
+
+        Args:
```
dosaboy, line 40: I do have a plan to switch the codebase to Google style doctrings (I agree they are better) but for now lets please stick to the existing convention (javadoc) i.e. ```@param <name>: <description>, @return: <description>```

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,147 @@ def installed_versions(self):
...
+                if lop in version_criterion:
+                    version_criterion[nop] = version_criterion[lop]
+                    del version_criterion[lop]
+                    version_criteria[idx] = version_criterion
+
+        # Step 2: Sort each criterion in itself, so the smallest version
+        # appears first
+        for idx, version_criterion in enumerate(version_criteria):
+            version_criterion = dict(sorted(version_criterion.items(),
+                                     key=lambda item: str(item[1])))
```
dosaboy, line 63: I know this existing code is string sorting and that is my bad but it is in fact not always safe for package versions e.g.

```
$ dpkg --compare-versions 1.2 gt 1.2~1 && echo "yes"
yes
$ echo "print('1.2' > '1.2~1')"| python3 
False
```

dosaboy: maybe it is ok for now but need think about when this will get us into trouble and whether it is worth fixing now

hotsos/core/ycheck/engine/properties/requires/types/apt.py
```diff
@@ -28,32 +28,147 @@ def installed_versions(self):
...
+        version_criteria = [x for x in version_criteria if len(x) > 0]
+
+        # Step 1: Replace legacy ops with the new ones
+        legacy_ops = {"min": "ge", "max": "le"}
+        for idx, version_criterion in enumerate(version_criteria):
+            for lop, nop in legacy_ops.items():
+                if lop in version_criterion:
+                    version_criterion[nop] = version_criterion[lop]
+                    del version_criterion[lop]
+                    version_criteria[idx] = version_criterion
```
dosaboy, line 57: i don think this line (57) is needed? since version_criteria is already in version_criterion at position idx?

dosaboy:
lgtm, sorry this took so long to get merged...

PR #770, opened 2023-11-02
Improve ovn version mismatch checks

dosaboy:
Removes the event implementation and relies solely on scenarios with better wording to clearly cover
both cases of major and minor release upgrades.


jrodrigu-canonical:
It looks good to me!

PR #774, opened 2023-11-15
Added two more apparmor profiles

brianphaley:
On my system, apparmor_status also reports modes
'kill' and 'unconfined', so add them to the parser.

While testing I noticed the unit test kept matching the processes section in addition to the profiles
one, leading to more profiles than were in the list. Changing the regex to be stricter resolved the issue.

Updated the openstack fake_data_root to add these
profiles so they could be tested, even though that is frowned upon.

TrivialFix

### DIFF
--- a/hotsos/core/host_helpers/apparmor.py
+++ b/hotsos/core/host_helpers/apparmor.py
@@ -29,9 +29,9 @@ class ApparmorHelper(object):
         """
         s = FileSearcher()
         seqdef = SequenceSearchDef(
-                            start=SearchDef(r"(\d+) (\S+) are in (\S+) mode."),
-                            body=SearchDef(r"\s+(\S+)"),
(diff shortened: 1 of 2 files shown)


nicolasbock (approved):

tests/unit/fake_data_root/openstack/sos_commands/apparmor/apparmor_status
```diff
@@ -255,6 +255,8 @@ apparmor module is loaded.
    snap.lxd.user-daemon
    virt-aa-helper
 0 profiles are in complain mode.
+0 profiles are in kill mode.
+0 profiles are in unconfined mode.
```
dosaboy, line 259: Yeah I know it's a pain but if we start doing this it is a slippery slope. You could copy this file and append in your test maybe?

PR #783, opened 2024-02-07
[ceph] Link the KB to resolve large omaps

pponnuvel:


brianphaley (approved):
I guess putting a link to a KB is fine, it's done in one other place, even if some random person that downloads hotsos might not have access.

PR #787, opened 2024-02-20
Added detection support for LP# 1974138

alanbach:
This PR adds detection support for LP# 1974138.

### DIFF
--- a/hotsos/core/plugins/openstack/common.py
+++ b/hotsos/core/plugins/openstack/common.py
@@ -214,6 +214,22 @@ class OpenstackBase(object):
 
         return apache2_certificates_expiring
 
+    @cached_property
+    def apache2_allow_encoded_slashes_on(self):
+        """ Returns True if AllowEncodedSlashes On is found. """
+        if not self.ssl_enabled:
+            return False
+
+        with open(self.apache2_ssl_config_file) as fd:
+            for line in fd:
+                if line.strip().startswith('#'):
+                    continue
+
+                ret = re.search(r'[Aa]llow[Ee]ncoded[Ss]lashes\s+[Oo]n', line)
+                if ret:
+                    return True
+        return False
+
 
 class OpenstackChecksBase(OpenstackBase, plugintools.PluginPartBase):
     plugin_name = "openstack"
--- a/hotsos/defs/scenarios/storage/ceph/ceph-rgw/bugs/lp1974138.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-rgw/bugs/lp1974138.yaml
@@ -1,0 +1,22 @@
+checks:
+  node_is_ceph_rgw_and_beast_supported_version:
+    apt:
+      librgw2:
+        - min: 14.2.0
+  ssl_enabled:
+    property: hotsos.core.plugins.openstack.OpenstackBase.ssl_enabled
+  apache2_allow_encoded_slashes_on:
+    property: hotsos.core.plugins.openstack.OpenstackBase.apache2_allow_encoded_slashes_on
+conclusions:
+  lp1974138:
+    decision:
+      - node_is_ceph_rgw_and_beast_supported_version
+      - ssl_enabled
+      - not: apache2_allow_encoded_slashes_on
+    raises:
+      type: LaunchpadBug
+      bug-id: 1974138
+      message: >-
+        In multisite S3 RGW environments, objects containing prefixes (/),
+        do not sync from primary to secondary sites if AllowEncodedSlashes
+        Apache directive is missing or set to OFF.
--- a/tests/unit/test_openstack.py
+++ b/tests/unit/test_openstack.py
@@ -126,6 +126,18 @@ Listen 35347
 </VirtualHost>
 """
 
+APACHE2_SSL_CONF_RGW = """
+Listen 433
+<VirtualHost 10.5.2.135:433>
+    ServerName 10.5.100.2
+    SSLEngine on
+    SSLCertificateFile /etc/apache2/ssl/keystone/cert_10.5.100.2
+    SSLCertificateChainFile /etc/apache2/ssl/keystone/cert_10.5.100.2
+    SSLCertificateKeyFile /etc/apache2/ssl/keystone/key_10.5.100.2
+    AllowEncodedSlashes On
+</VirtualHost>
+"""
+
 CERTIFICATE_FILE = """
 Certificate:
     Data:
@@ -1301,3 +1313,17 @@ class TestOpenstackApache2SSL(TestOpenstackBase):
     def test_ssl_expiration_true(self):
         base = openstack_core.OpenstackBase()
         self.assertTrue(len(base.apache2_certificates_expiring), 1)
+
+    @utils.create_data_root(
+        {'etc/apache2/sites-enabled/openstack_https_frontend.conf':
+         APACHE2_SSL_CONF_RGW})
+    def test_1974138_apache2_allow_encoded_slashes_true(self):
+        base = openstack_core.OpenstackBase()
+        self.assertTrue(base.apache2_allow_encoded_slashes_on)
+
+    @utils.create_data_root(
+        {'etc/apache2/sites-enabled/openstack_https_frontend.conf':
+         APACHE2_SSL_CONF})
+    def test_lp1974138_apache2_allow_encoded_slashes_false(self):
+        base = openstack_core.OpenstackBase()
+        self.assertFalse(base.apache2_allow_encoded_slashes_on)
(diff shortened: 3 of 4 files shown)


hotsos/core/plugins/openstack/common.py
```diff
@@ -214,6 +214,20 @@ def apache2_certificates_expiring(self):
...
+        if not self.ssl_enabled:
+            return False
+
+        with open(self.apache2_ssl_config_file) as fd:
+            for line in fd:
+                ret = re.search(r'AllowEncodedSlashes On /(\S+)', line)
+                if ret:
+                    return True
+
+            return False
```
pponnuvel, line 229: Probably better to intend his return at `with` level so that it returns `False` if open fails too.

hotsos/defs/scenarios/storage/ceph/ceph-rgw/bugs/lp1974138.yaml
```diff
@@ -0,0 +1,24 @@
+checks:
+  node_is_ceph_rgw_and_beast_supported_version:
+    apt:
+      librgw2:
+        - min: 15.2.0
```
pponnuvel, line 5: Beast is default frontend since Nautilus. Do you mean use `14.2.0`?

hotsos/defs/scenarios/storage/ceph/ceph-rgw/bugs/lp1974138.yaml
```diff
@@ -0,0 +1,24 @@
...
+  ssl_enabled:
+    property: hotsos.core.plugins.openstack.OpenstackBase.ssl_enabled
+  apache2_allow_encoded_slashes_on:
+    property:
+      path: hotsos.core.plugins.openstack.OpenstackBase.apache2_allow_encoded_slashes_on
+      ops: [[eq, false]]
```
pponnuvel, line 11: You don't need to compare boolean property. You can simply drop this line and use "not" at the `conclusions` like:

```
      - node_is_ceph_rgw_and_beast_supported_version
      - ssl_enabled
      - not: apache2_allow_encoded_slashes_on
```

nicolasbock (approved):
Looks great @alanbach !

nicolasbock:
Thanks @alanbach ! Please squash those commits into one.

hotsos/core/plugins/openstack/common.py
```diff
@@ -214,6 +214,19 @@ def apache2_certificates_expiring(self):
...
 
+    @cached_property
+    def apache2_allow_encoded_slashes_on(self):
+        """ Returns True if AllowEncodedSlashes On is found. """
+        if not self.ssl_enabled:
+            return False
+
+        with open(self.apache2_ssl_config_file) as fd:
+            for line in fd:
+                ret = re.search(r'AllowEncodedSlashes On /(\S+)', line)
```
pponnuvel, line 225: The regex is wrong - that's why tests are failing. 

Since you don't look at anything after `On`, rest needs to be dropped. It's also better to account for multiple spaces between `AllowEncodedSlashes` and `On`. So it would be:
```
                ret = re.search(r'AllowEncodedSlashes\s+On', line)
```

Should it work for lowercase `on` too? In that case, you can use:
```
                ret = re.search(r'AllowEncodedSlashes\s+[Oo]n', line)
```

Likewise for `AllowEncodedSlashes`:
```
                ret = re.search(r'[Aa]llow[Ee]ncoded[Ss]lashes\s+[Oo]n', line)
```

tests/unit/test_openstack.py
```diff
@@ -1301,3 +1313,17 @@ def test_ssl_expiration_false(self):
     def test_ssl_expiration_true(self):
         base = openstack_core.OpenstackBase()
         self.assertTrue(len(base.apache2_certificates_expiring), 1)
+
+    @utils.create_data_root(
+        {'etc/apache2/sites-enabled/openstack_https_frontend.conf':
+         APACHE2_SSL_CONF_RGW})
+    def test_1974138_apache2_allow_encoded_slashes_true(self):
+        base = openstack_core.OpenstackBase()
+        self.assertTrue(base.apache2_allow_encoded_slashes_on, 1)
```
pponnuvel, line 1322: The second argument is a message on error, probably not what you intended. It can be:
```
self.assertTrue(base.apache2_allow_encoded_slashes_on)
```

tests/unit/test_openstack.py
```diff
@@ -1301,3 +1313,17 @@ def test_ssl_expiration_false(self):
...
+
+    @utils.create_data_root(
+        {'etc/apache2/sites-enabled/openstack_https_frontend.conf':
+         APACHE2_SSL_CONF})
+    def test_lp1974138_apache2_allow_encoded_slashes_false(self):
+        base = openstack_core.OpenstackBase()
+        self.assertTrue(base.apache2_allow_encoded_slashes_on, 0)
```
pponnuvel, line 1329: Using `assertFalse`:
```
self.assertFalse(base.apache2_allow_encoded_slashes_on)
```

pponnuvel (changes requested):

hotsos/core/plugins/openstack/common.py
alanbach: I did test these, although the build goes through and the tests pass, the resulting binary is not detecting the target LP when executed against my reference SOS report. Need to look deeper.

pponnuvel: @alanbach I can look take a look. Which sosreport it doesn't work for? Btw, you can run hotsos locally with your changes without building a binary.

dosaboy:
@alanbach as mentioned, please squash the commits into one and rebase to resolve conflicts. you also have these tests failing:

tests.unit.test_openstack.TestOpenstackApache2SSL.test_1974138_apache2_allow_encoded_slashes_true
tests.unit.test_openstack.TestOpenstackApache2SSL.test_lp1974138_apache2_allow_encoded_slashes_false

In general though the patch looks good and is following the correct approach for writing a scenario.

tests/unit/test_openstack.py
```diff
@@ -126,6 +126,30 @@
...
+Listen 433
+<VirtualHost 10.5.2.135:433>
+    ServerName 10.5.100.2
+    SSLEngine on
+    SSLCertificateFile /etc/apache2/ssl/keystone/cert_10.5.100.2
+    SSLCertificateChainFile /etc/apache2/ssl/keystone/cert_10.5.100.2
+    SSLCertificateKeyFile /etc/apache2/ssl/keystone/key_10.5.100.2
+    AllowEncodedSlashes On
+</VirtualHost>
+"""
```
dosaboy, line 151: looks like APACHE2_SSL_CONF_RGW is duplicated?

dosaboy (changes requested):

tests/unit/test_openstack.py
alanbach: Git sometimes makes you wonder :-D

alanbach: Duplicated CONF block has been resolved.

hotsos/core/plugins/openstack/common.py
alanbach: @pponnuvel I'm building the binary to make sure all the steps are clean and passed. 

I'm not quite sure if this regex actually works. I took a SOS report from one of my RGW units affected with this LP and running the built sosreport against it does not seem to detect and report the issue.

alanbach: I internally shared the SOS report

pponnuvel: @alanbach 

I ran it and it does detect:
```
storage:
  ceph:
    release:
      name: quincy
      days-to-eol: 2993
    services:
      systemd:
        enabled:
          - ceph-radosgw
        masked:
          - radosgw
      ps: 
        - radosgw (1) 
    dpkg:
      - ceph-common 17.2.6-0ubuntu0.22.04.3
      - python3-ceph-argparse 17.2.6-0ubuntu0.22.04.3
      - python3-ceph-common 17.2.6-0ubuntu0.22.04.3
      - python3-cephfs 17.2.6-0ubuntu0.22.04.3
      - python3-rados 17.2.6-0ubuntu0.22.04.3
      - python3-rbd 17.2.6-0ubuntu0.22.04.3
      - radosgw 17.2.6-0ubuntu0.22.04.3
  bugs-detected:
    https://bugs.launchpad.net/bugs/1974138: In multisite S3 RGW environments, objects
      containing prefixes (/), do not sync from primary to secondary sites if AllowEncodedSlashes
      Apache directive is missing or set to OFF.
````

pponnuvel (approved):

dosaboy (approved):

PR #793, opened 2024-03-01
Use project_copyright to avoid redefining a python builtin

pponnuvel:


doc/source/conf.py
```diff
@@ -7,7 +7,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = 'hotsos'
-copyright = '2023, edward.hope-morley@canonical.com'
+project_copyright = '2023, edward.hope-morley@canonical.com'
```
nicolasbock, line 10: The documentation says this is simply an alias to the `copyright` option. What is the advantage of changing it?

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-project_copyright

pponnuvel: Makes linters happy (Python's has [copyright](https://docs.python.org/3/library/constants.html#copyright))

nicolasbock: Ah, ok, I agree, that's a good thing :)

nicolasbock (approved):

PR #795, opened 2024-03-13
Fix tox command to run single test

wilkmar:
The existing tox command is missing '--' and fails to run properly. Added this tiny fix.


nicolasbock (approved):

dosaboy (approved):
good catch!

PR #801, opened 2024-04-02
Add huge pages check

wilkmar:
The patch adds logic for testing if there are too many huge pages defined in the system, which are not being fully utilised, but occupy memory which can't be used for any other purpose. Such situations can negatively impact performance or even lead to the out of memory situations.

This patch also introduces a small refactoring of the VMStat class in the hotsos/core/plugins/kernel/memory.py, in order to avoid code duplication (between VMStat and MemInfo classes).

### DIFF
--- a/hotsos/defs/scenarios/kernel/memory.yaml
+++ b/hotsos/defs/scenarios/kernel/memory.yaml
@@ -8,12 +8,26 @@ vars:
   # and is not necessarily representative of current state.
   min_compaction_success: 10000
   max_compaction_failures_pcent: 10
+  hugetlb_to_mem_total_percentage: '@hotsos.core.plugins.kernel.memory.MemInfo.hugetlb_to_mem_total_percentage'
+  mem_avail_to_mem_total_percentage: '@hotsos.core.plugins.kernel.memory.MemInfo.mem_avail_to_mem_total_percentage'
+  hugep_used_to_hugep_total_percentage: '@hotsos.core.plugins.kernel.memory.MemInfo.hugep_used_to_hugep_total_percentage'
+  mem_total_gb: '@hotsos.core.plugins.kernel.memory.MemInfo.mem_total_gb'
+  mem_available_gb: '@hotsos.core.plugins.kernel.memory.MemInfo.mem_available_gb'
+  hugetlb_gb: '@hotsos.core.plugins.kernel.memory.MemInfo.hugetlb_gb'
+  # Arbitrary thresholds set for the memory allocated for the huge
+  # pages to total memory and memory available to total memory.
+  hugetlb_to_mem_total_threshold_percent: 80
+  mem_available_to_mem_total_thershold_percent: 3
 checks:
   low_free_high_order_mem_blocks:
     varops: [[$nodes_with_limited_high_order_memory], [length_hint]]
   high_compaction_failures:
     - varops: [[$compact_success], [gt, $min_compaction_success]]
     - varops: [[$compaction_failures_percent], [gt, $max_compaction_failures_pcent]]
+  too_many_free_hugepages:
+    - property: hotsos.core.plugins.kernel.memory.MemInfo.huge_pages_enabled
+    - varops: [[$hugetlb_to_mem_total_percentage], [gt, $hugetlb_to_mem_total_threshold_percent]]
+    - varops: [[$mem_avail_to_mem_total_percentage], [lt, $mem_available_to_mem_total_thershold_percent]]
 conclusions:
   low_free_high_order_mem_blocks:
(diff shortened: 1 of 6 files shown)


hotsos/defs/scenarios/kernel/memory.yaml
```diff
@@ -44,4 +58,23 @@ conclusions:
...
+        the system maybe be suffering from having not enough available memory. Review /proc/meminfo,
+        your current huge pages configuration and check if you can allocate less memory for the
+        huge pages. For details, see the kernel documentation at
+        https://docs.kernel.org/admin-guide/mm/hugetlbpage.html.
+      format-dict:
+        hugetlb: $hugetlb_gb
+        hugepcent: $hugetlb_to_mem_total_percentage
+        mem_total: $mem_total_gb
+        mem_avail: $mem_available_gb
+        availthrsh: $mem_available_to_mem_total_thershold_percent
```
pponnuvel, line 79: I'd add the `e` here i.e. `availthresh` to be less confusing ;-)

pponnuvel (approved):
LGTM

pponnuvel:
A question for @dosaboy.

IIRC, the purpose of `hotsos/defs/tests/scenarios` is to replace unit tests for scenarios as much as possible. Had I done this patch, I'd have done a `copy-from-original` of `/proc/meminfo` in `hotsos/defs/tests/scenarios/kernel/hugepages.yaml` (no `mock` stuff) and not bothered with any changes in `tests/unit/test_kernel.py`. Do we still need to add unit tests for scenarios in `tests/unit/` As Marcin's done here?

dosaboy:
@pponnuvel yes that's correct you can rely entirely on unit tests written in yaml. The way this is written is totally fine though but as you say if the hugepages.yaml test were to use copy-from-original to copy across some real data root files, adding in overrides for the test then that would obviate the need for the mocking in this case and that would have the added benefit of the underlying code being tested as well (since it would not be mocked). I'll let @wilkmar have final call on that but yes I would recommend this approach.

wilkmar:
@dosaboy @pponnuvel updated test scenario .yaml according to you suggestions. Thanks for your feedback.

pponnuvel (approved):
LGTM. One comment - but not a real issue.

tests/unit/test_kernel.py
```diff
@@ -209,6 +217,28 @@ def test_slab_major_consumers(self):
                     'vmap_area (2700.0k)']
         self.assertEqual(top5, expected)
 
+    @utils.create_data_root(
```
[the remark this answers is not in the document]
wilkmar: @pponnuvel you are right. Just reverted all changes in 'tests/unit/test_kernel.py'

dosaboy:
@wilkmar you could go the extra step and have a "pass" and "fail" test i.e. rename your test to hugepages_fail.yaml (putting ```target-name: hugepages.yaml``` at the top of the file) and the add an extra test called hugepages_pass.yaml which contains a proc/meminfo that does not trigger the issue (i.e. raised-issues: would be an empty dict).

hotsos/defs/scenarios/kernel/memory.yaml
```diff
@@ -44,4 +58,23 @@ conclusions:
...
+      type: MemoryWarning
+      message: >-
+        This system is using huge pages which are consuming {hugetlb} GB, that is {hugepcent}% out
+        of {memtotal} GB total system memory. Available memory is only {memavail} GB, which is
+        less than {availthreshold}%. At the same time only {hpusedpcent}% of the huge pages are being
+        actively used. Memory reserved for the huges pages can't be used for any other purpose but
+        the system maybe be suffering from having not enough available memory. Review /proc/meminfo,
+        your current huge pages configuration and check if you can allocate less memory for the
+        huge pages. For details, see the kernel documentation at
+        https://docs.kernel.org/admin-guide/mm/hugetlbpage.html.
```
dosaboy, line 73: I'd like to propose reworking this slightly as follows:

```
This host is using hugepages which are consuming {hugetlb} out of {memtotal}GB total system memory i.e. {hugepcent}%. This leave {memavail}GB available memory for processes not using hugepages. Also, only {hpusedpcent}% of the hugepages are being actively used. Memory reserved for hugepages can't be used for any other purpose which may starve the system of available memory. Please review /proc/meminfo and your current hugepages configuration to see if you can reduce the hugepages allocation. For details, see the kernel documentation at https://docs.kernel.org/admin-guide/mm/hugetlbpage.html.
```

wilkmar: done

dosaboy:
lgtm, thanks marcin!

Issue #627, opened 2023-05-31
EOL check for kernel version


nishant-dash:
Hello,

I see EOL checks for openstack, but I was wondering if we could have an EOL check for the kernel version.

For example, when using the `--kernel` plugin, currently it is something along the lines of
```console
kernel:
  version: 5.13.0-35-generic
  boot: ...
  cpu:
  ...
```

So perhaps something similar to Openstack, such as 

```console
kernel:
  version: 5.13.0-35-generic
  days-to-eol: xyz
  boot: ...
  cpu:
  ...
```
and a note when its past EOL

dosaboy:
EOL info can be found at https://ubuntu.com/about/release-cycle#ubuntu-kernel-release-cycle (you need to zoom in a lot and the graph will turn into actual dates)

PR #808, opened 2024-04-12
Add kernel eof info check

dosaboy:
Resolves: #627

### DIFF
--- a/hotsos/core/plugins/kernel/common.py
+++ b/hotsos/core/plugins/kernel/common.py
@@ -1,10 +1,46 @@
 import os
 import re
 from functools import cached_property
+from datetime import datetime
 
+from hotsos.core.host_helpers.cli import CLIHelper
 from hotsos.core import host_helpers, plugintools
 from hotsos.core.config import HotSOSConfig
+from hotsos.core.log import log
 from hotsos.core.plugins.kernel.config import KernelConfig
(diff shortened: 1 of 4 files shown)


hotsos/core/plugins/kernel/common.py
```diff
@@ -1,10 +1,46 @@
...
 from hotsos.core.config import HotSOSConfig
+from hotsos.core.log import log
 from hotsos.core.plugins.kernel.config import KernelConfig
+from hotsos.core.plugins.system import SystemBase
+
+
+# NOTE(dosaboy): when updating this, refer to the kernel supported versions
+# page: https://ubuntu.com/about/release-cycle#ubuntu-kernel-release-cycle
+KERNEL_EOL_INFO = {
+    '5.19':
```
mustafakemalgilor, line 17: The page includes the kernel EOL info for 6.5, but 5.19 is now absent (IDK if it was ever present).

Also, I found the [distro-info-data](https://salsa.debian.org/debian/distro-info-data/) Debian package which includes [this](https://salsa.debian.org/debian/distro-info-data/-/blob/main/ubuntu.csv?ref_type=heads) CSV file with all the ubuntu release lifecycle information. Still, it does not include the supported kernels for each release --  might be useful for checking the release itself's EOL status.

PR #810, opened 2024-04-12
Only execute conclusions until a match is found

dosaboy:
For a set of conclusions with varying priority, we want to execute them in order of priority and only go as far as the first set of matches of the same priority so as to avoid unecessarily executing conclusions where the result will never be used.


pponnuvel (approved):
LGTM

PR #812, opened 2024-04-15
Add checks for Juju/Pebble CVE-2024-3250

sombrafam:


hotsos/defs/scenarios/juju/juju_pebble_cve.yaml
```diff
@@ -0,0 +1,63 @@
...
+      ops: [ [ pkg_version_gt, "3.4" ], [ pkg_version_le, 3.4.1 ] ]
+
+conclusions:
+  pebble_cve:
+    decision:
+      - has_affected_pebble
+    raises:
+      type: PebbleWarning
+      message: >-
+        The pebble snap or package is affected by a CVE. Please upgrade to a
```
nicolasbock, line 47: You should mention which CVE the package is affected by. And ideally also add a URI to that CVE.

hotsos/core/ycheck/engine/properties/requires/common.py
```diff
@@ -78,6 +78,36 @@ def not_installed(self):
         return set(self.installed).symmetric_difference(_all)
 
 
+class PackageVersionOperations(object):
```
nicolasbock, line 81: Please add tests for this class.

nicolasbock (commented):
Could you add a test for the

hotsos/core/ycheck/engine/properties/requires/common.py
sombrafam: Yeah, I just want to get some feedback to know if this is the right thing.

hotsos/defs/scenarios/juju/juju_pebble_cve.yaml
```diff
@@ -0,0 +1,63 @@
...
+          max: 646
+    apt:
+      pebble:
+        - min: 1.4.0
+          max: 1.4.1
+        - min: 1.7.0
+          max: 1.7.3
+        - min: 1.10.0
+          max: 1.10.1
+  has_2_9_affected_juju:
```
pponnuvel, line 15: Since the CVE actually only affects Pebble (affects Juju only as a dependency), wouldn't it be sufficient to check the Pebble versions (as above)?

sombrafam: juju also needed to be patched. I think pebble is a library dependency.

hotsos/core/ycheck/engine/properties/requires/common.py
dosaboy: The reason you are adding this is because you want to compare versions for a binary that is not installed by a package and so therefore we cannot use the apt or snap helpers. This is very specific to Juju so imho we should implement the logic in the juju plugin itself. Having said that, since these versions are dpkg style versions we should try to leverage the existing logic rather than re-implementing it. Lets try to find a way to re-use the logic in the existing apt requires property to compare these versions even though they didn't come from an apt package. I can have a go at adding support for this in a separate patch to reduce complexity.

hotsos/defs/scenarios/juju/juju_pebble_cve.yaml
dosaboy: Agreed. I was going to suggest doing ```bug-id: ...``` but https://ubuntu.com/security/CVE-2024-3250 has a LP link that goes nowhere since the issue is not in the ubuntu archives. I think it might be time add support for a native cve: key. I will do this seperately.

dosaboy:
Hi erlon, thank you for submitting this patch, we definitely need to get a check in for this cve asap.

You need to make a few adjustments here. Firstly, your patch is based on very old code (last updated Wed Apr 12 12:16:03 2023). You need to update your local repo and rebase your code to get the latest code from upstream (incl. pebble support). I have made a few comments inline and set a couple of action items to followup on.

sombrafam:
I don't see your inline comments. I can see Nickola's and Pon's, but not yours.

dosaboy:
sorry i forgot to submit them, they should be there now

hotsos/core/ycheck/engine/properties/requires/common.py
sombrafam: yeah, I didn't do any version checking. Just used `distutils.version`. We can replace our `packaging.DPKGVersion` for `distutils.version.LooseVersion` which is more generic and I suppose more well-established.

hotsos/defs/scenarios/juju/juju_pebble_cve.yaml
pponnuvel: Right. I checked this a bit. It seems there's no way to check "if this Juju is using pebble x". So I think this is fine. But I think we should take out the pebble bits out of this check and add it as a separate scenario as Pebble could be used independent of Juju for something else (also it needs to cover snap install too along with `apt`).

hotsos/defs/scenarios/juju/juju_pebble_cve.yaml
dosaboy: See https://github.com/canonical/hotsos/pull/814

pponnuvel (approved):
LGTM

Left a [question](https://github.com/canonical/hotsos/pull/812/files#r1585080951) - it's for @dosaboy to answer as well.

hotsos/core/issues/issue_types.py
```diff
@@ -185,6 +185,17 @@ def url(self):
         return "{}{}".format(self.base_url, self.id)
 
 
+class MitreCVE(CVETypeBase):
+
+    @property
+    def base_url(self):
+        return 'https://www.cve.org/CVERecord?id='
```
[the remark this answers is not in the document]
dosaboy: i don't really mind either way and we did discuss it so im ok to go with this for now.

PR #821, opened 2024-04-18
Fix some pylint warnings

brianphaley:
find_most_recent_start() was using None as a default value, changed to use {} as it was confusing pylint, unsubscriptable-object warning.

consider-using-with is now enforced as there are no longer any occurences of code triggering it.


dosaboy:
@brianphaley i attempted to resolve the pylintrc conflict through the github interface but not sure how it will turn out

dosaboy (approved):
lgtm

PR #824, opened 2024-04-19
[rabbitmq] simplify max_node deduction using builtin 'max'

pponnuvel:


hotsos/core/plugins/rabbitmq/report.py
```diff
@@ -72,14 +72,7 @@ def skewed_nodes(self):
...
             if skewed_queue_nodes:
-                max_node = None
-                for node_name in skewed_queue_nodes:
-                    if max_node is None:
-                        max_node = node_name
-                    elif (skewed_queue_nodes[node_name] >=
-                            skewed_queue_nodes[max_node]):
-                        max_node = node_name
-
+                max_node = max(skewed_queue_nodes, key=skewed_queue_nodes.get)
```
mustafakemalgilor, line 75: Just a nitpick: the `max()` will return the first encountered maximum item where there are multiple maximum items (i.e., having the largest value and being equal), but the algorithm above would return the last encountered maximum since the comparison is ">=".

pponnuvel: Good point. But I am not sure if the original logic of "return the last item on multiple maximals" was intentional. 

Unfortunately `>` is [hardcoded](https://github.com/python/cpython/blob/main/Python/bltinmodule.c#L1921). Don't see a way to preserve the existing behaviour with builtin `max`.

As the original author, Ed will probably know whether this difference is critical :)

nicolasbock: You could run the `max()` on a reversed list.

pponnuvel: Thanks @nicolasbock!

This seems to do the trick:
```
max(dict(reversed(list(skewed_queue_nodes.items()))), key=skewed_queue_nodes.get)
```

But this is almost in the [ioccc](https://www.ioccc.org/) territory ;-) So I'll wait for Ed to respond.

dosaboy: starting to feel like maybe not worth changing?

pponnuvel: ack

Issue #818, opened 2024-04-18
Auto-deduced YAML test case names collide when the file name includes more than one dot 


mustafakemalgilor:
Assuming the following YAML file names, the auto-deduced test case names are the same even though the file names are different:

- lp1896125-17.0.0.yaml
- lp1896125-17.0.1.yaml

These files yield `lp1896125-17` as the test name, which leads to a name collision.

Ideally, the code should strip away the file extension, sanitize the remainder, and use that as a test name.

dosaboy:
@mustafakemalgilor while we should probably fix that, I feel it makes sense to give the files meaningful names so that you can tell what they are doing from their name so e.g. suffix like _pass.yaml _fail.yaml is what i usually use.

mustafakemalgilor:
Sure, in this context the suffixes are version numbers for different keystone releases I'd like to test the scenario against. I wanted to be verbose on test cases for ensuring that the new version comparison logic works as intended for non-superficial use cases.

PR #825, opened 2024-04-19
utils/TemplatedTestGenerator/test_method_name: support dots in name

mustafakemalgilor:
test_method_name now properly removes the file extension from the file name and uses the remainder even if there are multiple dots present in the name.

Fixes: #818

### DIFF
--- a/tests/unit/utils.py
+++ b/tests/unit/utils.py
@@ -229,8 +229,11 @@ class TemplatedTestGenerator(object):
     @property
     def test_method_name(self):
         """ Test method name uses the original name. """
-        name = self.test_sub_path.split('.')[0]
-        name = name.replace('/', '_')
+        name = os.path.splitext(self.test_sub_path)[0]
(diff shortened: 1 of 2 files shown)


tests/unit/utils.py
```diff
@@ -229,7 +229,7 @@ def target_path(self):
     @property
     def test_method_name(self):
         """ Test method name uses the original name. """
-        name = self.test_sub_path.split('.')[0]
+        name = os.path.splitext(self.test_sub_path)[0]
```
dosaboy, line 232: i believe we need to also do name.replace('.', '_') since attributes cannot have a '.' in there name.

dosaboy: hmm actually because we are since we are using getattr/setattr it is allowed but still I feel that test names should not have a '.' in them.

mustafakemalgilor: Ack, will do.

PR #826, opened 2024-04-19
utils/TemplatedTest: raise exception when target scenario is not found

### LINKED ISSUE #819: The YAML test files do not raise error when the corresponding scenario file is absent
When the test with the name ".../tests/scenarios/openstack/keystone/foo.yaml" exists, but the scenario ".../scenarios/openstack/keystone/foo.yaml" is absent, the test proceeds to evaluate the conditions anyway. This is the same with the explicitly specified target-name property as well:

lp1896125-17-0-1.yaml
```yaml
target-name: lp1896125-no-such-file.yaml
data-root:
  files:
    sos_commands/dpkg/dpkg_-l: |
      ii  keystone 2:17.0.1-0ubuntu0.20.04.1 amd64
    var/log/keystone/keystone.log: |
          File "/usr/lib/python3.10/contextlib.py", line 135, in __enter__
            return next(self.gen)
          File "/usr/lib/python3/dist-packages/ldappool/__init__.py", line 391, in connection
            raise MaxConnectionReachedError(self.uri)
      ldappool.MaxConnectionReachedError: ldaps://ldapserver.example.com
raised-bugs:
  # none
```

Test results:

```terminal
scenario/lp1896125-rework 16:23 > tox -e py3 -- lp1896125-17-0-1

{0} tests.unit.test_openstack.TestOpenstackScenarios.test_keystone_bugs_lp1896125-17-0-1 [0.005862s] ... ok

======
Totals
======
Ran: 1 tests in 0.0059 sec.
 - Passed: 1
 - Skipped: 0
 - Expected Fail: 0
 - Unexpected Success: 0
 - Failed: 0
Sum of execute time for each test: 0.0059 sec.

==============
Worker Balance
==============
 - Worker 0 (1 tests) => 0:00:00.005862
  py3: OK (0.85=setup[0.02]+cmd[0.83] seconds)
  congratulations :) (0.88 seconds)
  
  ```

### PR DESCRIPTION
**Pull request by mustafakemalgilor**
the yaml scenario tests were silently proceeding when a corresponding scenario file was not present. now, it raises a FileNotFoundError when the scenario is absent.

Fixes #819

### DIFF
--- a/tests/unit/test_ut_scenario_loader.py
+++ b/tests/unit/test_ut_scenario_loader.py
@@ -160,11 +160,19 @@ class TestScenarioTestLoader(ScenarioTestsBase):
         with tempfile.TemporaryDirectory() as dtmp:
             self.create_scenarios(dtmp, levels=3)
             self.create_tests(dtmp, levels=3)
+
+            with open(os.path.join(dtmp, 'test1.yaml'),
+                      'w') as fd:
+                fd.write(FAKE_SCENARIO)
             with mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp):
-                utils.TemplatedTest('test1.yaml', {}, [], [], [], '/')()(self)
+                utils.TemplatedTest(
+                    os.path.join(dtmp, 'test1.yaml'),
+                    {}, [], [], [], '/')()(self)
 
     def test_scenarios(self):
-        with tempfile.TemporaryDirectory() as dtmp:
+        with tempfile.TemporaryDirectory() as dtmp,\
+             mock.patch.object(utils, 'DEFS_DIR', dtmp),\
+             mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp + '/tests'):
             self.create_scenarios(dtmp, levels=5)
             self.create_tests(dtmp, levels=5)
             orig_config = HotSOSConfig.CONFIG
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/unit/test_ut_scenario_loader.py [hunk 1]
```diff
@@ -160,70 +160,94 @@ def test_is_def_filter(self):
...
-                utils.TemplatedTest('test1.yaml', {}, [], [], [], '/')()(self)
+                utils.TemplatedTest(
+                    os.path.join(dtmp, 'test1.yaml'),
+                    {}, [], [], [], '/')()(self)
 
     def test_scenarios(self):
-        with tempfile.TemporaryDirectory() as dtmp:
+        with tempfile.TemporaryDirectory() as dtmp,\
+             mock.patch.object(utils, 'DEFS_DIR', dtmp),\
+             mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp + '/tests'):
```
dosaboy, line 175: Any chance we could use parenthesis instead of slash, it just hurts my eyes a little less:

```
        with (tempfile.TemporaryDirectory() as dtmp,
              mock.patch.object(utils, 'DEFS_DIR', dtmp),
              mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp + '/tests')):
```

tests/unit/test_ut_scenario_loader.py
```diff
@@ -160,70 +160,94 @@ def test_is_def_filter(self):
...
+
+                self.assertTrue(raised)
             finally:
                 HotSOSConfig.set(**orig_config)
 
     def test_scenarios_deep(self):
-        with tempfile.TemporaryDirectory() as dtmp:
+        with tempfile.TemporaryDirectory() as dtmp,\
+             mock.patch.object(utils, 'DEFS_DIR', dtmp),\
+             mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp + '/tests'):
```
dosaboy, line 212: same here for parenthesis vs. slash

tests/unit/test_ut_scenario_loader.py
```diff
@@ -160,70 +160,94 @@ def test_is_def_filter(self):
...
 
-                    self.assertTrue(raised)
+                self.assertTrue(raised)
             finally:
                 HotSOSConfig.set(**orig_config)
+
+    def test_scenarios_no_such_target(self):
+        with tempfile.TemporaryDirectory() as dtmp,\
+             mock.patch.object(utils, 'DEFS_DIR', dtmp),\
+             mock.patch.object(utils, 'DEFS_TESTS_DIR', dtmp + '/tests'):
```
dosaboy, line 242: and again same here for parenthesis vs. slash

tests/unit/test_ut_scenario_loader.py
```diff
@@ -160,70 +160,94 @@ def test_is_def_filter(self):
...
+            self.create_tests(dtmp, levels=5)
+            # Do not create scenarios intentionally
+            try:
```
dosaboy, line 245: You can use ```with self.assertRaises(FileNotFoundError):``` and I also realise I should have done this in the above tests.

tests/unit/utils.py
```diff
@@ -75,6 +76,10 @@ def __init__(self, target_path, data_root, mocks, expected_bugs,
                  expected_issues, sub_root):
         self.sub_root = sub_root
         self.target_path = target_path
```
dosaboy, line 78: self.target path should be pointing to the scenario so in fact you should be able to check that rather than create a new "scenario_path"

tests/unit/test_ut_scenario_loader.py
mustafakemalgilor: Of course, that looks better to my eyes too.

tests/unit/test_ut_scenario_loader.py
mustafakemalgilor: Will do

tests/unit/utils.py
mustafakemalgilor: The code needs an absolute path to the scenario, so I thought that having an intermediate variable would be a good idea. Should the "target_path" also be set to the absolute path of the scenario?

dosaboy: no, it is the target as in https://hotsos.readthedocs.io/en/latest/contrib/testing.html#test-reference so it is a path but relative to a scenario (most of the time it is a filename). But ```os.path.join(self.sub_root, target_path)``` gives you a path relative to ```defs```  e.g. with *target_path* set to "unattended_upgrades.yaml" you would get ```scenarios/system/unattended_upgrades.yaml``` which is the scenario that is expected to exist.

mustafakemalgilor: So I tried `if not os.path.exists(os.path.join(self.sub_root, self.target_path)):` but it's failing for some test cases, so I still need to use the `DEFS_DIR` (i.e. `if not os.path.exists(os.path.join(DEFS_DIR, self.sub_root, self.target_path)):`) in order to make it work. Am I missing something here?

tests/unit/test_ut_scenario_loader.py
mustafakemalgilor: @dosaboy it seems like "with" with parantheses introduced in python 3.10 so the py3.8 job fails in CI.

tests/unit/utils.py
dosaboy: it think the way it is currently is good. scenario path is now ```os.path.exists(os.path.join(DEFS_DIR, self.sub_root, self.target_path))``` which is correct.

Issue #815, opened 2024-04-16
Add custom binary version helper


dosaboy:
We have requirement types for apt, snap, pebble etc but some application binaries are shipped/installed without a package and we currently have no similar way to check their version as if they were a package. One example of such a binary is Juju. We could add e.g. YRequirementTypeGenericBin that takes as an argument an interface that can provide versions etc and use that to apply the version range check like we do for packages.

PR #828, opened 2024-04-22
Add generic binary requirements checker

dosaboy:
We have requirement types for packages like apt, snap etc but do not have a way to get information from a binary. Some applications are not installed by packages but may still have e.g. a version so having a way to query this is useful. This tries to provide a common way to do do this and uses the Juju plugin/binary as an initial use-case.

Resolves: #815

### DIFF
--- a/doc/source/contrib/language_ref/property_ref/requirement_types.rst
+++ b/doc/source/contrib/language_ref/property_ref/requirement_types.rst
@@ -98,6 +98,39 @@ Cache keys:
   * package - name of each installed package
   * version - version of each installed package
 
+Binary
+---
+
+Takes a binary name and optional list of version ranges. Returns True if
+the binary exists and if provided, version is within ranges.
+
+Usage:
+
+.. code-block:: yaml
+
+    binary:
+      name: mybin
+      handler: path.to.handler
+
+Or with version ranges as follows:
+
+.. code-block:: yaml
+
+    binary:
+      handler: path.to.bininterfacehandler
+      mybin:
+        - min: 0.0
+          max: 1.0
+        - min: 4.0
+          max: 5.0
+
+NOTE: the version checking logic is currently the same as for the :ref:`Apt` type.
+
+Cache keys:
+
+  * binary - name of each installed binary
+  * version - version of each installed binary
+
 Snap
 ----
 
--- a/hotsos/core/plugins/juju/resources.py
+++ b/hotsos/core/plugins/juju/resources.py
@@ -1,6 +1,7 @@
 import glob
 import os
 import re
+import subprocess
 from functools import cached_property
 
 import yaml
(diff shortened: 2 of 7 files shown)


hotsos/core/ycheck/engine/properties/requires/types/binary.py
```diff
@@ -0,0 +1,59 @@
...
+
+        _result = True
+        # bail on first fail i.e. if any not installed
+        if not items.not_installed:
+            for binary, versions in items:
+                log.debug("binary %s installed=%s", binary, _result)
+                if not versions:
+                    continue
+
+                _result = items.package_version_within_ranges(binary, versions)
```
brianphaley, line 43: I guess this will do the right thing and you don't need a binary_version_within_ranges() type function?

brianphaley (commented):
Realized this is just a draft change

hotsos/core/ycheck/engine/properties/requires/types/binary.py
dosaboy: Correct, as long as the versions follow apt/dpkg logic it works. If this doesnt work at some later point we can make it overridable but for now we can re-use the apt logic.

doc/source/contrib/language_ref/property_ref/requirement_types.rst
```diff
@@ -98,6 +98,39 @@ Cache keys:
...
+
+Takes a binary name and optional list of version ranges. Returns True if
+the binary exists and if provided, version is within ranges.
+
+Usage:
+
+.. code-block:: yaml
+
+    binary:
+      name: mybin
```
dosaboy, line 112: TODO: need to implement this in the code

hotsos/core/plugins/juju/resources.py
```diff
@@ -204,3 +204,16 @@ def charm_names(self):
             return []
 
         return list(self.charms.keys())
+
+
+class JujuBinaryInterface(JujuBase):
```
pponnuvel, line 209: Do we need to have such a BinaryInterface class for each snap e.g. `kubectl`, `lxd`, etc?

dosaboy: No because snaps are covered by YRequirementTypeSnap

pponnuvel: But we do for `juju` only because older ones are apt pkg?

dosaboy: juju is installed in different ways depending on the host, the juju client is typically a snap but the juju controller is not installed by a (apt/snap) package so we have to query the installed binary directly instead. Here is an example from a juju controller:

```
root@juju-da71e1-controller-0:/home/ubuntu# dpkg -l| grep juju
root@juju-da71e1-controller-0:/home/ubuntu# snap list| grep juju
juju-db  4.4.18         160    4.4/stable     juju-qa      -
root@juju-da71e1-controller-0:/home/ubuntu# dpkg -S /etc/systemd/system/jujud-machine-0.service
dpkg-query: no path found matching pattern /etc/systemd/system/jujud-machine-0.service
root@juju-da71e1-controller-0:/home/ubuntu# dpkg -S /var/lib/juju/tools/machine-0/jujud
dpkg-query: no path found matching pattern /var/lib/juju/tools/machine-0/jujud
root@juju-da71e1-controller-0:/home/ubuntu# /var/lib/juju/tools/machine-0/jujud --version
2.9.49-ubuntu-amd64
```

dosaboy: to be fair, in this patch the JujuBinaryInterface.get_version() is calling JujuMachine.version which gets the version from the config file. I think I will extend JujuMachine to first actually query the binary then go to config as a fallback.

pponnuvel: Thanks for the details.

pponnuvel (approved):
LGTM

PR #829, opened 2024-04-23
[unittest] Define addDuration method to support 3.12

pponnuvel:
Change:
https://github.com/python/cpython/pull/12271/commits/d88087ea57f116d754b92c5189d75777f717f4ef

This essentially suppresses the warnings:
RuntimeWarning: TestResult has no addDuration method\n  warnings.warn("TestResult has no addDuration method")

### DIFF
--- a/tests/unit/utils.py
+++ b/tests/unit/utils.py
@@ -450,6 +450,13 @@ class BaseTestCase(unittest.TestCase):
         else:
             LoggingManager().start(level=logging.WARNING)
 
+    def _addDuration(self, *args, **kwargs):  # For Python >= 3.12
+        """ Python 3.12 needs subclasses of unittest.TestCase to implement
+        this in order to record times and execute any cleanup actions once
+        a test completes regardless of success. Otherwise it emits a warning.
+        This generic implementation helps suppress it and possibly others in
+        the future. """
+
     def tearDown(self):
         LoggingManager().stop()
         HotSOSConfig.reset()


tests/unit/utils.py
```diff
@@ -441,6 +441,9 @@ def setUp(self):
         else:
             LoggingManager().start(level=logging.WARNING)
 
+    def _addDuration(self, result, elapsed):  # For Python >= 3.12
```
dosaboy, line 444: is overriding this method not just masking a real issue i.e. that addDuration() does not exist?

pponnuvel: It's a new function in 3.12 that subclasses are expected to implement:
https://docs.python.org/3/library/unittest.html#unittest.TestResult.addDuration

It seems for tracking testcase execution times and for running any cleanups. I thought we don't care about either of these right now.

mustafakemalgilor: I believe it would be better to have a docstring stating that instead of having "pass" as the function body for future travelers.

pponnuvel: ack, done

pponnuvel:
> @pponnuvel the CI failure is unrelated to this MP, please rebase your MP after #834 is merged.

Thanks, @mustafakemalgilor I was getting antsy on how adding a comment could trigger failures :)

mustafakemalgilor:
> > @pponnuvel the CI failure is unrelated to this MP, please rebase your MP after #834 is merged.
> 
> Thanks, @mustafakemalgilor I was getting antsy on how adding a comment could trigger failures :)

The CI works in mysterious ways ;)

mustafakemalgilor:
LGTM :+1:

tests/unit/utils.py
```diff
@@ -450,6 +450,12 @@ def setUp(self):
...
+        """ Python 3.12 needs subclasses of unittest.TestCase to implementxi
```
dosaboy, line 454: typo: s/implementxi/implement/

tests/unit/utils.py
dosaboy: @pponnuvel shouldn't the method be called addDuration()?

dosaboy: Ok I now see my confusion, addDuration is for TestResult whereas \_addDuration is for TestCase. For what its worth the commit message is a bit misleading since it is missing the "_".

dosaboy: In order to avoid any forward compatibility issues, we could also define a more generic method i.e.:

```
def _addDuration(self, *args, **kwargs):  # For Python >= 3.12
...
```

dosaboy (approved):
lgtm thanks pon!

PR #831, opened 2024-04-23
Bump `pylint` dependency

nicolasbock:
Closes: https://github.com/canonical/hotsos/issues/830

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -50,4 +50,5 @@ disable=
  unspecified-encoding,
  consider-using-f-string,
  unused-private-member,
- no-member
+ no-member,
+ duplicate-code,
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -2,5 +2,5 @@ bashate
 flake8==6.0.0
 flake8-import-order==0.18.2
 stestr
-pylint==2.17.4
+pylint==3.0.4
 rpdb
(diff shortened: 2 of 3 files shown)


test-requirements.txt
```diff
@@ -2,5 +2,5 @@ bashate
 flake8==6.0.0
 flake8-import-order==0.18.2
 stestr
-pylint==2.17.4
+pylint==3.0.0
```
pponnuvel, line 5: The default on Noble I see is `3.0.3`. Is there a reason not to use that?

nicolasbock: No, I figured I'd start very gently :smile:

pponnuvel: Ah, I didn't realise this is a draft!

nicolasbock: I am getting lots of errors at this point and I am trying to figure out how to fix those.

pylintrc
```diff
@@ -19,35 +19,36 @@ score=yes
...
  too-many-branches,
  too-many-instance-attributes,
- too-many-ancestors,
- too-many-public-methods,
  too-many-lines,
+ too-many-locals,
  too-many-nested-blocks,
+ too-many-public-methods,
  too-many-statements,
- protected-access,
```
dosaboy, line 38: this reordering makes it hard to see what has been enabled and disabled. is there anyway we could maintain the original order so that diff only shows what has changed?

test-requirements.txt
dosaboy: if you want to do it it'd be fine to do as a separate patch that makes it clear in the commit that no real changes are being made.

pylintrc
nicolasbock: Ok, will do that

PR #833, opened 2024-04-24
Add `yamllint` to the project

mustafakemalgilor:
since the project is using *many* YAML files, it would be beneficial to use a linter both in tox testing and CI. This patch adds [yamllint](https://github.com/adrienverge/yamllint) 1.35.1 to the project as a test dependency. this patch also adds a new tox target named `yamllint` which runs yamllint at the which runs yamllint at the hotsos/ and tests/ directories.
 
 the yamllint currently uses the `yamllintrc` file supplied in this patch. this configuration inherits from the default configuration of the yamllint, and excludes only one path, i.e.: `fake_data_root`, because it contains test data captured from some environment and should be kept as-is. For further reference regarding available options, please see [the docs](https://yamllint.readthedocs.io/).
 
  
summary of changes made to yaml files:
   - changed all truth values (True, False) to (true, false)
   - added linter suppression comments for false positives
   - removed all trailing spaces, except for test file contents
   - removed all redundant empty lines from EOF
   - fixed indentation issues reported by the linter
    
PS: please don't be scared of the gargantuan list of YAML file changes, these are all just cosmetic changes to make the linter happy.


mustafakemalgilor:
Removing the CI step for now, because it completely removes all tasks from the "tests.yaml" file since the file has been changed. This raises another question: "Should we separate ".yaml" files so newly added CI steps do not interfere with the execution of the existing ones?"

nicolasbock (commented):
LGTM. Good idea :smile:

nicolasbock (approved):

yamllintrc
```diff
@@ -0,0 +1,13 @@
+extends: default
+
+ignore:
+  - '**/hotsos/tests/unit/fake_data_root/**'
+  - '**/.tox/**'
+  - '**/examples/**'
+  - '**/.github/**'
+
+rules:
+  # 80 chars should be enough, but don't fail if a line is longer
```
mustafakemalgilor, line 10: this should be `120` chars instead of `80`

dosaboy: So i was thinking that maybe we should ignore this one completely and my reasoning is that having to wrap the search expressions over multiple lines makes it really easy to inadvertently insert an unexpectced whitespace char which could break the expr.

mustafakemalgilor: You have a valid point, but I would argue that we should keep the line limit. IMO, if an expression is longer than 120 characters, it should have a valid justification and the warning should be suppressed only if expressing the check with a shorter expression is not possible. This would encourage refactoring the expressions. Lucky for us, the number of long expressions is rather small so I can go ahead and add the suppressions without having to comment on every single "expr" with the exception.

mustafakemalgilor: Updated the MP: now long expressions are kept as-is and the warning is suppressed individually.

mustafakemalgilor (commented):
I missed reverting `expr` changes on these:

- src_migration.yaml
- lp1948906.yaml
- lp1996230.yaml
- lp2029857.yaml

And there are some unintended changes to tests/unit/fake_data_root which are remnants of before "ignore" path.

yamllintrc
```diff
@@ -0,0 +1,13 @@
...
+  line-length:
+    max: 120
+    level: warning
```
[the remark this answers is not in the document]
mustafakemalgilor: done

mustafakemalgilor:
> I missed reverting `expr` changes on these:
> 
> * src_migration.yaml
> * lp1948906.yaml
> * lp1996230.yaml
> * lp2029857.yaml
> 
> And there are some unintended changes to tests/unit/fake_data_root which are remnants of before "ignore" path.

Fixed.

yamllintrc
```diff
@@ -0,0 +1,29 @@
+extends: default
+
+ignore:
+  - '**/hotsos/tests/unit/fake_data_root/**'
+  - '**/.tox/**'
+  - '**/examples/**'
+  - '**/.github/**'
+
+rules:
+  # https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.line_length
```
dosaboy, line 10: should be https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.document_start

yamllintrc
```diff
@@ -0,0 +1,29 @@
...
+  line-length: disable
+  # https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.anchors
+  anchors:
+    forbid-undeclared-aliases: true
+    forbid-duplicated-anchors: true
+    forbid-unused-anchors: true
+  # https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy
+  truthy:
+    allowed-values: ['true', 'false', 'on']
```
dosaboy, line 22: on?

mustafakemalgilor: I'm going to remove ` ignore: - '**/.github/**'` on the CI enablement patch, and 'on' is a keyword in GitHub workflow syntax so I'm excluding that.

yamllintrc
mustafakemalgilor: Updated again, removed the line length check altogether.

dosaboy (approved):
lgtm, thank you for adding this.

PR #834, opened 2024-04-24
utils/TemplatedTest: fix scenario related test failures

mustafakemalgilor:
the create_scenarios function got squashed somehow so the main branch still got the old version, hence causing test/CI failures after the `raise exception when target scenario is not found` patch got merged to main.


mustafakemalgilor:
@dosaboy ready to go.

dosaboy:
i think this may have been caused by the patch in https://github.com/canonical/hotsos/issues/818 landing first but not sure

PR #835, opened 2024-04-25
Move lp1980211 scenario to Openstack and extend

dosaboy:
Moves the scenario for LP 1980211 to the Openstack plugin set since it requires Neutron agents. Also extends the check and tests to cover the case where the package contains the fix but we still see the error.


hotsos/defs/scenarios/openstack/neutron/bugs/lp1980211.yaml
```diff
@@ -0,0 +1,61 @@
...
+        version: '@checks.has_pyovs_with_reconnection_bug.requires.version'
+  neutron_server_error_and_pkg_w_fix:
+    decision:
+      - not: has_pyovs_with_reconnection_bug
+      - is_neutron_server
+      - has_ovsdb_reconnection_errors
+    raises:
+      type: OpenstackWarning
+      message: >-
+        This is host is running neutron-server and appears to be suffering from
```
brianphaley, line 41: s/This is host is/This host is

Fix below as well

brianphaley (changes requested):
Might as well fix the typo

hotsos/defs/scenarios/openstack/neutron/bugs/lp1980211.yaml
dosaboy: roger

brianphaley (approved):

PR #837, opened 2024-04-26
gh.actions: add yamllint to ci workflow file

mustafakemalgilor:


mustafakemalgilor:
Interestingly the pipeline did run this time. I believe it was because I formatted the tests.yaml file in the previous PR which prevented the pipeline from being run, or it could be a one-time fluke.

PR #838, opened 2024-04-26
Fix openstack/neutron/bugs/lp1980211.yaml

dosaboy:
This check applies to both neutron-server and
neutron-ovn-metadata-agent but the logs are
different for both and the regex pattern did
not take this into account.


mustafakemalgilor (approved):
LGTM

PR #851, opened 2024-04-30
Release snap to stable channel

dosaboy:
Updates gh workflow to release snap build to
all channels.

### DIFF
--- a/.github/workflows/snapbuild.yaml
+++ b/.github/workflows/snapbuild.yaml
@@ -20,4 +20,4 @@ jobs:
         SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
       with:
         snap: ${{ steps.build.outputs.snap }}
-        release: edge
+        release: stable


nicolasbock (commented):
Why? All channels will have the same version.

dosaboy:
> Why? All channels will have the same version.

We have 4 channels, this is just to get the same version in all channels. If I could delete 3 and keep stable i would but I cant do that so this is the alternative for now.

nicolasbock:
Couldn't you simply publish to `stable` instead of the other channels?

mustafakemalgilor:
> > Why? All channels will have the same version.
> 
> We have 4 channels, this is just to get the same version in all channels. If I could delete 3 and keep stable i would but I cant do that so this is the alternative for now.

The snapcraft [documentation] says that if a channel is closed, then the updates will be received from the lower risk level channel:

> Closing channels
> A channel can be closed by a snap publisher when there is no longer a snap that fits a channel’s original purpose or specification.
> 
> For example, when a specific risk-level channel is closed, the snap store will select a snap from a more stable risk-level of the same track. If the original channel is re-opened, snaps will once again be selected from the original channel.
> 
> This approach is commonly used for beta testing. If a snap is following a beta channel that is then closed, the store will offer the snap from the candidate channel. If the candidate channel is not available, the snap from the stable channel will be selected. If the beta channel re-opens, the snap will once again be selected from that channel.

So according to this documentation, if we were to close all but "stable", everybody would get their updates from it.  

[documentation]: https://snapcraft.io/docs/channels

dosaboy:
> > > Why? All channels will have the same version.
> > 
> > 
> > We have 4 channels, this is just to get the same version in all channels. If I could delete 3 and keep stable i would but I cant do that so this is the alternative for now.
> 
> @dosaboy
> 
> The snapcraft [documentation](https://snapcraft.io/docs/channels) says that if a channel is closed, then the updates will be received from the lower risk level channel:
> 
> > Closing channels
> > A channel can be closed by a snap publisher when there is no longer a snap that fits a channel’s original purpose or specification.
> > For example, when a specific risk-level channel is closed, the snap store will select a snap from a more stable risk-level of the same track. If the original channel is re-opened, snaps will once again be selected from the original channel.
> > This approach is commonly used for beta testing. If a snap is following a beta channel that is then closed, the store will offer the snap from the candidate channel. If the candidate channel is not available, the snap from the stable channel will be selected. If the beta channel re-opens, the snap will once again be selected from that channel.
> 
> So according to this documentation, if we were to close all but "stable", everybody would get their updates from it.

i will look into closing them, if i can i will

nicolasbock:
You could close the other channels and only publish to stable.

dosaboy:
ok ive closed edge, beta and candidate and will update this patch to push to stable.

PR #852, opened 2024-04-30
ycheck/requires/types/apt: decouple version range utilities from APT

mustafakemalgilor:
The version range functions 'normalize_version_criteria' and 'is_version_within_ranges' are not exclusive to apt and could be used by other classes. This patch moves these two functions to host_helpers/packaging/DPKGVersion, along with the unit tests. Added new unit tests for requires/types/apt. Reflected the changes to requires/types/binary and its unit tests.

Added __repr__ class member fn to DPKGVersion to get a readable value when `print()` ing the DPKGVersion objects.

`package_version_within_ranges` is now called `is_version_with in_ranges` and it accepts a version instead of a package name.

BinCheckItems no longer inherits from APTCheckItems.

### DIFF
--- a/hotsos/core/host_helpers/packaging.py
+++ b/hotsos/core/host_helpers/packaging.py
@@ -32,6 +32,9 @@ class DPKGVersion(object):
 
         return True
 
+    def __repr__(self):
+        return str(self)
+
     def __str__(self):
         return self.a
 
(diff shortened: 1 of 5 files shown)


dosaboy:
@mustafakemalgilor thanks for this. Before i get fully into the review, because the diff makes it quite hard to see what changed vs. what was moved can you tell me to what extent this is a list and shift of the code and how different the new normalize_version_criteria() is from the old one?

mustafakemalgilor:
> @mustafakemalgilor thanks for this. Before i get fully into the review, because the diff makes it quite hard to see what changed vs. what was moved can you tell me to what extent this is a list and shift of the code and how different the new normalize_version_criteria() is from the old one?

Sure.

`normalize_version_criteria` is a verbatim copy, except it's now a static function instead of a class member one.

`package_version_within_ranges`:
-  is renamed to `is_version_within_ranges`
-  is now a static function
-  replaced `pkg` (the name of the package) with `version`
-  [no longer calls packaging_helper.get_version()](https://github.com/canonical/hotsos/pull/852/files#diff-0fd0e2ec19e26950fbd7d62de94619bdfaa5f3bc3e3b40e574c7a21416992178L136) as it now gets the version directly as a parameter
- the rest is verbatim copy

mustafakemalgilor:
> @mustafakemalgilor thanks for this. Before i get fully into the review, because the diff makes it quite hard to see what changed vs. what was moved can you tell me to what extent this is a lift and shift of the code and how different the new normalize_version_criteria() is from the old one?

@dosaboy  btw you can checkout this branch on your local and use ` git diff --color-moved HEAD~1` to get a diff that shows the changed/deleted/moved parts in a different color.

dosaboy:
@mustafakemalgilor lgtm but can you please rebase because I think this might fail flake8 with changes that are now on master.

mustafakemalgilor:
> @mustafakemalgilor lgtm but can you please rebase because I think this might fail flake8 with changes that are now on master.

ofc, will do.

PR #854, opened 2024-05-03
hotsos/cli: replace deprecated importlib.resources.path with files()

mustafakemalgilor:
the said function is deprecated in python 3.12 and will be removed. the new equivalent function is files(). replaced resources.path() in hotsos/cli/get_repo_info with files() and also updated the unit test.


mustafakemalgilor:
files() is not included in 3.8 python API, so dropping this for now.

Issue #850, opened 2024-04-29
hotsos on arm64 nodes is taking too long in some cases


nishant-dash:
In investigating an issue with stale ports in ovn, I found that one some arm64 nodes, ovn has a lot (anywhere between 30-800) stale port entries on a host with no associated tap devices. This requires manual cleanup using ovs-vsctl like so `ovs-vsctl del-port br-int ...`

When trying to run hotsos on this affected node (with stale ports), hotsos takes a really long time and had to be killed.

On a non-affected node (that has 2 ports in ovn however), hotsos runs in a few seconds.

The hotsos command I am using is
```console
hotsos --openvswitch --kernel --openstack --system
```

hotsos version: 1.17.0+1276~ubuntu22.04.1

mustafakemalgilor:
Hi @nishant-dash 

Is it the same when you grab a sosreport from that node and feed it to the hotsos? If so, an example sosreport would be useful to dig into this issue.

dosaboy:
I think I have some ideas as to what is contributing to the slowness and I'm working on a solution atm. Will have something up soon.

dosaboy:
I believe we should have a substantial improvement with some patches landed recently including the one above and also https://github.com/canonical/hotsos/commit/00452081be6340786c1b55285ccc4551d131348b and some recent changes to [search](https://github.com/dosaboy/searchkit). I have run some tests against small and large sosreports and am definitely seeing improvements. There is definitely a lot of room for improvement still, particularly around the speed of search execution and I now we have further changes queued up for that but these recent patch hopefully remove some slowness reduce memory consumption.

dosaboy:
We now perform all event and scenario searches (per-plugin) in one go and also allow any other searches to be performed in this way. This will remove any duplicate searches or searching of long files more than once which should result in a net improvement.

dosaboy:
closing this out since it does not appear to be an issue any more.

PR #856, opened 2024-05-07
Group all event searches

dosaboy:
To avoid searching the same file more than once we now gather all searches from all events and executed all searches prior to executing event callbacks so that the results are shared with all and avoiding any repeated searches of the same file.

Also removes extraneous regex from some search patterns.

Related-To: #850


hotsos/core/ycheck/events.py
```diff
@@ -225,7 +239,238 @@ def __call__(self):
...
+                event = None
+                # skip plugin name at start and skip ".search" at the end.
+                for node in event_search_prop_path.split('.')[1:-1]:
+                    if branch is None:
+                        branch = getattr(events, node)
+                    else:
+                        branch = getattr(branch, node)
+
+                    # the last one will be the one we want
+                    event = branch
```
mustafakemalgilor, line 419: This could be de-indented out of the loop as the `branch` will still point to the last element after the loop.

hotsos/core/ycheck/events.py
```diff
@@ -225,7 +239,238 @@ def __call__(self):
...
+        events = YDefsSection(HotSOSConfig.plugin_name,
+                              cls.get_defs(group) or {})
+        for prop in events.manager.properties.values():
+            for item in prop:
+                if not issubclass(item['cls'], search.YPropertySearch):
+                    break
+
+                search_props.add(item['path'])
+
+        if len(search_props) > 0:
```
mustafakemalgilor, line 406: This condition could be reversed in order to have less indentation further down, i.e.:

```python

if not search_props:
    log.debug("no events found - no searches to load")
    return
```

hotsos/core/ycheck/events.py
```diff
@@ -235,123 +480,115 @@ class EventHandlerBase(YHandlerBase, EventProcessingUtils):
...
-        for section_name, section in self.event_definitions.items():
-            for event, event_meta in section.items():
-                search_tag = event_meta['tag']
+        for section_name, section in self.events.items():
+            for event, fullname in section.items():
+                event_search = self.search_registry[fullname]['search']
+                search_tag = event_search.unique_search_tag
                 seq_def = None
-                if event_meta['passthrough']:
+                if bool(event_search.passthrough_results):
```
mustafakemalgilor, line 586: I think instead of casting this to `bool` here we should ensure that the property always returns a bool, i.e.:

```python
    @property
    def passthrough_results(self):
        if not isinstance(self.content, dict):
            return False

        return bool(self.content.get('passthrough-results', False))
```

hotsos/core/ycheck/events.py
```diff
@@ -385,6 +623,10 @@ def run(self, results):
                 if not ret:
                     continue
 
+                log.debug("event %s.%s callback '%s' return size is %d bytes",
+                          event_results_obj.section, event, callback_name,
+                          sys.getsizeof(ret))
```
mustafakemalgilor, line 628: `sys.getsizeof()` could be misleading for nested lists/dicts/etc. as it does not follow references.

```python
Python 3.12.3 (main, Apr 10 2024, 05:33:47) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> class b:
...     def __init__(self):
...             self.dict = {'a':5}
... 
>>> obj = b()
>>> sys.getsizeof(obj)
48
>>> class c:
...     def __init__(self):
...             pass
... 
>>> obj2 = c()
>>> sys.getsizeof(obj)
48
>>> sys.getsizeof(obj2)
48
>>> obj.dict = {'a':5, 'b':10, 'c':12}
>>> sys.getsizeof(obj)
48
>>> sys.getsizeof(obj.dict)
184

hotsos/core/ycheck/events.py
dosaboy: fair point, could event get rid of branch and just use event. ill make that change

hotsos/core/ycheck/events.py
dosaboy: ack

hotsos/core/ycheck/events.py
dosaboy: yeah it is in act not giving me the information i was hoping

hotsos/core/ycheck/events.py
dosaboy: agreed

hotsos/core/ycheck/events.py
dosaboy: i have gone and removed this now

mustafakemalgilor (approved):
LGTM

PR #858, opened 2024-05-08
YDefsLoader: add variable interpolation support

mustafakemalgilor:
the default YAML format does not support variable interpolation e.g. referring to another key's value, but it's possible to add it with the help of 'add_constructor` and `add_implicit_resolver`.

Example usage:

```yaml
c: cool
x:
  y: abcdefg
  z: hijklmno
  t: ${c}

foo: ${x.y}${x.z} ${x.t}
# foo's value would be `abcdefghijklmno cool`
```

### DIFF
--- a/hotsos/core/ycheck/engine/common.py
+++ b/hotsos/core/ycheck/engine/common.py
@@ -1,9 +1,110 @@
 import abc
 import os
+import re
 
 import yaml
 from hotsos.core.config import HotSOSConfig
 from hotsos.core.log import log
+
+
+class YRefKeyNotFoundException(Exception):
+    def __init__(self, key):
+        message = f"{key} could not be found."
+        super().__init__(message)
+
+
+class YRefReachedMaxAttemptsException(Exception):
+    def __init__(self, key):
+        message = f"Max search attempts have been reached for {key}"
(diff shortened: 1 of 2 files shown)


dosaboy:
This is interesting. one bit of initial feedback i'd have is that it is provide an alternate way to define variables that doesn't behave in the same way e.g. with the existing ```vars:``` you can set a variable to a literal, property, factory etc. If we could somehow meld the two that would be cool but I think having two independent ways of doing this might not be a benefit. Thoughts?

mustafakemalgilor:
> This is interesting. one bit of initial feedback i'd have is that it is providing an alternate way to define variables that doesn't behave in the same way e.g. with the existing `vars:` you can set a variable to a literal, property, factory etc. If we could somehow meld the two that would be cool but I think having two independent ways of doing this might not be a benefit. Thoughts?

I feel the same way as well. The proposition value of this patch is not to provide an alternate way to bring properties or replace any other run-time interpreted values, on the contrary, it aims to complement it. I'll experiment with the existing scenarios and see how these two will behave together.

mustafakemalgilor:
Closing this in favor of https://github.com/canonical/hotsos/pull/913

PR #859, opened 2024-05-09
[juju] Fix the affected juju versions

pponnuvel:
Two changes:
1. Juju started using Pebble only from 2.9.0
2. Rest are "off-by-one"

Ref: https://discourse.canonical.com/t/juju-pebble-security-bulletin/3503

### DIFF
--- a/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
+++ b/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
@@ -3,14 +3,14 @@ checks:
     binary:
       handler: hotsos.core.plugins.juju.JujuBinaryInterface
       juju:
-        - min: '1.0.0'
-          max: '2.9.49'
+        - min: '2.9.0'
+          max: '2.9.48'
         - min: '3.0.0'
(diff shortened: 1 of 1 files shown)


hotsos/defs/scenarios/juju/juju_binary_cve.yaml
```diff
@@ -3,14 +3,14 @@ checks:
     binary:
       handler: hotsos.core.plugins.juju.JujuBinaryInterface
       juju:
-        - min: '1.0.0'
-          max: '2.9.49'
+        - min: '2.9.0'
+          max: '2.9.48'
```
sombrafam, line 7: So, I think there's a bug in the version checker. While I was creating the tests, I tried this (2.9.48) with the juju version 2.9.48, but it wasn't catching, so I bumped it to 49.

sombrafam: I run the tests again and for some reason, they're working the way they should, so we are good to go.

sombrafam (approved):

PR #864, opened 2024-05-12
Replace deprecated resources.path()

pponnuvel:

PR #867, opened 2024-05-13
pylint: bump version to 3.1.0

mustafakemalgilor:
... to make it consistent with the `searchkit`. fixed the new warnings issued by the pylint.

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -63,6 +63,10 @@ def get_repo_info():
             return fd.read().strip()
 
     # pypi
+    # NOTE: The pylint warning is suppressed for W4902 because the
+    # alternative (i.e. resources.files) is not available for python
+    # 3.8, which is a supported environment for hotsos.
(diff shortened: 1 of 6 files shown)


brianphaley (approved):

brianphaley (approved):
Pon is fixing the resources.path issue in https://github.com/canonical/hotsos/pull/864

mustafakemalgilor:
> Pon is fixing the resources.path issue in #864

actually I tried to fix that as well (#854) , but the thing is the alternative `resources.files` is missing in py3.8 so I didn't want to introduce two different code paths just to please the pylint. unfortunately it's not a drop-in replacement so we cannot do "import this if py3.8 else that" either. I guess we have to wait until we drop the py3.8 support.

pponnuvel (approved):
It makes sense to suppress it for now (until 3.8 support is dropped).

dosaboy (approved):
lgtm!

PR #869, opened 2024-05-15
Move coverage to its own tox section

dosaboy:
Running coverage every time stestr runs gives
unexpected results and errors especially when
running a single unit test. This moves coverage
to its own section in tox.ini and adds it to the
default list of commands to be run.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 skipsdist = True
-envlist = py3,pep8,pylint,bashate,yamllint,functional
+envlist = py3,coverage,pep8,pylint,bashate,yamllint,functional
 sitepackages = False
 minversion = 3.18.0
 
@@ -24,10 +24,6 @@ deps =
 commands =
     coverage erase
     stestr run --serial --test-path {[testenv]unit_tests} {posargs}
-    coverage combine
-    coverage report
-    coverage html -d cover
-    coverage xml -o cover/coverage.xml
 
 [testenv:pep8]
 allowlist_externals = flake8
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -69,3 +64,12 @@ commands = bash {toxinidir}/tools/test/functest.sh
 
 [testenv:docs]
 commands = sphinx-build -j auto -d {toxinidir}/doc/build/doctrees -b html {toxinidir}/doc/source doc/build/html
+
+[testenv:coverage]
+commands =
+    coverage erase
```
brianphaley, line 70: Do you need to have an stestr line here to run some tests? I'm just going by the neutron section

mustafakemalgilor: I think it's not necessary to do so since we're overriding the PYTHON env variable to capture the coverage data. But still, I think we need to move the "coverage erase" line above the "stestr run" in order to ensure that we don't lose the coverage data captured by the previous run. 

Also, we need to introduce a "depends = py3" here so tox can determine the run order when i.e. "--parallel" is given.

tox.ini
```diff
@@ -22,12 +22,7 @@ deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
 commands =
-    coverage erase
```
mustafakemalgilor, line 25: I think it would make sense to keep this line since its purpose is to clean the coverage date from the old runs. The coverage data is generated from the "stestr" run (we're overriding the PYTHON to allow coverage to do that).

tox.ini
```diff
@@ -69,3 +64,12 @@ commands = bash {toxinidir}/tools/test/functest.sh
 
 [testenv:docs]
 commands = sphinx-build -j auto -d {toxinidir}/doc/build/doctrees -b html {toxinidir}/doc/source doc/build/html
+
+[testenv:coverage]
+commands =
```
mustafakemalgilor, line 69: We need a `depends = py3` here, otherwise `tox` wouldn't be able to see that `coverage` env needs `py3` is needed to be run first.

mustafakemalgilor (changes requested):

mustafakemalgilor (approved):
LGTM :+1:

PR #871, opened 2024-05-15
[tox/coverage] cleanup tmp directories from coverage

pponnuvel:
Coverage leaves the tp dirs behind if tox run is interruped.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ sitepackages = False
 minversion = 3.18.0
 
 [testenv]
+allowlist_externals = rm
 basepython = {env:TOX_PYTHON:python3}
 unit_tests = {toxinidir}/tests/unit/
 passenv =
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -18,10 +19,12 @@ setenv = VIRTUAL_ENV={envdir}
          PYFILES={toxinidir}/setup.py {toxinidir}/hotsos/ {[testenv]unit_tests}
          non-utc-tz: TZ=EST+5
          PYTHON=coverage run --source hotsos --parallel-mode
+         HOSTNAME=hostname
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
 commands =
+    rm -rf {toxinidir}/.coverage.{env:HOSTNAME}.*
```
mustafakemalgilor, line 27: We should be able to avoid this "rm" stuff. I'll submit a patch to make sure that "coverage erase" cleans these files too.

mustafakemalgilor:
@pponnuvel I've submitted #872 which should fix the issue. "coverage clean" should've deleted those files in the first place, but it failed to do so due to a misconfiguration.

PR #872, opened 2024-05-16
coverage: ensure that "coverage clean" does its job

mustafakemalgilor:
removed parameters from "coverage run" as they should be read from the .coveragerc file.

"coverage clean" now can see "coverage run" is being executed in parallel mode and can now clean the artifacts properly.

tox:testenv:coverage now runs "coverage clean" after the coverage report generation.

Supersedes #871


mustafakemalgilor:
We should also find a way to capture the interrupt signal and call the 'coverage clean'.

pponnuvel (approved):
This seems to do the job. LGTM

mustafakemalgilor:
> We should also find a way to capture the interrupt signal and call the 'coverage clean'.

This is done.

mustafakemalgilor:
Now added the optional '-coverage' to py3 so we're now able to skip coverage related stuff on py3 env run.

tox.ini
```diff
@@ -1,10 +1,11 @@
 [tox]
 skipsdist = True
-envlist = py3,coverage,pep8,pylint,bashate,yamllint,functional,hotyvalidate
+envlist = py3-{coverage},coverage,pep8,pylint,bashate,yamllint,functional,hotyvalidate
```
brianphaley, line 3: So in the py3.10 test run, the unit tests are still run twice - once in "Run with tox" (which had no coverage info) and the other with "Run coverage", so I'm not sure this is working as intended?

brianphaley (changes requested):

tox.ini
mustafakemalgilor: I checked the last CI run and couldn't see the second run. The tests were run in the "Test with tox" job, and then the "Run coverage" job only generated the reports from the collected data:

```
Run tox -e coverage
coverage create: /home/runner/work/hotsos/hotsos/.tox/coverage
coverage installdeps: -r/home/runner/work/hotsos/hotsos/requirements.txt, -r/home/runner/work/hotsos/hotsos/test-requirements.txt
coverage installed: alabaster==0.7.16,astroid==3.1.0,attrs==23.2.0,autopage==0.5.2,Babel==2.15.0,bashate==2.1.1,certifi==2024.2.2,cffi==1.16.0,charset-normalizer==3.3.2,click==8.1.7,cliff==4.7.0,cmd2==2.4.3,coverage==7.5.1,cryptography==42.0.7,dill==0.3.8,distro==1.9.0,docutils==0.21.2,extras==1.0.0,fasteners==0.19,fixtures==4.1.0,flake8==6.0.0,flake8-import-order==0.18.2,idna==3.7,imagesize==1.4.1,importlib_metadata==7.1.0,iso8601==2.1.0,isort==5.13.2,Jinja2==3.1.4,MarkupSafe==2.1.5,mccabe==0.7.0,packaging==24.0,pathspec==0.12.1,pbr==6.0.0,platformdirs==4.2.2,prettytable==3.10.0,progress==1.6,propertree==1.1.0.post5,pycodestyle==2.10.0,pycparser==2.22,pyflakes==3.0.1,Pygments==2.18.0,pylint==3.1.0,pyperclip==1.8.2,python-dateutil==2.9.0.post0,python-subunit==1.4.4,pytz==2024.1,PyYAML==6.0.1,requests==2.31.0,rpdb==0.1.6,searchkit==0.4.1.post5,simplejson==3.19.2,six==1.16.0,snowballstemmer==2.2.0,Sphinx==7.3.7,sphinxcontrib-applehelp==1.0.8,sphinxcontrib-devhelp==1.0.6,sphinxcontrib-htmlhelp==2.0.5,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.7,sphinxcontrib-serializinghtml==1.1.10,stestr==4.1.0,stevedore==5.2.0,testtools==2.7.1,tomli==2.0.1,tomlkit==0.12.5,typing_extensions==4.[11](https://github.com/canonical/hotsos/actions/runs/9130157969/job/25106403007#step:9:12).0,urllib3==2.2.1,voluptuous==0.14.2,wcwidth==0.2.13,yamllint==1.35.1,zipp==3.18.2
coverage run-test-pre: PYTHONHASHSEED='0'
coverage run-test: commands[0] | coverage combine
Combined data file .coverage.fv-az777-823.1860.XltsTZnx
coverage run-test: commands[1] | coverage report
Name                                                              Stmts   Miss Branch BrPart   Cover
----------------------------------------------------------------------------------------------------
hotsos/cli.py                                                       189    117     81     10  37.41%
```

Quick way to simulate the same thing: `tox -e py3-coverage,coverage`. This would also run the tests only once.

tox.ini
```diff
@@ -1,10 +1,11 @@
...
 sitepackages = False
 minversion = 3.18.0
 
 [testenv]
+allowlist_externals=bash, echo
```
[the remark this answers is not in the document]
mustafakemalgilor: remnants of and old test. removed.

tox.ini
```diff
@@ -17,13 +18,15 @@ setenv =
...
+    coverage: COV_ERASE=coverage erase
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
 commands =
-    coverage erase
-    stestr run --serial --test-path {[testenv]unit_tests} {posargs}
+    # Wrap stestr in bash and handle the interrupt so we can clean after it
+    coverage: coverage erase
+    bash -c 'function inth()\{ {env:COV_ERASE:/bin/true}; \}; trap inth INT; stestr run --serial --test-path {[testenv]unit_tests} {posargs}'
```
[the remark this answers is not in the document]
mustafakemalgilor: It's weird since it's already listed in `allowlist_externals`. I'll use the explicit path of the bash, i.e. "/usr/bin/bash" to see whether that removes the warning.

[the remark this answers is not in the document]
mustafakemalgilor: Yeah, that gets rid of the warning.

tox.ini
mustafakemalgilor: The only time that tests run for a second time is for the "non-utc-tz" job, which runs all the tests again in a different time zone than UTC (we had some issues in the past due to the timezone...).

mustafakemalgilor: @brianphaley let me know if I'm missing something.

mustafakemalgilor: > So I can still run 'tox -e py3' correct?
Yes, correct. 


> Maybe we can just add a comment to that effect?
Sure, I'll add a comment to describe what we're doing.

tox.ini
```diff
@@ -1,10 +1,12 @@
 [tox]
 skipsdist = True
-envlist = py3,coverage,pep8,pylint,bashate,yamllint,functional,hotyvalidate
+envlist = py3-{coverage},coverage,pep8,pylint,bashate,yamllint,functional,hotyvalidate
 sitepackages = False
 minversion = 3.18.0
 
 [testenv]
+allowlist_externals = 
+    /usr/bin/bash
```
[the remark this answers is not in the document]
mustafakemalgilor: Yeah, seems like that way. Replaced "/usr/bin/bash" with "bash", no warnings.

brianphaley (approved):
Thanks for all the work Mustafa, and coverage is at 84.96% :)

nicolasbock (approved):

tox.ini
dosaboy: The curly brackets are not actually needed around {coverage} are they?

Also, from my testing this does all seem to work well (I havent checked the github actions yet and will do next) but for the sake of intuitiveness if like to rename py3-coverage to something clearer i.e. when it is run it is actually doing a coverage erase then running unit tests and it in fact exists as a preliminary stage for the converage testenv so maybe we should rename it to py3-pre-coverage so something like that.

dosaboy: ok im wrong :) py3-covergage both cleans and generates coverage data so this is fine. The curl brackets are indeed harmless but also unnecessary it seems according to https://tox.wiki/en/4.15.0/faq.html#platform-specification since they would be needed if more than one item were specified comma-seperated.

dosaboy (approved):

mustafakemalgilor:
So I've made a few significant changes:

- Coverage now writes its intermediate files to the tox's temp directory, so no tidying would be necessary (removed the bash interrupt handler thingy)
- Moved .coveragerc contents to tox.ini
- Renamed tox "coverage" env to "coveragereport"

I think the patch is now much cleaner.

dosaboy:
nice cleanups in the final stage! lgtm!

PR #875, opened 2024-05-17
Fix the way coverage is run

brianphaley:
Move all coverage commands into the 'coverage' section of tox.ini, else it will be run for every unit test run, for example, 'tox -e py3'.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,12 +17,10 @@ setenv =
     PYTHONHASHSEED=0
     TESTS_DIR={[testenv]unit_tests}
     non-utc-tz: TZ=EST+5
-    PYTHON=coverage run --source hotsos --parallel-mode
 deps =
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
 commands =
-    coverage erase
     stestr run --serial --test-path {[testenv]unit_tests} {posargs}
 
 [testenv:pep8]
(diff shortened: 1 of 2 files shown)


mustafakemalgilor:
The existing method allows us to capture coverage data on the first run, so I think it's redundant to run the whole unit test suite again just for the coverage.

tox.ini
```diff
@@ -61,11 +59,16 @@ commands = sphinx-build -j auto -d {toxinidir}/doc/build/doctrees -b html {toxin
 
 [testenv:coverage]
 depends = py3
```
dosaboy, line 61: since we are running the unit tests here now do we need to depend on py3 still?

brianphaley: Don't know about the py3 depends, can take a look, but probably don't need it.

I can understand only wanting to run the unit tests once and generate coverage, but it means I can't run them locally and not generate coverage. I don't know what the best thing to do here is.

mustafakemalgilor: > I can understand only wanting to run the unit tests once and generate coverage, but it means I can't run them locally and not generate coverage. I don't know what the best thing to do here is.

Yeah, it's not ideal. If there's a way to check if the 'coverage' environment is being queued in tox, we should be able to avoid running the coverage related stuff in the test run. I'll look into that.

Nevertheless, I believe we still should avoid running the tests again since it both extends the CI run duration and the default tox run's duration.

mustafakemalgilor:
Hi @brianphaley 

I addressed your concerns in #872, please have a look.

brianphaley:
#872 is good now, will close this one

PR #879, opened 2024-05-20
Fix BFD state change check

dosaboy:
The way we were checking for bfd state transitions was incorrectly counting the number of actual transitions. What we really care about is the the state transitions to down. The logic to do this has had to be moved into code since it is not possible to do it purely in yaml.

### DIFF
--- a/hotsos/core/plugins/openvswitch/ovs.py
+++ b/hotsos/core/plugins/openvswitch/ovs.py
@@ -1,6 +1,9 @@
+import os
 import re
+import uuid
 from functools import cached_property
 
+from hotsos.core.config import HotSOSConfig
 from hotsos.core.factory import FactoryBase
 from hotsos.core.log import log
 from hotsos.core.host_helpers import (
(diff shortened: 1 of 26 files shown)


hotsos/core/plugins/openvswitch/ovs.py
```diff
@@ -200,6 +207,70 @@ def offload_enabled(self):
         return False
 
 
+class OVSBFD(OpenvSwitchBase):
+    # pylint: disable=C0301
```
mustafakemalgilor, line 211: Instead of disabling the long line warning we can allow unbreakable long lines (such as URLs) with the following pylint option:

https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#ignore-long-lines

dosaboy: ill try it out

dosaboy:
i plan to update this patch to leverage the new global searcher

Issue #882, opened 2024-05-22
error: "failed to import and call property hotsos.core.plugins.kernel.memory.MemInfo.hugetlb_to_mem_total_percentage"


mustafakemalgilor:
Hotsos encountered an error processing a sosreport from xenial/pike environment:

```
2024-05-22 13:12:23,552 113086 DEBUG hotsos.plugin.kernel [-] instantiating class hotsos.core.plugins.kernel.memory.MemInfo (from_cache=False)
2024-05-22 13:12:23,552 113086 ERROR hotsos.plugin.kernel [-] failed to import and call property hotsos.core.plugins.kernel.memory.MemInfo.hugetlb_to_mem_total_percentage
Traceback (most recent call last):
  File "/workspace/github/hotsos/hotsos/core/ycheck/engine/properties/common.py", line 442, in get_property
    _obj = getattr(_obj, _prop)
           ^^^^^^^^^^^^^^^^^^^^
  File "/workspace/github/hotsos/hotsos/core/plugins/kernel/memory.py", line 40, in __getattr__
    raise AttributeError('attribute {} not found in {}.'.
AttributeError: attribute hugetlb_to_mem_total_percentage not found in MemInfo.. Did you mean: 'mem_avail_to_mem_total_percentage'?
2024-05-22 13:12:23,552 113086 ERROR hotsos.plugin.kernel [-] function '_result' raised exception: attribute hugetlb_to_mem_total_percentage not found in MemInfo.
2024-05-22 13:12:23,552 113086 ERROR hotsos.plugin.kernel [-] requires.YPropertyVarOps.result raised the following
2024-05-22 13:12:23,552 113086 ERROR hotsos.plugin.kernel [-] caught exception when running requirement: attribute hugetlb_to_mem_total_percentage not found in MemInfo.
2024-05-22 13:12:23,553 113086 ERROR hotsos.plugin.kernel [-] something went wrong while executing check too_many_free_hugepages
Traceback (most recent call last):
  File "/workspace/github/hotsos/.tox/py3/lib/python3.12/site-packages/propertree/propertree2.py", line 346, in __getattribute__
    return self._get_override_attribute(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/github/hotsos/.tox/py3/lib/python3.12/site-packages/propertree/propertree2.py", line 589, in _get_override_attribute
    return super()._get_override_attribute(name)
[...]

PR #884, opened 2024-05-22
_BaseProcKeyValue must not raise AttributeError for valid keys

dosaboy:
Implementations of the _BaseProcKeyValue base class need to specify which keys they expect to find in the target they are reading so that if not found they dont raise an AttributeError as this gets bubbled up to caller unecessarily.

Resolves: #882

### DIFF
--- a/hotsos/core/plugins/kernel/memory.py
+++ b/hotsos/core/plugins/kernel/memory.py
@@ -6,6 +6,11 @@ from hotsos.core.utils import sorted_dict
 
 
 class _BaseProcKeyValue(object):
+    # Set to list of keys we expect to find in the file. If these keys are not
+    # found their value will be returned as DEFAULT_RETURN_VALUE instead of
+    # raising an AttributeError.
+    VALID_KEYS = []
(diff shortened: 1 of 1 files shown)


hotsos/core/plugins/kernel/memory.py
```diff
@@ -37,11 +42,15 @@ def __getattr__(self, key):
                     self._proc_file_keys_values[key] = value
                     return value
 
+        if key in self.VALID_KEYS:
+            return self.DEFAULT_RETURN_VALUE
```
brianphaley, line 46: nit: I guess this could just be 'return 0' as is done above on L36 since sub-classes don't seem to override it anywhere.

brianphaley (approved):

hotsos/core/plugins/kernel/memory.py
dosaboy: They don't currently override it but i thought should make it possible to do so since different files targets might want a different default. In reality they will probably all want 0 though yeah.

PR #893, opened 2024-06-04
config/SectionalConfigBase: replace DIY ini parser with configparser

mustafakemalgilor:
SectionalConfigBase is a MacGyver'd INI parser in disguise. Python has a reliable way of reading ini files, which is the `configparser` standard module. This patch replaces the regex-based custom ini parsing algorithm with configparser. Renamed the SectionalConfigBase class to IniConfigBase to better reflect what it *really* is.

Removed "all" function and added "all_keys" function instead.

### DIFF
--- a/hotsos/core/host_helpers/config.py
+++ b/hotsos/core/host_helpers/config.py
@@ -1,6 +1,6 @@
 import abc
 import os
-import re
+import configparser
 
 from hotsos.core.log import log
 
@@ -80,14 +80,11 @@ class ConfigBase(abc.ABC):
         """ Get a config value. """
 
 
-class SectionalConfigBase(ConfigBase):
+class IniConfigBase(ConfigBase):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._sections = {}
-        # this provides an easy sectionless lookup but is prone to collisions.
-        # always returns the last value for key found in config file.
-        self._flattened_config = {}
+        self.config = None
         self._load()
 
     @staticmethod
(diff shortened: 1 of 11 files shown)


hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,62 @@ def bool_str(val):
...
+
+        # 2: No section specified.
+        # The default section is not included in the section
+        # set so append it
+        for sec in self.config.sections() + ["DEFAULT"]:
+            v = self.config.get(sec, key, fallback=None)
+            log.debug("ini read value -> %s:%s = %s", section, key, v)
+            if v is None:
+                continue
+            return post_processing(v)
```
brianphaley, line 144: nit: could change to an affirmative check and avoid the continue

if v:
    return post_processing(v)

hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,62 @@ def bool_str(val):
...
+
+            if expand_to_list:
+                return self.expand_value_ranges(v)
+
+            return v
+
+        # 1: Section is specified.
+        if section is not None:
+            # Look for section name, case-insensitive
+            for sect in (section.upper(), section.lower()):
```
brianphaley, line 129: Interesting we have to do upper() and lower(), is that to match 'DEFAULT'? I noticed default_section is an argument to ConfigParser so just curious.

mustafakemalgilor: configparser is case-sensitive so the code tries both the lower-case and upper-case to check both instances to avoid the hassle.  I noticed that this would not work for e.g. CamelCase section names. I'll change this to make the comparison in upper_case().

hotsos/core/host_helpers/config.py
mustafakemalgilor: Done

hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,62 @@ def bool_str(val):
...
+            log.debug("ini read value -> %s:%s = %s", section, key, v)
+            if v is None:
+                continue
+            return post_processing(v)
+
+        return None
 
     def _load(self):
         if not self.exists:
+            log.debug("config file %s not exists", self.path)
```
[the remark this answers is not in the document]
mustafakemalgilor: Done

hotsos/core/host_helpers/config.py
mustafakemalgilor: Done

hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,61 @@ def bool_str(val):
...
+            v = self.config.get(sec, key, fallback=None)
+            log.debug("ini read value -> %s:%s = %s", section, key, v)
+            if v:
+                return post_processing(v)
+
+        return None
 
     def _load(self):
         if not self.exists:
+            log.debug("config file %s does not exists", self.path)
```
pponnuvel, line 149: s/exists/exist

hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,61 @@ def bool_str(val):
...
+
+            if expand_to_list:
+                return self.expand_value_ranges(v)
+
+            return v
+
+        # 1: Section is specified.
+        if section is not None:
+            # Look for section name, case-insensitive
+            for sect in self.config.sections() + ["DEFAULT"]:
```
pponnuvel, line 129: When section is given, do we need DEFAULT here?

mustafakemalgilor: Since we're not using the `get()` to retrieve data, we need "DEFAULT" as default is a valid section name that the user can explicitly give.

pponnuvel (approved):

hotsos/core/host_helpers/config.py
```diff
@@ -100,62 +97,61 @@ def bool_str(val):
...
-        with open(self.path) as fd:
-            return fd.read()
+        """ If section is None, then search all sections and return first
+        match. """
+        log.debug("ini read call -> %s:%s", section, key)
+
+        if not self.config:
+            return None
+
+        def post_processing(v):
```
dosaboy, line 112: inline functions are hard to test, can we move this outside please?

mustafakemalgilor: this inline function is just a named lambda in reality, but fair enough.

mustafakemalgilor: done

dosaboy (approved):
lgtm!

PR #897, opened 2024-06-10
Re-enable no-member rule in pylint

dosaboy:


pponnuvel (approved):
LGTM

pylintrc
```diff
@@ -31,7 +31,6 @@ disable=
  missing-class-docstring,
  missing-function-docstring,
  missing-module-docstring,
- no-member,
```
brianphaley, line 34: Just a question - adding the disable=E1101 statements has the same affect as this one disable, are there other places we need to have this enabled?

dosaboy: Thanks Brian. So I this rule really needs enabling because it is very easy/common to add an attribute that does not exist (i did this recently and pylint did not catch it because of this being disabled which is why I am keen to re-enable it). I know it is ugly having to add all these inline disables but they are a consequence of using baseclasses with the propertree library and I don't expect that we will need to do this normally. I did tinker with modifying the code to reduce the number of inline disables but ended up breaking things so reverted to this approach for now.

PR #898, opened 2024-06-10
Catch and verify log calls

brianphaley:
There are some tests for invalid data that cause either log.warning() or log.error() to be called. In order to not pollute test output (these are shown in RED), mock the calls and actually verify they were called.

In the case of template-generated tests, just mock the log call to prevent the output.

TrivialFix


tests/unit/test_host_helpers.py
```diff
@@ -259,8 +259,11 @@ def test_get_date_w_tz(self):
 
     @utils.create_data_root({'sos_commands/date/date':
                              'Thu Mar 25 10:55:05 123UTC 2021'})
-    def test_get_date_w_invalid_tz(self):
+    @mock.patch('hotsos.core.host_helpers.cli.log.error')
+    def test_get_date_w_invalid_tz(self, mock_log_error):
```
mustafakemalgilor, line 263: Instead of mocking logger instances, you can use unittest `self.assertLogs(...)` function. The `test_get_date_w_invalid_tz` can be rewritten as follows:

```python
    @utils.create_data_root({'sos_commands/date/date':
                             'Thu Mar 25 10:55:05 123UTC 2021'})
    def test_get_date_w_invalid_tz(self):
        with self.assertLogs(logger="hotsos", level='ERROR') as log:
            self.assertEqual(host_helpers.cli.CLIHelper().date(), "")
            self.assertEqual(len(log.output), 1)
            self.assertIn('has invalid date string', log.output[0])
```

mustafakemalgilor (approved):
LGTM 👍

PR #901, opened 2024-06-12
Support scenarios using jsondecode commands

dosaboy:
Some scenarios use jsondecode commands which writes the output to a file. This was failing if the
decoded value was a dict so now we re-encode the
dict to json if we need to write it to a file. Also ensures that if the search preloaders raise an exception we catch it, log it and move on to.


wilkmar (approved):
lgtm

PR #902, opened 2024-06-14
scenario: check if "ldap_use_tokengroups" is enabled for ad domains

### LINKED ISSUE #890: Issue a warning when "ldap_use_tokengroups" is "true" on sssd-ad deployments
The use of token groups is known to be causing issues with group memberships[1]:

```
After selecting a custom ldap_search_base, the group membership no longer displays correctly.

If you use a non-standard LDAP search bases, please disable the TokenGroups performance enhancement by setting ldap_use_tokengroups=False. Otherwise, the AD provider would receive the group membership via a special call that is not restricted by the custom search base which causes unpredictable results

Typically, users configure a custom ldap_search_base to limit the groups the user is a member of. Please see [this blog post](https://jhrozek.wordpress.com/2016/12/09/restrict-the-set-of-groups-the-user-is-a-member-of-with-sssd/)formoreinformationonthesubject).

```
We had several issues where the root cause was attributed to this. It would be nice to raise an issue when "sssd-ad" is present in the environment and the "ldap_use_tokengroups" option is set to "true". 

[1]: https://sssd.io/troubleshooting/ad_provider.html

mustafakemalgilor:
added a new sssd plugin to retrieve a list of domains that ldap_use_ tokengroups is enabled for. added a new scenario to issue a warning when a sssd domain is using "ad" id provider and ldap_use_tokengroups is not explicitly disabled.

Fixes #890

### DIFF
--- a/hotsos/core/host_helpers/config.py
+++ b/hotsos/core/host_helpers/config.py
@@ -87,6 +87,9 @@ class IniConfigBase(ConfigBase):
         self.config = None
         self._load()
 
+    def __bool__(self):
+        return self.config is not None
+
     @staticmethod
     def bool_str(val):
         if val.lower() == "true":
@@ -112,6 +115,10 @@ class IniConfigBase(ConfigBase):
         return v
 
     @property
+    def all_sections(self):
+        return self.config.sections() + ["DEFAULT"]
+
+    @property
     def all_keys(self):
         return [x for option in self.config.items()
                 for x in list(option[1].keys())]
@@ -127,7 +134,7 @@ class IniConfigBase(ConfigBase):
         # 1: Section is specified.
         if section is not None:
             # Look for section name, case-insensitive
-            for sect in self.config.sections() + ["DEFAULT"]:
+            for sect in self.all_sections:
                 if section.upper() == sect.upper():
                     return self.post_processing(
                         self.config.get(sect, key, fallback=None),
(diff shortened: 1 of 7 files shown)


dosaboy:
I'd like to hear your thoughts in this but I feel that instead of creating a new plugin for "auth", this might fit better in the existing "system" plugin. My rationale for this is mainly that authentication in this context is really a function of the system and not the individual applications for which we have independent plugins. I also feel that regardless, the output of this new addition should appear immediately after (if not within) the "system:" output and currently it is going to be right at the and after "kernel:" at plugin_root_index 15. So at the very least we'd need to shift all plugin_root_index by one and put this new one at plugin_root_index 2 OR we just incorporate this into the system plugin. Thoughts?

mustafakemalgilor:
> I'd like to hear your thoughts in this but I feel that instead of creating a new plugin for "auth", this might fit better in the existing "system" plugin. My rationale for this is mainly that authentication in this context is really a function of the system and not the individual applications for which we have independent plugins. I also feel that regardless, the output of this new addition should appear immediately after (if not within) the "system:" output and currently it is going to be right at the and after "kernel:" at plugin_root_index 15. So at the very least we'd need to shift all plugin_root_index by one and put this new one at plugin_root_index 2 OR we just incorporate this into the system plugin. Thoughts?

I was unsure that sssd would fit into that category but since you've confirmed that "system" is intended to be used the way you've described, we can just put it there. I'll make the changes.

mustafakemalgilor:
> we just incorporate this into the system plugin

@dosaboy done

PR #908, opened 2024-06-22
Improves kernlog search efficiency

dosaboy:
Some of the CallTrace handlers were unnecessarily storing the contents of every line in kern.log which massively slows down the search and consumes memory.


mustafakemalgilor (approved):
LGTM

PR #917, opened 2024-07-02
Re-enables pylint too-many-nested-blocks check

dosaboy:


hotsos/client.py
```diff
@@ -63,34 +63,37 @@ def _get_very_short_format(self, summary):
         filtered = {}
         for plugin in summary:
             for key in self.FILTER_SCHEMA:
-                if key in summary[plugin]:
-                    if key not in filtered:
-                        filtered[key] = {}
+                if key not in summary[plugin]:
```
brianphaley, line 66: My only nit is the code in _get_short_format() is the same (doesn't cause pylint issue) so could be changed similarly. But it's only a nit...

brianphaley (approved):

PR #924, opened 2024-07-03
Re-enables pylint inconsistent-return-statements check

dosaboy:


mustafakemalgilor (approved):

hotsos/core/host_helpers/systemd.py
```diff
@@ -100,7 +100,7 @@ def start_time(self):
             if len(sections) == 0:
                 log.warning("no active status found for %s.service (state=%s)",
                             self.name, self.state)
-                return
+                return None
```
pponnuvel, line 103: `None` is the default value, right? Are we emphasising explicit returns? Maybe, a Coding Guidelines, if one doesn't exist, might be useful for future (perhaps Canonical-wide given the amount of Python we write).

hotsos/core/plugins/kernel/sysfs.py
```diff
@@ -16,37 +16,41 @@ def get(relpath):
...
         path = os.path.join(HotSOSConfig.data_root, 'sys', relpath)
-        if not os.path.exists(path):
-            return
+        if os.path.exists(path):
+            with open(path) as fd:
+                value = fd.read()
 
-        with open(path) as fd:
-            value = fd.read()
+            return value.strip()
```
pponnuvel, line 23: Better to put this return "inside" `with` ?

hotsos/core/plugins/sosreport.py
```diff
@@ -31,16 +31,20 @@ def data_root_is_sosreport(self):
...
         path = os.path.join(HotSOSConfig.data_root, 'version.txt')
         if not os.path.exists(path):
-            return
+            return None
 
         with open(path) as fd:
             for line in fd:
-                if line.startswith('sosreport:'):
-                    return line.partition(' ')[2].strip()
+                if not line.startswith('sosreport:'):
```
pponnuvel, line 42: In future, one might very well think the LHS is better and rewrite that way again ;-) (except the final `return None` of course).

hotsos/core/plugins/storage/bcache.py
```diff
@@ -19,9 +19,11 @@ class BcacheConfig(ConfigBase):
 
     def get(self, key, section=None, expand_to_list=False):
         cfg = os.path.join(self.path, key)
-        if os.path.exists(cfg):
-            with open(cfg) as fd:
-                return fd.read().strip()
+        if not os.path.exists(cfg):
+            return None
+
+        with open(cfg) as fd:
```
pponnuvel, line 25: Following the same, wouldn't you want a final `return None` in case `open` fails?

hotsos/core/plugins/kernel/sysfs.py
dosaboy: fixed

hotsos/core/plugins/storage/bcache.py
dosaboy: if open() fails it will throw an exception

PR #925, opened 2024-07-04
cli/get_repo_info: suppress importlib.path deprecation warning

xmkg:
we're aware that the importlib.path is being deprecated but we still want to support python 3.8, which does not have the `as_file` function.

also noticed that when the resources.path() used in a with clause, it raises an exception when the file is not present in Python 3.8.10, which is the latest python 3.8 version in Focal at the moment. Therefore, used contextlib.suppress(Exception) to deal with that as well.

Relevant past MP's: #864, #854

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import sys
 import threading
+import warnings
 from importlib import metadata, resources
 from dataclasses import dataclass, fields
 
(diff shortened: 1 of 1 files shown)


mustafakemalgilor:
@dosaboy rebased this one, should be trivial enough to land.

nicolasbock (approved):

pponnuvel (approved):
LGTM

xmkg:
@dosaboy feel free to land or close this one, and let me know should anything else is needed.

PR #929, opened 2024-07-05
hotsos/core: add name aliasing support

### LINKED ISSUE #912: Add property aliasing to Python properties
Currently, property names are long because they must be the full import path to the property, e.g.:

```yaml
@hotsos.core.plugins.kernel.memory.MemInfo.mem_avail_to_mem_total_percentage
```

The given example property is implemented in the MemInfo class as follows:

`hotsos/core/plugins/kernel/memory/MemInfo.py`:
```python
class MemInfo(_BaseProcKeyValue):
    # /* .... */
    @property
    def mem_avail_to_mem_total_percentage(self):
        return round((self.MemAvailable * 100) / self.MemTotal)
```

We could allow aliasing such properties to a shorter, more comprehensible name with a decorator, e.g.:

`hotsos/core/plugins/kernel/memory/MemInfo.py`:
```python
class MemInfo(_BaseProcKeyValue):
    # /* .... */
    @property
    @alias('kernel.mem.avail_to_total_pct')
    def mem_avail_to_mem_total_percentage(self):
        return round((self.MemAvailable * 100) / self.MemTotal)
```

... so we can refer to it in YAML files like:

```yaml
@kernel.mem.avail_to_total_pct
```

xmkg:
at the moment, scenarios are using the long import paths in order to reference to a Python property. this feature allows assigning an alias to a Python class or variable in order to make using plugin property interfaces easier.

Added aliases to the plugins.
Updated existing scenarios to use aliases.

Fixes #912


hotsos/core/plugins/kernel/memory.py
```diff
@@ -63,6 +65,7 @@ def path(self):
         return os.path.join(HotSOSConfig.data_root, 'proc/vmstat')
 
     @property
+    @alias("kernel.vmstat.compaction_failures_pct")
```
pponnuvel, line 68: I think this isn't needed since the class is aliased. Can be directly used as `kernel.vmstat.compaction_failures_percent`.

pponnuvel (approved):
The aliasing naming seems a bit odd to me. For example, `juju.base`, `kernel.calltrace`, `ceph.daemon.all-osds` and so on give the impression it's a package/class hierarchy when they are just a given name.

Given we could mix with the original way of writing/accessing the properties, the use of dots seems a little misleading.

Perhaps we can drop the dot (`A.B`) in aliases and use a single word name such as `jujuBase`/`jujubase`/`juju_base` or something like that?

hotsos/core/plugins/kernel/memory.py
mustafakemalgilor: nice catch, I'll remove it

mustafakemalgilor:
> The aliasing naming seems a bit odd to me. For example, juju.base, kernel.calltrace, 'ceph.daemon.all-osds` and so on give the impression it's a package/class hierarchy when they are just a given name.

> Given we could mix with the original way of writing/accessing the properties, the use of dots seems a little misleading.

The dot separation represents a hierarchy, where the first part is the plugin name, then the optional subsections, and finally the property name. I made it this way so if we write a YAML schema(*another important QOL improvement that will make it easier to write scenarios) for the scenarios for IDE/editor assistance in the future, it would be convenient for tab completion.

As for the possible confusion, we can easily distinguish an alias from the full path as the package names always have to start with `hotsos.` (*unless we start supporting external plugins), so if it starts with "hotsos", it's a full property path, otherwise, it's an alias.

pponnuvel:
> The dot separation represents a hierarchy, where the first part is the plugin name, then the optional subsections, and finally the property name. 

Now you say that, it makes sense. But it's not obvious otherwise. Also nothing prevents anyone from giving arbitrary names. So might be useful to have a section ("Alias naming") in documentation on the naming convention.

xmkg:
@dosaboy if it makes sense to land this one just let me know and I'll rebase it

PR #936, opened 2024-07-10
pylintrc: re-enable unnecessary-lambda

mustafakemalgilor:
suppressed the warning for the legit usages.

### DIFF
--- a/tests/unit/test_ycheck_properties.py
+++ b/tests/unit/test_ycheck_properties.py
@@ -779,9 +779,13 @@ class TestYamlProperties(utils.BaseTestCase):
 
     def test_get_datetime_from_result(self):
         result = mock.MagicMock()
+        # The lambda is necessary here. _result is changing so
+        # we can't bind result's get() function directly, otherwise
+        # we'd have to update the side effect too.
+        # pylint: disable-next= unnecessary-lambda
         result.get.side_effect = lambda idx: _result.get(idx)
-
         _result = {1: '2022-01-06', 2: '12:34:56.123'}
+
         ts = ExtraSearchConstraints._get_datetime_from_result(result)
         self.assertEqual(ts, datetime.datetime(2022, 1, 6, 12, 34, 56))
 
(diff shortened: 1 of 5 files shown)


mustafakemalgilor:
reference: https://docs.python.org/3/library/functions.html#open

brianphaley (approved):

tests/unit/test_ycheck_properties.py
```diff
@@ -779,9 +779,13 @@ def test_yaml_def_requires_grouped(self):
 
     def test_get_datetime_from_result(self):
         result = mock.MagicMock()
+        # The lambda is necessary here. _result is changing so
+        # we can't bind result's get() function directly, otherwise
+        # we'd have to update the side effect too.
+        # pylint: disable-next= unnecessary-lambda
```
dosaboy, line 785: the correct way to do this is to use a "fake" e.g.


```
_result = {}

def fake_get(idx):
   return _result.get(idx)

result.get.side_effect = fake_get
...
```

mustafakemalgilor: I think you've just re-invented lambda 😄

dosaboy: it wants you to use functions/methods not lambda

dosaboy: and fwiw i am 100% against nested functions except for the specific use-case of setting a side_effect in a test case.

dosaboy: im happy to merge and fix later

dosaboy (approved):

tests/unit/test_ycheck_properties.py
mustafakemalgilor: I think it's complaining because it believes the usage of lambda is redundant, but in fact, it is not. _result has runtime polymorphism and Pylint isn't smart enough to figure that out. it's the limitation of static analysis and Pylint cannot infer that without the help of runtime profiling. suppressing the warning is a legitimate way here. nevertheless, I think converting it to a nested function is just postponing the suppression, because someday Pylint may start to complain about it too.

PR #937, opened 2024-07-10
pylintrc: re-enable unspecified-encoding

mustafakemalgilor:
explicitly request "utf-8" encoding for all open() calls. When unspecified, Python retrieves the platform-dependent value via locale.getencoding() call, which should evaluate to "utf-8" for Linux-y OS'es.


mustafakemalgilor:
reference: https://docs.python.org/3/library/functions.html#open

brianphaley (approved):

pponnuvel (approved):

PR #939, opened 2024-07-10
pylintrc: re-enable too-many-lines

brianphaley:
To get under the 1000 line limit I did the following:

1) Moved exception classes to more specific files
2) Moved common definitions to common files
3) Moved ceph classes into more specific files
4) Moved unit tests to more specific files, following
   the model of having the unit test live in the same
   directory structure of the file it is testing
5) Moved some definitions into tests.utils
6) More, but all along these lines

There could probably be more breakdown of the tests into specific files, but I only did what was necessary to pass.


brianphaley:
This is going to need  re-base once some of these other patches merge, just wanted to get something out before the weekend

hotsos/core/host_helpers/cli.py
```diff
@@ -1,127 +1,32 @@
...
 from hotsos.core.config import HotSOSConfig
-from hotsos.core.host_helpers.common import HostHelpersBase
+from hotsos.core.host_helpers.common import (
+    CmdOutput,
+    get_ps_axo_flags_available,
+    HostHelpersBase,
+    reset_command,
+    run_post_exec_hooks,
+    run_pre_exec_hooks,
+    SourceRunner,
```
dosaboy, line 19: I feel like these are not general enough to belong in host_helpers.common and would be better placed in e.g. host_helpers.cli_common.py

hotsos/core/host_helpers/cli.py
```diff
@@ -1,127 +1,32 @@
...
+)
+from hotsos.core.host_helpers.exceptions import (
+    catch_exceptions,
+    CLI_COMMON_EXCEPTIONS,
+    CLIExecError,
+    CommandNotFound,
+    SourceNotFound,
```
dosaboy, line 26: same as above for these

hotsos/core/plugins/openstack/openstack.py
```diff
@@ -262,9 +276,10 @@
                   'keystone': KEYSTONE_EXCEPTIONS,
                   'manila': MANILA_EXCEPTIONS,
                   'masakari': MASAKARI_EXCEPTIONS,
-                  'neutron': NEUTRON_EXCEPTIONS + OVSDBAPP_EXCEPTIONS,
+                  'neutron': NEUTRON_EXCEPTIONS + OVSDBAPP_EXCEPTIONS +
+                  NEUTRONCLIENT_EXCEPTIONS,
                   'nova': NOVA_EXCEPTIONS + PYTHON_LIBVIRT_EXCEPTIONS +
-                  NEUTRONCLIENT_EXCEPTIONS + OS_VIF_EXCEPTIONS,
+                  OS_VIF_EXCEPTIONS,
```
dosaboy, line 282: i think we need to leave NEUTRONCLIENT_EXCEPTIONS in the nova set since we would expect to find those exceptions in the nova logs but not the neutron logs right?

brianphaley:
I split this change up into many patches, can abandon this now.

PR #941, opened 2024-07-12
pylintrc: re-enable too-many-arguments

mustafakemalgilor:
started using dataclass types for grouping long list of arguments.

Suppressed the warning for test functions.

### DIFF
--- a/hotsos/cli.py
+++ b/hotsos/cli.py
@@ -5,6 +5,7 @@ import subprocess
 import sys
 import threading
 from importlib import metadata, resources
+from dataclasses import dataclass, fields
 
 import click
 import distro
@@ -15,7 +16,7 @@ from hotsos.core.config import HotSOSConfig
 from hotsos.core.log import log, LoggingManager
 from hotsos.client import (
     HotSOSClient,
-    OutputManager,
+    SUPPORTED_SUMMARY_FORMATS
 )
 
 SNAP_ERROR_MSG = """ERROR: hotsos is installed as a snap which only supports
@@ -116,7 +117,7 @@ def set_plugin_options(f):
 
 
 def get_defs_path():
-    """ Get patch to HotSOS defs. """
+    """ Get path to HotSOS defs. """
 
     # source
     defs = os.path.join(get_hotsos_root(), 'defs')
@@ -188,7 +189,53 @@ def progress_spinner(show_spinner, path):
         thread.join()
 
 
-def main():  # pylint: disable=R0915
+@dataclass(frozen=True)
+class CLIArgs:
+    """Command line arguments."""
+
+    data_root: click.Path
+    version: bool
+    defs_path: str
+    templates_path: str
+    all_logs: bool
+    debug: bool
+    quiet: bool
+    save: bool
+    output_format: str
+    html_escape: bool
+    short: bool
+    very_short: bool
+    force: bool
+    event_tally_granularity: str
+    max_logrotate_depth: int
+    max_parallel_tasks: int
+    list_plugins: bool
+    machine_readable: bool
+    output_path: str
+    command_timeout: int
+    sos_unpack_dir: str
+    scenario: str
+    event: str
+
+    @classmethod
+    def filter_kwargs(cls, **kwargs):
+        """Filter the given kwargs by the names present on this
+        class and produce two distinct dict's where the first
+        dictionary is the fields present, the second is the
+        remainder part. Used for filtering the kwargs to separate
+        the arguments needed for this class and the other excess
+        arguments."""
+
+        def_fields = {field.name for field in fields(cls)}
+        cleaned_kwargs = {
+            key: value for key, value in kwargs.items() if key in def_fields
+        }
+        remainder = {k: kwargs[k] for k in set(kwargs) - set(cleaned_kwargs)}
+        return (cleaned_kwargs, remainder)
+
+
+# pylint: disable=R0915
+def main():
     @click.command(name='hotsos')
     @click.option('--event', default='',
                   help=('Filter a particular event name. Useful for '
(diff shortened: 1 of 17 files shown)


hotsos/client.py
```diff
@@ -116,46 +129,41 @@ def minimise(self, summary, mode):
...
+    def get(self, params: OutputManagerParameters = OutputManagerParameters()):
+        if params.plugin:
+            filtered = {params.plugin: self._summary[params.plugin]}
         else:
             filtered = self._summary
 
-        if minimal_mode:
-            filtered = self.minimise(filtered, minimal_mode)
-
-        if fmt not in self.SUMMARY_FORMATS:
```
mustafakemalgilor, line 129: This is no longer needed as we now enforce fmt to be one of the values in the `SUPPORTED_SUMMARY_FORMATS`

hotsos/client.py
```diff
@@ -116,46 +129,41 @@ def minimise(self, summary, mode):
         log.warning("Unknown minimalmode '%s'", mode)
         return summary
 
-    def get(self, fmt='yaml', html_escape=False, minimal_mode=None,
-            plugin=None, max_level=2):
-        if plugin:
-            filtered = {plugin: self._summary[plugin]}
+    def get(self, params: OutputManagerParameters = OutputManagerParameters()):
```
dosaboy, line 132: Passing mutable types as param defaults is not a good idea imho for the simple reason that changes (however accidental) to the parameter made during one call of get() will persist to future calls to get(). It would be safer to have the caller pass it in rather than set a default.

hotsos/core/host_helpers/network.py
```diff
@@ -105,7 +109,8 @@ def stats(self):
         s = FileSearcher()
         seqdef = SequenceSearchDef(
                     # match start of interface
-                    start=SearchDef(IP_IFACE_NAME_TEMPLATE.format(self.name)),
+                    start=SearchDef(IP_IFACE_NAME_TEMPLATE.format(
+                        self.name)),
```
dosaboy, line 113: nit: unnecessary?

hotsos/client.py
dosaboy:
```
from dataclasses import dataclass

@dataclass
class MyOptions:
    flag: bool = False


def foo(flag, opt: MyOptions = MyOptions()):
    if flag:
        opt.flag = True
    else:
        print(opt.flag)

foo(False)
foo(True)
foo(False)
```
 produces

```
foo(False)
foo(True)
foo(False)
```

hotsos/plugin_extensions/openstack/agent/events.py
```diff
@@ -241,14 +247,19 @@ def __call__(self, event):
                 results.append({'date': e.get(1), 'time': e.get(2),
                                 'key': lb_id})
 
-            ret = self.categorise_events(event, results=results,
-                                         key_by_date=False)
+            ret = self.categorise_events(
+                event,
+                results=results,
+                options=type(self).EventProcessingOptions(key_by_date=False)
```
dosaboy, line 253: nit: it isn't necessary to use type() here? 
 since this does work either why i wont block on this and can fix in followup patch

hotsos/cli.py
```diff
@@ -303,49 +345,53 @@ def cli(data_root, version, defs_path, templates_path, all_logs, debug,
...
-                if list_plugins:
+                if arguments.list_plugins:
                     sys.stdout.write('\n'.join(plugintools.PLUGINS.keys()))
                     sys.stdout.write('\n')
                     return
 
-                with progress_spinner(not quiet and not debug, drm.name):
+                with progress_spinner(
+                        not arguments.quiet and not arguments.debug, drm.name):
                     plugins = []
```
dosaboy, line 395: This is now unfortunately broken. we were using kwargs that were not explicitly defined i.e. from  (**kwargs) as plugin names to be run and since almost everything is now in kwargs this now blows up when you use any on the command line. We therefore need to find a way to identify plugin names in kwargs - maybe could pull the nested new_init() function out of filter_unexpected_field() and generalise it to use for this purpose also.

dosaboy (changes requested):

hotsos/cli.py
mustafakemalgilor: Good catch, I'll fix it shortly

mustafakemalgilor: Should be good to go.

hotsos/cli.py
```diff
@@ -188,7 +189,52 @@ def progress_spinner(show_spinner, path):
         thread.join()
 
 
-def main():  # pylint: disable=R0915
+def filter_unexpected_fields(cls):
```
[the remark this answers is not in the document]
mustafakemalgilor: This method is now incorporated into the CLIArgs class with a docstring.

hotsos/plugin_extensions/openstack/agent/events.py
mustafakemalgilor: It's not necessary, it's just shorter to type `type(self).` instead of `EventProcessingUtils.`

hotsos/core/host_helpers/network.py
mustafakemalgilor: Should be fixed now

hotsos/client.py
mustafakemalgilor: That was an interesting gotcha, thanks Ed!

hotsos/plugin_extensions/openstack/agent/events.py
mustafakemalgilor: Also, I wanted to avoid using `self.` because it looks like a method call to me, which is not the case. But it might still be confused with a static or class method call, so I doubt the merit of doing that now.

dosaboy: this is why i dont like nested functions and classes ;)

dosaboy: well, that and they are harder to test

dosaboy (approved):
lgtm

PR #954, opened 2024-07-26
Support detect qemu-system-arm64 vms from Nova

dosaboy:
qemu processes launched by Openstack Nova on arm64 machines look different in ps output so we need to take that into account.

### DIFF
--- a/hotsos/core/plugins/openstack/nova.py
+++ b/hotsos/core/plugins/openstack/nova.py
@@ -1,6 +1,7 @@
 import os
 import re
 from functools import cached_property
+from dataclasses import dataclass
 
 from hotsos.core.plugins.kernel.sysfs import CPU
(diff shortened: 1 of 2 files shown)


hotsos/core/plugins/openstack/nova.py
```diff
@@ -24,52 +25,81 @@
...
-                instances[uuid] = guest
+                guest.memory_mbytes = int(ret.group(1))
+
+            instances[uuid] = guest
 
         return instances
 
+    @cached_property
+    def instances(self):
+        return self.get_instances(NovaQemuProcessIdentifierx86()) or \
```
dosaboy, line 100: these dont actually need to be instantiated so i might change that

nicolasbock (approved):
Looks good to me.

PR #964, opened 2024-07-31
Re-enable pylint too-many-ancestors check

dosaboy:


hotsos/core/plugins/openstack/common.py
```diff
@@ -243,14 +243,14 @@ def apache2_allow_encoded_slashes_on(self):
...
+class OpenStackChecks(plugintools.PluginPartBase):
     """ OpenStack checks. """
     plugin_name = "openstack"
     plugin_root_index = 4
 
-    @property
+    @cached_property
     def plugin_runnable(self):
-        return self.openstack_installed
+        return OpenstackBase().openstack_installed
```
brianphaley, line 253: I guess in a perfect world this would just be a function and not have to instantiate the class

brianphaley (approved):

hotsos/core/plugins/openstack/common.py
dosaboy: agreed this is not optimal, i will followup with a patch to improve it

PR #965, opened 2024-08-01
Re-enable pylint too-many-instance-attributes

dosaboy:


hotsos/plugin_extensions/openstack/service_network_checks.py
```diff
@@ -73,9 +74,9 @@ def get_phy_port_health_info(self):
         etc) for any outliers detected.
         """
         port_health_info = {}
-        for project in ['nova', 'neutron', 'octavia']:
-            _project = getattr(self, project)
-            if _project and _project.bind_interfaces:
+        for project in [f.name for f in fields(self.project_helpers)]:
+            _project = getattr(self.project_helpers, project)
+            if _project.bind_interfaces:
```
rmalz-c, line 79: _project can still be invalid object, old check is still necessary even after using fileds

hotsos/plugin_extensions/openstack/service_network_checks.py
```diff
@@ -149,12 +150,12 @@ def _get_router_iface_mtus(self):
     @summary_entry('router-port-mtus', get_min_available_entry_index() + 8)
     def summary_router_port_mtus(self):
         """ Provide a summary of ml2-ovs router port mtus. """
-        project = getattr(self, 'neutron')
-        if not (project and project.bind_interfaces):
+        bind_interfaces = self.project_helpers.neutron.bind_interfaces
```
rmalz-c, line 153: aa, neutron can be "None" 
Quck note, I'm not fan of putting everywhere "null checks" but since it was already there it makes sense to keep it.
eg.
bind_interfaces = self.project_helpers.neutron.bind_interfaces if self.project_helpers.neutron is not None else None

rmalz-c (commented):
couple of nitpicks

hotsos/core/host_helpers/cli.py
```diff
@@ -129,18 +142,21 @@ class FileCmd(CmdBase):
...
-        self.path = self.original_path = os.path.join(HotSOSConfig.data_root,
-                                                      path)
-        self.original_json_decode = json_decode
-        self.original_singleline = singleline
-        self.original_decode_error_handling = decode_error_handling
+        self.original_state = FileCmdState(os.path.join(HotSOSConfig.data_root,
+                                                        path),
+                                           json_decode, singleline,
+                                           decode_error_handling)
+        self.path = str(self.original_state.path)
```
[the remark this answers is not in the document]
dosaboy: ive fixed this by maintaining both current and backup state in the same object

hotsos/plugin_extensions/openstack/service_network_checks.py
dosaboy: you still have to split this over several lines though so i feel like you end up with the same thing?

dosaboy: re: neutron can be None, I don't believe this is the case anymore hence why i removed the check.

hotsos/core/host_helpers/cli.py
```diff
@@ -129,18 +152,16 @@ class FileCmd(CmdBase):
...
                  singleline=False, decode_error_handling=None):
-        self.path = self.original_path = os.path.join(HotSOSConfig.data_root,
-                                                      path)
-        self.original_json_decode = json_decode
-        self.original_singleline = singleline
-        self.original_decode_error_handling = decode_error_handling
+        self.state = FileCmdState(os.path.join(HotSOSConfig.data_root,
+                                               path), json_decode, singleline,
+                                  decode_error_handling)
+        self.reset()
```
rmalz-c, line 158: I think this code became unnecessary too complicated.
Thinking out lout, when you crate object of FileCmdState at the end of object creation you want to have values to be set on both path, json_decode (...etc) and orginal_path, original_json_decode (...etc)
Why not to call FileCmdState.reset in FileCmdState._post_init_?

Next, going to line 158, you are calling self.reset() to copy all attributes from self.state (FileCmdState) to self._attribute_name(FileCmd), is that necessary? You already have self.state.path why also add self.path?

rmalz-c: I'm not saying code won't work, but it's getting a little bit confusing.

hotsos/plugin_extensions/openstack/service_network_checks.py
rmalz-c: neutron, can't be None in current code but if you create object like:
        self.project_helpers = OSTProjectHelpers(NovaBase(), None, OctaviaBase())
it will be.
Resolving this as there might not be a point for writing a checks for code which does not exists

rmalz-c: Hmm it looks like I can't resolve the comments

hotsos/core/host_helpers/cli.py
dosaboy: i have changed the way this works now, and hopefully made it more intuitive. please take a look

rmalz-c (commented):
other than that lgtm

hotsos/core/host_helpers/cli.py
```diff
@@ -28,20 +29,26 @@
...
+
+    Provides a way to save original state and restore to that state.
+    """
+    def __post_init__(self):
         self.hooks = {}
-        self.reset()
+        # make a copy of field original values.
+        for f in fields(self):
+            setattr(self, 'original_' + f.name, f.type(getattr(self, f.name)))
+        super().__init__()
```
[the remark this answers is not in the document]
dosaboy: This is to support the case where CmdBase or an implementation thereof is used in multiple inheritance where one or more class has a __init__ that needs to be called. Currently I have structure things so that this is not necessarily needed but it is good practice to leave it so that in the future if any class used in conjunction with this one did implement __init__ it would get called - although because this is a dataclass it would not be able to pass any args/kwargs through to the parent.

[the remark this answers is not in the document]
dosaboy: ive gone ahead and removed that line. I agree that have it there given that it has no actual use in the current code is not right.

rmalz-c (approved):
lgtm

PR #967, opened 2024-08-07
Re-enable pylint too-many-locals

brianphaley:


brianphaley:
I tested this produced identical output to version 1.18.post71 using a very large sosreport from a compute node. I'm assuming I didn't bork anything else up in the process.

hotsos/core/plugintools.py
```diff
@@ -667,44 +686,25 @@ def _run_plugin_parts(self, global_searcher):
...
-            # We really do want to catch all here since we don't care why
-            # it failed but don't want to fail hard if it does.
-            except Exception as exc:  # pylint: disable=W0718
-                self.failed_parts.append(name)
-                log.exception("part '%s' raised exception: %s", name, exc)
-                output = None
-
-            if output:
-                for key, entry in output.items():
+                for key, entry in inst.output.items():
```
dosaboy, line 689: if we start to see failed_parts because this raises an AttributeError if inst.output is None then we need to either fix the plugin part to return {} instead or go back to checking for None.

This will also change the output in that it will save the output until the failure happens such that you will get partial output. But maybe that is more useful than no part output + failure msg.

dosaboy: lgtm though so im happy to land as is

dosaboy (approved):

PR #969, opened 2024-08-16
Fix module interface definitions

dosaboy:
Removes flake8 error ignores from __init__ files and
add __all__ definitions for module interfaces.

### DIFF
--- a/hotsos/core/issues/__init__.py
+++ b/hotsos/core/issues/__init__.py
@@ -1,5 +1,10 @@
 from .issue_types import *  # noqa: F403,F401
-from .utils import (   # noqa: F403,F401
+from .utils import (
     IssuesManager,
     IssueContext,
 )
+
+__all__ = [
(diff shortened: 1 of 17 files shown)


hotsos/core/issues/__init__.py
```diff
@@ -1,5 +1,10 @@
-from .issue_types import *  # noqa: F403,F401
-from .utils import (   # noqa: F403,F401
+from .issue_types import *  # noqa:F403,F401
```
brianphaley, line 1: Don't think missing space after noqa: matters

brianphaley (approved):

hotsos/core/issues/__init__.py
dosaboy: ack that was unintentional will fix

pponnuvel (approved):
LGTM

PR #972, opened 2024-09-25
Add scenario check for LP 2017748

dosaboy:


nicolasbock (approved):

zhhuabj (approved):
lgtm
I tested it, this is my test result - https://pastebin.canonical.com/p/KCmrsBC4JX/

hotsos/defs/scenarios/openstack/neutron/bugs/lp2017748.yaml
```diff
@@ -0,0 +1,19 @@
+checks:
+  has_2017748:
+    input:
+      path: 'var/log/neutron/neutron-ovn-metadata-agent.log'
+    search:
+      expr: '([\d-]+ [\d:]+.\d{3}) .+neutron.agent.ovn.metadata.agent \[-\] There is no metadata port for network \S+ or it has no MAC or IP addresses configured, tearing the namespace down if needed'
```
pponnuvel, line 6: Is it possible to skip the "date" part? The rest of the strings would uniquely match in order to catch the issue. I know the log lines would have to have the timestamp for "constraints" to work but it's not necessary we regex it, right? Possibly a tiny improvement in compiling the regex is all.

This isn't unique to this scenario. But broadly applicable to most scenarios.

pponnuvel (approved):
LGTM

hotsos/defs/scenarios/openstack/neutron/bugs/lp2017748.yaml
dosaboy: The short answer is no, this is necessary and the reason is because of the use here of "extra" constraints. See https://github.com/canonical/hotsos/blob/21571a334dfde58ea1f79b0f612efe9771d96a4c/hotsos/core/ycheck/engine/properties/search.py#L91 for a more in-depth explanation.

PR #980, opened 2024-10-03
Fix landscape-server version

pponnuvel:


pponnuvel:
My guess wasn't right ;-)

PR #982, opened 2024-10-04
[ceph-osd] improve osdfull reason and suggest workaround

pponnuvel:
There's been evidence that bdev_async_discard was the root cause of this issue. So if anyone encountering this problem should disable it.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/ceph-mon/osd_unusual_raw.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-mon/osd_unusual_raw.yaml
@@ -9,11 +9,17 @@ conclusions:
     raises:
       type: CephOSDWarning
       message: >-
-        Found OSD(s) {bad_osds} with larger raw usage size than the combined
-        data+meta+omap usage. While a certain discrepancy is to be expected due to
-        Ceph's using space not accounted by data+meta+omap columns, these are more
-        than {limit}% and potentially indicate a bug in Ceph. If these OSDs appear
-        full or misbehave, please restart them and possibly file a bug in Ceph tracker.
+        Found OSD(s) {bad_osds} with larger raw usage size than data+meta+omap
+        combined. While a discrepancy is to be expected due to Ceph using space
+        not accounted by data+meta+omap columns, usage is greater than {limit}%
+        and likely indicates high discard ops sent to disk which is often
+        the case for workloads with frequent rewrites.
+
+        If these OSDs appear full or misbehave please restart them.
+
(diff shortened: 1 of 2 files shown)


hotsos/defs/scenarios/storage/ceph/ceph-mon/osd_unusual_raw.yaml
```diff
@@ -13,7 +13,12 @@ conclusions:
         data+meta+omap usage. While a certain discrepancy is to be expected due to
         Ceph's using space not accounted by data+meta+omap columns, these are more
         than {limit}% and potentially indicate a bug in Ceph. If these OSDs appear
-        full or misbehave, please restart them and possibly file a bug in Ceph tracker.
+        full or misbehave, please restart the problematic OSD(s).
+        In case, the problem is more persistent (i.e. OSD restarts do not help),
+        then disable bdev_async_discard option for the OSDs which is known to
+        send high discard ops to disk and wouldn't help for workloads with
+        frequent rewrites.
+
```
dosaboy, line 21: Can I propose the following re-wording:

        Found OSD(s) {bad_osds} with larger raw usage size than data+meta+omap
        combined. While a discrepancy is to be expected due to Ceph
        using space not accounted by data+meta+omap columns, this are greater
        than {limit}% and potentially indicates a bug in Ceph.
        
        If these OSDs appear full or misbehave please restart the problematic OSD(s).

        If the problem persists (i.e. OSD restarts do not help) you should disable
        bdev_async_discard for OSDs known to send high discard ops to disk
        (although perhaps not for workloads with frequent rewrites).

Assuming that it still makes sense.

pponnuvel: Modified again. Please check.

dosaboy: ok looks better, i one final tweak to make:

    Found OSD(s) {bad_osds} with larger raw usage size than data+meta+omap
    combined. While a discrepancy is to be expected due to Ceph using space
    not accounted by data+meta+omap columns, usage is greater than {limit}%
    and likely indicates high discard ops sent to disk which is often
    the case for workloads with frequent rewrites.

    If these OSDs appear full or misbehave please restart them.

    If the problem persists (i.e. OSD restarts do not help) you should disable
    bdev_async_discard for OSDs. For charmed Ceph, this option is controlled
    via the bdev-enable-discard flag which should be set to 'disable'.

Issue #984, opened 2024-10-10
CVE warnings for Juju


pponnuvel:
The following CVEs were recently patched and we'd need to warn about them.

[CVE-2024-7558](https://github.com/juju/juju/security/advisories/GHSA-mh98-763h-m9v4)
[CVE-2024-8037](https://github.com/juju/juju/security/advisories/GHSA-8v4w-f4r9-7h6x)
[CVE-2024-8038](https://github.com/juju/juju/security/advisories/GHSA-xwgj-vpm9-q2rq)

PR #985, opened 2024-10-11
[juju] Add checks for 3 CVEs

pponnuvel:
Closes: #984

### DIFF
--- a/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
+++ b/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
@@ -1,5 +1,10 @@
+vars:
+  msg_common: >-
+    is the running version of Juju on this host which is
+    affected by a known security vulnerability. Please upgrade
+    to the latest version to get the fix.
 checks:
-  has_affected_juju_binary:
+  has_affected_juju_binary_check1:
     binary:
       handler: hotsos.core.plugins.juju.JujuBinaryInterface
       juju:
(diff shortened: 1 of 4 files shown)


hotsos/defs/scenarios/juju/juju_binary_cve.yaml
```diff
@@ -11,15 +16,54 @@ checks:
           max: '3.3.3'
         - min: '3.4.0'
           max: '3.4.1'
+  has_affected_juju_binary_check2:
+    binary:
+      handler: hotsos.core.plugins.juju.JujuBinaryInterface
+      juju:
+        - min: '2.9.0'
+          max: '2.9.50'
```
dosaboy, line 24: @xmkg with the logic you added can the min be omitted here i.e. so that we only define the 'max'?

xmkg: Yes, you should be able to do:

```
      juju:
        - ge: '2.9.0' # or `min:`
```

https://github.com/canonical/hotsos/blob/3f4a26d5d60d405f4f99bd94a27b518fb0264385/hotsos/defs/scenarios/openstack/keystone/bugs/lp1896125.yaml#L6

pponnuvel: I don't know exactly how `lt` is implemented. But it doesn't seem to work quite how I'd want.

For example, if I use
```
 -lt : 2.8.9
```
without `min`, it produces the warning when it shouldn't as the CVE only existed from 2.9.0 onwards.

xmkg: It's less than '<'. So if the cve is 2.9.0 and onwards and still unfixed, you can change it to greater equal (ge), and add a less than 3 (assuming juju 3 is unaffected).

dosaboy: fwiw there is some documentation here but now that I look at it i realise it could do with an update - https://hotsos.readthedocs.io/en/latest/contrib/language_ref/property_ref/requirement_types.html#apt

pponnuvel: Sorry I was a bit clumsy in my previous comment. I *did not* use `-lt: 2.8.9`, rather I expected the following to change on top of this PR to work:
```
diff --git a/hotsos/defs/scenarios/juju/juju_binary_cve.yaml b/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
index b6407063..d6624425 100644
--- a/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
+++ b/hotsos/defs/scenarios/juju/juju_binary_cve.yaml
@@ -20,8 +20,7 @@ checks:
     binary:
       handler: hotsos.core.plugins.juju.JujuBinaryInterface
       juju:
-        - min: '2.9.0'
-          max: '2.9.50'
+        - lt: '2.9.50'
         - min: '3.0.0'
           max: '3.1.9'
         - min: '3.2.0'
diff --git a/hotsos/defs/tests/scenarios/juju/juju_binary_cve.yaml b/hotsos/defs/tests/scenarios/juju/juju_binary_cve.yaml
index 5a388e83..e370effe 100644
--- a/hotsos/defs/tests/scenarios/juju/juju_binary_cve.yaml
+++ b/hotsos/defs/tests/scenarios/juju/juju_binary_cve.yaml
@@ -2,21 +2,5 @@ mock:
   patch:
     hotsos.core.plugins.juju.resources.JujuBinaryInterface.get_version:
       kwargs:
-        return_value: 3.4.1
+        return_value: 2.8.9
 raised-bugs:
-  https://www.cve.org/CVERecord?id=CVE-2024-3250: >-
-    3.4.1 is the running version of Juju on this host which is
-    affected by a known security vulnerability. Please upgrade
-    to the latest version to get the fix.
-  https://www.cve.org/CVERecord?id=CVE-2024-7558: >-
-    3.4.1 is the running version of Juju on this host which is
-    affected by a known security vulnerability. Please upgrade
-    to the latest version to get the fix.
-  https://www.cve.org/CVERecord?id=CVE-2024-8037: >-
-    3.4.1 is the running version of Juju on this host which is
-    affected by a known security vulnerability. Please upgrade
-    to the latest version to get the fix.
-  https://www.cve.org/CVERecord?id=CVE-2024-8038: >-
-    3.4.1 is the running version of Juju on this host which is
-    affected by a known security vulnerability. Please upgrade
-    to the latest version to get the fix.
```

But it doesn't since we needed both the min & max versions (between which the CVEs exist). So I guess no changes are needed here.

dosaboy: ack thanks Pon, i'm good to land as-is, we can always optimise as a separate patch if needs be

PR #996, opened 2025-01-07
Add Sunbeam support to OpenStack plugin

dosaboy:
Also:
 * fixes Juju units summary to include charm revision where charm has no repo-info file.
 * adds snap channel to summary output
 * shows pods and containers from microk8s

Resolves: #995


tests/unit/test_openstack.py
```diff
@@ -232,6 +232,22 @@
...
 
+SUNBEAM_SNAP_LIST = """
+Name                  Version                Rev    Tracking            Publisher    Notes
+bare                  1.0                    5      latest/stable       canonical**  base
+core20                20240911               2434   latest/stable       canonical**  base
+core22                20241119               1722   latest/stable       canonical**  base
+core24                20240920               609    latest/stable       canonical**  base
+juju                  3.5.5                  29058  3.5/stable          canonical**  -
+lxd                   5.0.4-497fe1e          31333  5.0/stable/…        canonical**  -
+microceph             18.2.4+snapa97ae91192  1234   reef/edge           canonical**  held
```
pponnuvel, line 243: Do we need a followup issue to track these "edges" and convert them to stable at some point?

pponnuvel (approved):

tests/unit/test_openstack.py
dosaboy: sunbeam 2024.1 is not released yet and so has not 2024/1/stable channel. I think adding a scenario to detect usage of the edge snap makes sense but we need wait until a stable snap is available.

dosaboy:
I'll go ahead and merge. We can always add more later.

Issue #1001, opened 2025-01-11
Support checking k8s pod logs


dosaboy:
Sosreport contains kubectl pod logs (incl. microk8s.kubectl) so we can add support to be able to query them for errors.

PR #1002, opened 2025-01-11
Add kubectl logs cli command

dosaboy:
Supports native, microk8s and k8s variants.

Resolves: #1001

### DIFF
--- a/hotsos/core/host_helpers/cli/catalog.py
+++ b/hotsos/core/host_helpers/cli/catalog.py
@@ -372,6 +372,26 @@ class SunbeamOVSOFCtlBinCmd(OVSOFCtlBinCmd):
     BIN_CMD = 'openstack-hypervisor.ovs-ofctl'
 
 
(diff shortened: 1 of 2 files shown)


hotsos/core/host_helpers/cli/catalog.py
```diff
@@ -713,6 +733,18 @@ def __init__(self):
             'ip_link':
                 [BinCmd('ip -s -d link'),
                  FileCmd('sos_commands/networking/ip_-s_-d_link')],
+            'kubectl_logs':
+                [KubectlLogsBinCmd('kubectl'),
+                 # Canonical K8s
+                 KubectlLogsBinCmd('microk8s.kubectl'),
+                 # MicroK8s
```
brianphaley, line 740: My only comment is that the comments don't seem to match the commands?

dosaboy: Oops! Thanks Brian will fix.

PR #1003, opened 2025-01-11
Fall back to dmesg if kern.log not found

dosaboy:
On some hosts kern.log may be missing so we fall back to dmesg for calltrace checking if this is the case.


dosaboy:
need to use journalctl -k not var/log/dmesg

PR #1004, opened 2025-01-11
Use journald for kernlogs if kern.log not found

dosaboy:
On some hosts or sosreports kern.log may be missing so we fall back to journald for calltrace checking if this is the case.


brianphaley (approved):

dosaboy:
> Doh, missed test failure

Interesting it didn't fail in py310

brianphaley (approved):

PR #1006, opened 2025-01-20
Compress hotsos log using gzip

dosaboy:
The hotsos.log file can get very large so best to store compressed.


hotsos/client.py
```diff
@@ -184,12 +185,27 @@ def _save_to_file(path, content):
...
+    def compress(path):
+        """
+        Compress path to path.gz and delete path.
+        """
+        cpath = path + '.gz'
+        with open(path, 'rb') as fd_in:
+            with gzip.open(cpath, 'wb') as fd_out:
+                shutil.copyfileobj(fd_in, fd_out)
+
+        os.remove(path)
```
pponnuvel, line 198: Is it still OK to delete even if compression failed for some reason?

dosaboy: that would throw an exception i presume so it would not get that far

pponnuvel (approved):
LGTM

PR #1010, opened 2025-03-03
Improve format of netlink socks check output

dosaboy:


tests/unit/test_kernel.py
```diff
@@ -686,6 +686,17 @@ def test_netlink_parse_with_drops(self):
...
+        {'proc/net/netlink': PROC_NETLINK,
+         'sos_commands/process/lsof_M_-n_-l_-c': LSOF_MNLC}
+    )
+    def test_netlink_parse_with_drops_str(self):
+        uut = NetLink()
+        awd = uut.all_with_drops_str
+        self.assertEqual(awd,
+                         "inode=34703 "
+                         "procs=[ovs-vswit/13936, qemu-syst/20986]")
+
```
brianphaley, line 699: Think there is an extra blank line here as well.

brianphaley: Doh, new class below so ignore that comment

brianphaley (approved):

PR #1017, opened 2025-03-25
[ceph] Skip suboptimal PG count warning on low utilisation

### LINKED ISSUE #1016: Rework suboptimal PG count warning
We have `ceph-mon/pg_imbalance.yaml` that warns if the PG count is outside the range of 50-200 per OSD.

But this is going to be an issue for almost all newly deployed Ceph clusters (especially at handover stage) which won't have much data.

Since the PG autoscaler is enabled by default in recent Ceph versions, this is less likely to be an issue for any cluster.

We can instead rework this so that this is only triggered if the total utilisation is above a minimum % of the cluster capacity.

pponnuvel:
When the cluster is empty or utilised at a very low level, the PG autoscaler won't increase the PGs/OSD to the optimal level - it'll be done as the cluster hosts more data. In such cases, the PGs/OSD count will be lower than what we consider as optimal range (50 - 200). In such cases, this warning has no benefit and becomes a false positive. So we only report this if the cluster usage is 10% or more.

Fixes #1016.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -233,10 +233,12 @@ class CephCluster():  # pylint: disable=too-many-public-methods
     """
     OSD_META_LIMIT_PERCENT = 5
     OSD_PG_MAX_LIMIT = 500
+    # min cluster utilisation required to trigger this warning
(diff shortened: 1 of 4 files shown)


dosaboy (approved):
lgtm

PR #1025, opened 2025-05-29
Fix systemd support for snapped services

dosaboy:
Services running as snap.<name> were not have their state represented correctly in the SystemdHelper
summary.


pponnuvel (approved):
LGTM

PR #1040, opened 2025-10-23
SET-1661 Add check for ceph heavy delete workload bug

dosaboy:


pponnuvel:
Please add these two tracker to the message:

https://tracker.ceph.com/issues/58274
https://tracker.ceph.com/issues/55324

PR #1041, opened 2025-10-29
Check recently reported laggy PGs

pponnuvel:


hotsos/defs/scenarios/storage/ceph/ceph-mon/recent_laggy_pgs.yaml
```diff
@@ -0,0 +1,19 @@
+checks:
+  recent_laggy_pgs:
+    input:
+      path: ['var/log/ceph/ceph*.log', 'var/snap/microceph/common/logs/ceph*.log']
+    search:
+      expr: '^([\d-]+)[T ]([\d:]+)\S+ .+laggy, .+'
+      constraints:
+        # A brief appearance isn't an issue (due to OSD overload or network blip).
+        min-results: 20
```
hillpd: A single log line is an aggregate and could represent dozens of PGs. I would suggest calling out any and all laggy PGs.

pponnuvel:
Thanks for review @hillpd. Please take another look.

PR #1042, opened 2025-12-01
Add http return code summary for openstack apis

dosaboy:
Expand APIs we check to include cinder, keystone and octavia and add an event def for http return codes to complement the existig http requests summary.


wilkmar (approved):
LGTM

PR #1043, opened 2025-12-02
Extend api events to include sunbeam logs

dosaboy:
We can check sunbeam logs by looking at the output
of kubectl logs for openstack pods/containers.

Also bumps propertree to get fix for branch
precedence.


hotsos/defs/events/openstack/http-requests.yaml
```diff
@@ -3,30 +3,75 @@
...
+
+keystone-sunbeam:
+  input:
+    command: kubectl_logs
+    options:
+      kwargs:
+        namespace: openstack
+        opt: keystone-0
+        subopts: -c keystone
+  expr: '.+(\d{2}/\w{3,5}/\d{4}).+ "(GET|POST|PUT|DELETE)'
```
wilkmar: Could we avoid duplicating the expression for the regular and the sunbeam-specific check? It looks like most of the OpenStack components here have the same expression for the regular and the sunbeam check.
Also, can we get the logs for sunbeam from the sos report (where you can't run the kubectl command)?
I have a single-node deployment and all pod logs are on the node in the following locations:
- `/var/log/containers/` - symlinks to the locations below:
- `/var/log/pods/<pod>/<container>/`
Could you check in your test env and in the example sos report if you have those logs? It would be great to have unified checks, as much as possible, for the regular and sunbeam

hotsos/defs/events/openstack/http-requests.yaml
```diff
@@ -3,33 +3,78 @@
...
+
+keystone-sunbeam:
+  input:
+    command: kubectl_logs
+    options:
+      kwargs:
+        namespace: openstack
+        opt: keystone-0
+        subopts: -c keystone
+  expr: '@hotsos.core.plugins.openstack.HTTPRequestExprs.sunbeam_nova'
```
brianphaley: s/sunbeam_keystone

hotsos/defs/events/openstack/http-requests.yaml
```diff
@@ -3,33 +3,78 @@
...
+
 keystone:
   input:
     path: 'var/log/apache2/keystone_access.log'
     options:
       disable-all-logs: true
-  expr: '.+(\d{2}/\w{3,5}/\d{4}).+ "(GET|POST|PUT|DELETE)'
+  expr: '@hotsos.core.plugins.openstack.HTTPRequestExprs.apache_common'
+
+
```
brianphaley: nit: extra blank line

hotsos/core/host_helpers/cli/commands/kubectl.py
```diff
@@ -68,11 +80,12 @@ def __init__(self):
 
             paths.extend([(f'sos_commands/kubernetes/pods/{cmd}'
                            f'_--kubeconfig_{conf}'
-                           '_--namespace_{namespace}_logs_{opt}') for conf in
+                           '_--namespace_{namespace}_logs_{opt}'
```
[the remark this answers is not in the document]
dosaboy: so i tested against a sunbeam but not charmed openstack although i did test against a sosreport from charmed openstack. having said that i have not tested against a sosreport from all apis, only nova. let me look into this.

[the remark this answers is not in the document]
dosaboy: ok so, reason it was not working is because noble sosreport is not collecting logs from the openstack namespace which it used to do so there is a bug in sosreport somewhere.

There is also /var/log/pods but this is (a) only for local pods and (b) does not use standard logrotate format so to make that work we need to support different logrotate formats in http://github.com/dosaboy/searchkit.

So i propose we leave as is, investigate why sosreport is behaving like this and extend searchkit logrotate tyes.

PR #1044, opened 2025-12-08
docs: Fix incorrect description of example

raineszm:
Just noticed this minor inconsistency while reading the docs.

Docs described an example as checking an apt package version, when it was instead checking the status of a systemd service.


nicolasbock (approved):
LGTM

Issue #1046, opened 2026-01-07
MicroOVN logs are not inspected


lathiat:
Currently we have various rules checking OVN and OVS logs, using a specific path. e.g. hotsos/defs/events/openvswitch/ovs/ovs-vswitchd.yaml takes input from var/log/openvswitch/ovs-vswitchd.log

However microovn doesn't log to this location, and the sosreport microovn plugin currently doesn't seem to collect an equivalent path only sos_commands/microovn/journalctl_--no-pager_--unit_snap.microovn

We should ensure parity between ovn and microovn

PR #1047, opened 2026-01-07
SET-766 Add support for multiple log locations

dosaboy:
Openvswitch/OVN can be installed from deb or snap which have different log path locations. To support this without having to duplicate multiple paths all over the yaml files we now have a PathFinder class that can be used to dynamically discover file paths. This is supported by the input ycheck property by providing '<pluginname>:<subpath>' as a value.

Currently implemented for the openvswitch plugin but easily extensible to others.

Resolves: #1046

### DIFF
--- a/hotsos/core/utils.py
+++ b/hotsos/core/utils.py
@@ -1,3 +1,6 @@
+import abc
+import glob
+import os
 import tempfile
 from collections import namedtuple
 from operator import attrgetter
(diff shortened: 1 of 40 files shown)


hotsos/core/utils.py
```diff
@@ -105,3 +107,20 @@ def sort_suffixed_integers(values, reverse=False):
...
     reals.sort(key=attrgetter('actual'), reverse=reverse)
     return [t.raw for t in reals]
+
+
+class PathFinderBase:
+    """
+    Discovers correct path to a file. This is used to support file locations
+    that differ depending on their install method e.g. snap vs. deb package.
+    """
+    PATHS = []
```
wilkmar: not a big deal, but the `PATHS` here could be an abstract property (ie `@property` + `@abstractmethod`, from the  `abc` module).

wilkmar (commented):
except for one minor comment, it LGTM.

For the microceph scenarios, the logs are also available via journal:
`sudo journalctl -xe -u snap.microceph.*`

The other example would be MAAS which can be installed from deb or snap. However, the newer versions of MAAS completely switched from file logging to journal logging.

dosaboy:
> except for one minor comment, it LGTM. Well, maybe a doc update for the `Input` property (that `path` accepts a component followed by the `:`), would be helpful.
> 
> For the microceph scenarios, the logs are also available via journal: `sudo journalctl -xe -u snap.microceph.*`
> 
> The other example would be MAAS which can be installed from deb or snap. However, the newer versions of MAAS completely switched from file logging to journal logging.

yep ill make the doc update for sure. so for the journal side of things i did give it some thought but could not think of a clean way to add that to the PathFinder. It is however possible to add it manually to input using the ```commands:``` stanza. I will have another think about how to do that though.

hotsos/core/utils.py
dosaboy: done.

dosaboy:
Im going to defer implementing the journalctl output to a separate patch since it is going to be non-trivial.

PR #1049, opened 2026-01-07
Extend CPU scaling governor warning to microceph

pponnuvel:


hotsos/defs/scenarios/storage/ceph/ceph-osd/system_cpufreq_mode.yaml
```diff
@@ -11,6 +11,12 @@ vars:
     order for changes to persist.
   scaling_governor: '@hotsos.core.plugins.kernel.sysfs.CPU.cpufreq_scaling_governor_all'
 checks:
+  ceph_osd_installed:
+    apt:
+      ceph-osd:
```
dosaboy: dont need the : at the end of the package name. It only needs it if you are going to specify version constraints.

hotsos/defs/scenarios/storage/ceph/ceph-osd/ceph-osd.yaml
```diff
@@ -1,6 +1,8 @@
...
-    # but it will typically be disabled if not in used so we use that as a
-    # means of knowing what type of node it is.
-    ceph-osd: enabled
+  or:
+    systemd:
+      # both mon and osd nodes will have a ceph-mon package and service installed
+      # but it will typically be disabled if not in used so we use that as a
+      # means of knowing what type of node it is.
+      ceph-osd: enabled
+    pebble: ceph-osd
```
dosaboy: i think you also need to add snap.microceph.osd to the list of systemd services?

hotsos/defs/scenarios/storage/ceph/ceph-mon/ceph-mon.yaml
```diff
@@ -13,3 +13,5 @@ requires:
     systemd:
       ceph-mon: enabled
     pebble: ceph-mon
+    systemd:
```
dosaboy: this should be:

```
# This requirement is used to define whether or not checks contained within this
# path are allowed to run. Both mon and osd nodes will have a ceph-mon package
# and systemd service installed so their existance alone is not enough to
# determine what kind of node this is. Since a ceph cluster will typically have
# more than one mon we also do not necessarily want to skip checks that can run
# if *any* mon is running. To achieve this we make this requirement sufficiently
# permissive such that the local ceph-mon service need not necessarily be
# active - checks themselves can apply more restrictive pre-requisites
# if necessary.
requires:
  # ensures that a ceph-mon unit is enabled but does not assert that it is active
  or:
    - systemd:
        ceph-mon: enabled
    - systemd:
        snap.microceph.mon: enabled
    - pebble: ceph-mon
```

hotsos/defs/scenarios/storage/ceph/ceph-osd/ceph-osd.yaml
```diff
@@ -1,6 +1,9 @@
+# Both mon and osd nodes will have a ceph-mon package and service installed
+# but it will typically be disabled if not in-use so we use that as a
+# means of knowing what type of node it is.
 requires:
-  systemd:
-    # both mon and osd nodes will have a ceph-mon package and service installed
-    # but it will typically be disabled if not in used so we use that as a
-    # means of knowing what type of node it is.
-    ceph-osd: enabled
+  or:
```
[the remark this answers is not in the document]
pponnuvel: It works but yamllint fails due to duplicate keys (`systemd`).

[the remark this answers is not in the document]
pponnuvel: After the latest push, it fails for functional test. Leave this to me, I'll come back.

pponnuvel:
Thanks Ed! Works for both ceph & microceph now. Please take a look.

dosaboy (approved):

PR #1055, opened 2026-01-15
SET-1834 Fix ceph binary command json decode

dosaboy:
Some ceph commands append their json output with a string that needs to be removed before it can be decoded. We already did this for sosreport commands but now have it for binary/host commands.

### DIFF
--- a/hotsos/core/host_helpers/cli/commands/ceph.py
+++ b/hotsos/core/host_helpers/cli/commands/ceph.py
@@ -9,20 +9,43 @@ from hotsos.core.host_helpers.exceptions import SourceNotFound
 CephAliases = ['microceph.']
 
 
+class CephJSONBinCmd(BinCmd):
+    """
+    Some ceph commands that use --format json have some extra text added to the
+    end of the output which causes it to be invalid json
+    so we have to strip that final line before decoding the contents.
+    """
+    def __init__(self, *args, last_line_filter=None, **kwargs):
+        super().__init__(*args, json_decode=True, **kwargs)
+        self.last_line_filter = last_line_filter
+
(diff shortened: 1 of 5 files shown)


hotsos/core/host_helpers/cli/commands/ceph.py
```diff
@@ -9,6 +9,36 @@
...
+        """
+        if output:
+            if self.first_line_filter:
+                line_filter = self.first_line_filter
+            else:
+                line_filter = self.last_line_filter
+
+            output = output.strip()
+            _split = output.splitlines()
+            if line_filter and _split[-1].startswith(self.last_line_filter):
```
pponnuvel: This works if `last_line_filter` is given. How is this supposed to handle if `first_line_filter` is given?

hotsos/core/host_helpers/cli/commands/ceph.py
```diff
@@ -9,6 +9,36 @@
...
+        Remove request line(s) before decoding.
+        """
+        if output:
+            if self.first_line_filter:
+                line_filter = self.first_line_filter
+            else:
+                line_filter = self.last_line_filter
+
+            output = output.strip()
+            _split = output.splitlines()
```
pponnuvel: Do you want to handle empty file (in case `_split` is empty at this point, there'll be an exception below).

hotsos/core/host_helpers/cli/commands/ceph.py
dosaboy: actually i did exactly that originally bu then reverted to the way the file version works. nothing is actually using the first line strip so i think ill just remove it because it doesnt make sense to have it there.

PR #1057, opened 2026-01-18
[storage] Add smartctl analysis

### LINKED ISSUE #125: Add smartctl analysis
Sosreports contain smartcl status information which we could and should check for hardware errors and report if any found.

pponnuvel:
Check disk errors and error counters reported by smartctl.

Fixes #125.

### DIFF
--- a/hotsos/core/plugins/storage/smartctl.py
+++ b/hotsos/core/plugins/storage/smartctl.py
@@ -1,3 +1,8 @@
+import os
+
+from hotsos.core.config import HotSOSConfig
+from hotsos.core.host_helpers import APTPackageHelper
+from hotsos.core.log import log
 from hotsos.core.plugins.storage import StorageBase
 
 
@@ -5,6 +10,23 @@ class SmartctlChecks(StorageBase):
     """Base class for smartctl storage checks."""
     @classmethod
     def is_runnable(cls):
-        # Implement logic to determine if smartctl checks should run.
-        # For now, always returns True.
-        return True
+        """
+        Determine whether smartctl checks should run.
+        @return: True if smartctl data is available, False otherwise
+        """
+
+        # Check if smartmontools package is installed
+        apt = APTPackageHelper(core_pkgs=['smartmontools'])
+        if apt.core:
+            log.debug("smartmontools package installed")
+            return True
+
+        # Check if smartctl output exists in sosreport
+        sos_commands = os.path.join(HotSOSConfig.data_root, 'sos_commands')
+        for path in [os.path.join(sos_commands, d) for d in ('nvme', 'ata')]:
+            if os.path.isdir(path) and os.listdir(path):
+                log.debug("smartctl data found in '%s'", path)
+                return True
+
+        log.debug("No smartctl data found")
+        return False
--- a/hotsos/plugin_extensions/storage/smartctl_summary.py
+++ b/hotsos/plugin_extensions/storage/smartctl_summary.py
@@ -1,6 +1,4 @@
-"""
-Plugin for analyzing disk health using smartctl output.
-"""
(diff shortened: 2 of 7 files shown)


hotsos/plugin_extensions/storage/smartctl_summary.py
```diff
@@ -0,0 +1,108 @@
...
+
+from hotsos.core.config import HotSOSConfig
+from hotsos.core.plugintools import get_min_available_entry_index
+from hotsos.core.plugintools import summary_entry
+from hotsos.core.ycheck.scenarios import YScenarioChecker
+from hotsos.core.log import log
+from hotsos.core.ycheck.common import GlobalSearcher
+
+
+class SmartctlSummary(YScenarioChecker):
```
dosaboy: YScenarioChecker is not intended to be used for anything other than processing all scenarios. Instead you need to use an implementation of hotsos.core.plugins.storage.StorageBase

hotsos/plugin_extensions/storage/smartctl_summary.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        results = self._search_sosreport(smartctl_dir)
+        if not results:
+            log.debug('No smartctl output found in sosreport')
+            return None
+
+        # Only include counters that are meaningful for RAW_VALUE > 0
+        # Exclude counters like Seek_Error_Rate, Raw_Read_Error_Rate, and
+        # Hardware_ECC_Recovered, which are not reliable for error detection
+        error_counters = [
```
hillpd: How was this list of SMART attributes selected? Some of these counters (for example, `Reallocated_Sector_Ct` and `UDMA_CRC_Error_Count`) are generally only meaningful when evaluated over time. A non-zero value in a single snapshot does not necessarily indicate an active problem. 

Additionally, `Reported_Uncorrect` is notoriously vendor specific in how it is implemented. It typically needs to be correlated with other SMART attributes  to determine whether it reflects an actual media failure or a transient link issue.

hillpd (changes requested):
Since this is a first-pass implementation and we only have a single data point, I would recommend being more conservative about which SMART counters we call out as actionable. I would limit guidance to stronger indicators such as SMART overall status of `FAILED`, `Current_Pending_Sector`, or `Offline_Uncorrectable` errors and treat other non-zero counters as just informational.

hotsos/core/plugins/storage/smartctl.py
```diff
@@ -0,0 +1,10 @@
+from hotsos.core.plugins.storage import StorageBase
+
+
+class SmartctlChecks(StorageBase):
+    """Base class for smartctl storage checks."""
+    @classmethod
+    def is_runnable(cls):
+        # Implement logic to determine if smartctl checks should run.
+        # For now, always returns True.
+        return True
```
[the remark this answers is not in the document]
pponnuvel: ack, done

hotsos/plugin_extensions/storage/smartctl_summary.py
```diff
@@ -0,0 +1,102 @@
...
+from hotsos.core.plugintools import get_min_available_entry_index
+from hotsos.core.plugintools import summary_entry
+
+
+class SmartctlSummary(SmartctlChecks):
+    """Analyze disk health using smartctl output."""
+
+    summary_part_index = get_min_available_entry_index()
+
+    @summary_entry('disk_health', index=get_min_available_entry_index())
```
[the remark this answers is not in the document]
pponnuvel: Removed

hotsos/plugin_extensions/storage/smartctl_summary.py
pponnuvel: I've now split them into 2 categories as you suggested @hillpd

PR #1059, opened 2026-01-20
[ceph] exclude non-bucket items leaf nodes (e.g. OSDs)

pponnuvel:


pponnuvel:
@dosaboy Sunbeamed it - works fine.

PR #1074, opened 2026-04-14
Skip install check in k8s cpufreq scenario

dosaboy:
The scenario won't be run if k8s is not installed so no need to check in the scenario itself. The install check was also not covering canonical k8s.

Closes: SET-2054


pponnuvel (approved):
LGTM

PR #1077, opened 2026-04-29
Fix check for LP#1934937

dosaboy:
Package name needs to be python3-oslo.messaging not oslo.messaging. Added a new test to cover this.


wilkmar (approved):
LGTM

brianphaley (approved):

PR #1080, opened 2026-05-09
[ceph] improve 'overlapping root' detection

pponnuvel:
The ceph-mgr log may not always contain the relevant errors or may have been rotated out. crush tree provides the more accurate view of the 'overlapping roots' issue as one or more OSDs being part of different trees can be seen directly.

Fixes #SET-2289.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -195,6 +195,21 @@ class CephCrushMap():
 
         return None
 
+    @staticmethod
+    def collect_osd_classes(node_id, nodes):
+        """Recursively collect device classes of all OSDs under a node."""
+        node = nodes.get(node_id)
+        if node is None:
+            return set()
+        if node.get('type') == 'osd':
+            dc = node.get('device_class')
+            return {dc} if dc else set()
+        classes = set()
+        for child_id in node.get('children', []):
+            classes.update(
+                CephCrushMap.collect_osd_classes(child_id, nodes))
+        return classes
+
     @cached_property
     def crush_tree_has_overlapping_roots(self):
         """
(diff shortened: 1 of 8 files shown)


pponnuvel:
json formatted shadow-tree isn't available in sosreport, so proposed it too: https://github.com/sosreport/sos/pull/4323

hotsos/defs/scenarios/storage/ceph/ceph-mgr/autoscaler_overlap_roots.yaml
```diff
@@ -1,21 +1,38 @@
+vars:
+  overlap_common_msg: >-
+    This will prevent PG autoscaler from scaling those pools. This happens
```
dosaboy, line 3: I think this message needs context for what "this" is.

pponnuvel: It's not printed as such - it's part of a shared message and there's more info prefixed before this msg which should give the context.

dosaboy: ah my bad sorry

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -195,6 +195,45 @@ def crushmap_equal_buckets_pretty(self):
...
+        A non-shadow root has overlapping roots when it contains OSDs from
+        multiple device classes.
+        """
+        crush_tree = CLIHelper().ceph_osd_crush_tree_json_decoded()
+        if not crush_tree:
+            return False
+
+        nodes = {n['id']: n for n in crush_tree.get('nodes', [])}
+
+        def collect_osd_classes(node_id):
```
dosaboy: i would really prefer we avoid using nested method, partly because it makes the code harder to review but also because you cannot write a unit test the nested method.

dosaboy (commented):
mostly good but one small thing.

hotsos/core/plugins/storage/ceph/cluster.py
pponnuvel: ack, de-nested.

PR #1085, opened 2026-06-01
Fix check for LP#2071780

ioanalazea:
Added the specific rev number.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/common/bugs/lp2071780.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/common/bugs/lp2071780.yaml
@@ -23,7 +23,7 @@ conclusions:
         This host appears to be affected by LP bug 2071780 where ceph-fs charm
         can restart ceph-mds too often on config-changed hooks. Logs show
         {num_restarts} ceph-mds starts within one hour, which can impact
-        CephFS clients. Consider moving to quincy/stable, squid/stable,
-        or newer.
+        CephFS clients. Upgrade the ceph-fs charm to revision >=293 (quincy/edge)
+        or revision >=200 (squid/stable).
       format-dict:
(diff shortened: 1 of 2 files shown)


hotsos/defs/scenarios/storage/ceph/common/bugs/lp2071780.yaml
```diff
@@ -23,7 +23,7 @@ conclusions:
         This host appears to be affected by LP bug 2071780 where ceph-fs charm
         can restart ceph-mds too often on config-changed hooks. Logs show
         {num_restarts} ceph-mds starts within one hour, which can impact
-        CephFS clients. Consider moving to quincy/stable, squid/stable,
```
dosaboy, line 26: i think what you had originally is better since when doing an upgrade you use the channel not the revision.

pponnuvel: I suggested to use specific revisions because one could be on a stable but on an older revision preceding this fix. Also the latest quincy/stable hasn't got this fix yet.

dosaboy: can we reword it to something like:

```
This host appears to be affected by LP bug 2071780 where ceph-fs charm
can restart ceph-mds too often on config-changed hooks. Logs show
{num_restarts} ceph-mds starts within one hour, which can impact
CephFS clients. Upgrade the ceph-fs charm to revision >=293 (quincy/stable)
or revision >=200 (squid/stable).
```

ioanalazea: Done, though I have replaced with `quincy/edge` since stable doesn't have the fix yet.

pponnuvel (approved):

PR #1086, opened 2026-06-10
[ceph] support for ps and local-osds in microceph

pponnuvel:
Fixes #SET-2197.

### DIFF
--- a/hotsos/core/host_helpers/systemd.py
+++ b/hotsos/core/host_helpers/systemd.py
@@ -352,6 +352,14 @@ class SystemdHelper(ServiceManagerBase):
             path = os.path.join(HotSOSConfig.data_root,
                                 'sys/fs/cgroup/system.slice')
 
+        if not os.path.exists(path):
+            # No cgroup data at all e.g. a sosreport from a snap-based
+            # deployment. Fall back to scanning ps directly.
+            log.warning("could not find cgroup system.slice on this host "
+                        "so falling back to looking up service '%s' "
+                        "processes in ps", path)
+            return self._ps_fallback_for_services()
+
         for svc in self.services:
             for svc in self.get_services_expanded(svc):
                 _path = os.path.join(path, f"{svc}.service",
@@ -371,6 +379,28 @@ class SystemdHelper(ServiceManagerBase):
                             if re.match(rf'^\S+\s+{int(pid)}\s+', line):
                                 ps_filtered.append(line)
                                 break
+
+        return ps_filtered
+
+    def _ps_fallback_for_services(self):
+        """
+        Fallback when cgroup.procs is not available (e.g. snap services).
+        Scans ps output directly for commands matching any of the
+        registered service expressions.
+
+        This is used when the entire cgroup directory is absent, so we
+        match using all service expressions because the binary names in
+        ps may differ from the systemd unit names (e.g.
+        snap.microceph.osd.daemon service runs /snap/.../bin/ceph-osd).
+
+        @return: list of matching ps lines.
+        """
+        ps_filtered = []
+        for line in self._ps:
+            for expr in self._service_exprs:
+                if self.get_cmd_from_ps_line(line, expr):
+                    ps_filtered.append(line)
+                    break
 
         return ps_filtered
 
--- a/hotsos/core/plugins/storage/ceph/common.py
+++ b/hotsos/core/plugins/storage/ceph/common.py
@@ -1,4 +1,6 @@
 import abc
+import glob
+import json
 import os
 import re
 import sys
@@ -305,6 +307,18 @@ class CephChecks(StorageBase):
         """
         Returns a list of CephOSD objects for osds found on the local host.
         """
+        osds = self._local_osds_from_ceph_volume()
+        if not osds:
+            osds = self._local_osds_from_microceph()
+
+        return osds
+
+    @staticmethod
+    def _local_osds_from_ceph_volume():
+        """
+        Returns a list of CephOSD objects parsed from ceph-volume lvm list
+        output.
+        """
         osds = []
 
         s = FileSearcher()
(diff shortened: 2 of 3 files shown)


hotsos/core/host_helpers/systemd.py
```diff
@@ -351,6 +351,11 @@ def _service_filtered_ps(self):
             path = os.path.join(HotSOSConfig.data_root,
                                 'sys/fs/cgroup/system.slice')
 
+        if not os.path.exists(path):
```
dosaboy, line 355: please add a debug log here to say that using fallback because no cgroup. this will make debugging easier.

pponnuvel: done

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -214,6 +216,14 @@ def is_runnable(cls):
         if ceph.apt.core or ceph.snaps.core:
             return True
 
+        # Fallback: detect microceph from its config or data files when
+        # snap list is unavailable (e.g. minimal sosreport).
```
dosaboy, line 222: isn't snap list available in the most minimal sosreports? also if the snap plugin doesnt run is /var/snap going to be included?

pponnuvel: I thought it won't be there if you do somehing like `-o ceph`. I'll check.

pponnuvel: Confirmed. If you specify a single plugin such as `sos report -o ceph_mon` then `snap` commands aren't collected but `/var/snap/microceph/` is still collected.

dosaboy: bah that's annoying, ok fair enough to do this then

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -216,6 +218,14 @@ def is_runnable(cls):
         if ceph.apt.core or ceph.snaps.core:
             return True
 
+        # Fallback: detect microceph from its config or data files when
```
dosaboy: I understand we are doing this for the situation where a sosreport was generated for microceph hosts but without including apt or snap info but it is also unlikely that the plugin with function correctly at all without that data (and the other system data that would normally be expected to exist). I suppose we can try it out but just be aware.

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -332,6 +354,67 @@ def local_osds(self):
 
         return osds
 
+    @staticmethod
+    def _local_osds_from_microceph():
```
dosaboy, line 350: Is this really the way we want to go i.e. relying on var/snap paths to exist in the sosreport  in order to detect OSDs? if there is nothing else right now i guess we could use that but it just seems like quite a fragile hack.

dosaboy: this function is also quite large so breaking it up a bit would be nice

hotsos/core/host_helpers/systemd.py
```diff
@@ -352,6 +352,12 @@ def _service_filtered_ps(self):
...
+            # No cgroup data at all e.g. a sosreport from a snap-based
+            # deployment. Fall back to scanning ps directly.
+            log.debug("%s doesn't exist. Falling back using ps.", path)
+            return self._ps_fallback_for_services()
```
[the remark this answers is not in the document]
pponnuvel: Done

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -332,6 +354,76 @@ def local_osds(self):
...
+            match = osd_re.search(os.path.basename(status_path))
+            if not match:
+                continue
+
+            osd_id = int(match.group(1))
+            fsid = CephChecks._parse_microceph_osd_fsid(status_path, osd_id)
+            dev = CephChecks._parse_microceph_osd_device(status_path, osd_id)
+            osds.append(CephOSD(osd_id, fsid, dev))
+
+        return osds
```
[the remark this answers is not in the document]
pponnuvel: done

hotsos/core/plugins/storage/ceph/common.py
pponnuvel: Yes, var/snap/microceph is place for all ceph related data (for microceph). The ceph plugin in sosreport collects all the necessary data for microceph.

I've refactored this function.

hotsos/core/plugins/storage/ceph/common.py
pponnuvel: Yes. All the microceph related data are all available in /var/snap/microceph, so all ceph checks should work. It's possible that other system/host related checks may not work but the alternative is to just return False and not run this plugin at all. So I think this doesn't make anything worse.

hotsos/core/host_helpers/systemd.py
```diff
@@ -352,6 +352,12 @@ def _service_filtered_ps(self):
             path = os.path.join(HotSOSConfig.data_root,
                                 'sys/fs/cgroup/system.slice')
 
+        if not os.path.exists(path):
+            # No cgroup data at all e.g. a sosreport from a snap-based
+            # deployment. Fall back to scanning ps directly.
+            log.debug("%s doesn't exist. Falling back to using ps.", path)
```
dosaboy: replace "%s doesn't exist. Falling back to using ps." with:

"could not find cgroup system.slice on this host so falling back to looking up service '%s' processes in ps"

dosaboy (commented):
so one concern I have about supporting this is that hotsos by design does not run scenarios, events etc if core information does not exist e.g. apt packages and snaps and so if we try to support a corner case like a sosreport with microceph info and nothing else then that might lead to unexpected results. There is a --force option to tell hotsos to try everything regardless but that also tends to give unexpected results. Would it not be reasonable to just expect a sosreport to contain snap info?

hotsos/core/host_helpers/systemd.py
pponnuvel: done.

dosaboy (approved):

PR #1091, opened 2026-06-22
[ceph] OSD imbalance across racks/AZs

### LINKED ISSUE #960: Warn on node/OSDs imbalance across AZs
racks/AZs may not necessarily be phyiscal racks. But we can at least flag if they're not balanced.

pponnuvel:
Similar to crushmap_bucket_checks check but this checks if the OSD counts (in the same dev class) are nearly equal. This will flag if OSD counts are uneven if weights happen to match across AZs. This is important to find OSD disk size variability and identify blast radius of OSD failures.

e.g. 4 OSDs of 3TiB vs. 8 OSDs of 1.5TiB, which would pass crushmap_bucket_checks check but will be flagged by this one.

Fixes #960.

Fixes #SET-2195.


hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -195,6 +205,143 @@ def crushmap_equal_buckets_pretty(self):
...
+                continue
+            if max(nonzero) < (
+                    CEPH_OSD_COUNT_IMBALANCE_THRESHOLD * min(nonzero)):
+                continue
+
+            parts = [f"{buckets[bid]['name']}="
+                     f"{per_bucket[bid].get(cls, 0)}"
+                     for bid in fd_bucket_ids
+                     if per_bucket[bid].get(cls, 0) > 0]
+            results.append((cls, parts))
```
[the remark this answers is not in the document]
pponnuvel: Fixed

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -195,6 +205,143 @@ def crushmap_equal_buckets_pretty(self):
...
+
+        count = 0
+        for item in bucket["items"]:
+            if item["id"] >= 0:
+                # Leaf node (OSD)
+                count += 1
+            elif item["id"] in buckets:
+                count += self._count_osds_in_bucket(buckets, item["id"])
+
+        return count
```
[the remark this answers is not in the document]
pponnuvel: Removed

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -195,6 +205,143 @@ def crushmap_equal_buckets_pretty(self):
...
+                    counts[cls] = counts.get(cls, 0) + cnt
+
+        return counts
+
+    def _check_class_imbalance(self, buckets, fd_bucket_ids,
+                               device_classes):
+        """Check for per-class OSD count imbalance across fd buckets.
+
+        @return: list of imbalance description strings
+        """
```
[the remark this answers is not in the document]
pponnuvel: Fixed

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -18,6 +18,11 @@
 
 CEPH_POOL_TYPE = {1: 'replicated', 3: 'erasure-coded'}
 
+# Ratio threshold for OSD count imbalance across failure-domain buckets.
+# A value of 1.33 means a warning is raised when the largest bucket has
+# >= 1.33x the OSDs of the smallest bucket.
+CEPH_OSD_COUNT_IMBALANCE_THRESHOLD = 1.33
```
dosaboy: Where does this number come from and how is it calculated? Can you provide a reference so that if it changes we can update the code?

pponnuvel: I've updated the comment.

This is an arbitrary value, not based on any doc, mainly to avoid triggering on minor imbalances.

pponnuvel:
Also refactored a bit as `hotsos/core/plugins/storage/ceph/cluster.py` is getting too big and produces `C0302: Too many lines in module (1052/1000) (too-many-lines)`

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -198,6 +211,131 @@ def crushmap_equal_buckets_pretty(self):
 
         return None
 
+    def _count_osds_by_class_in_bucket(self, buckets, bucket_id,
```
hillpd: This is inconsistent with _is_bucket_imbalanced(), which ignores 0 weight buckets. Both should either include/ignore buckets under the same conditions. I would favor ignoring 0 weighted buckets because it more accurately reflects the real state of the cluster.

hotsos/core/plugins/storage/ceph/cluster.py
hillpd: The tests cover a 8v4 imbalance, but this  ratio would flag a single-OSD difference whenever the smaller bucket has 3 or fewer OSDs (2v1, 3v2, and 4v3). I recommend pairing this with a floor: (max - min >= 2).

hillpd (changes requested):
Overall this is a good set of checks, but as written it will flag clusters that are perfectly fine.

CRUSH imbalance matters when it creates unusable capacity. These warnings are accurate when the number of failure domains equals the replica count (e.g. 3 AZs with 3-way replication): every AZ holds one replica of everything, so the cluster fills to the smallest AZ and anything above that is unusable. When AZs > replicas, that problem mostly goes away. CRUSH places proportionally to weight, and capacity is only stranded once a single AZ exceeds sum(weights)/replicas. The imbalance has to be far more extreme before anything is lost.

In practice our deployments always use balanced buckets, so this works well as a handover check: any imbalance it flags is a deviation from our reference architecture and worth investigating. For other supported deployments, where uneven failure domains can be intentional and harmless, the check will need be improved in the future.

hotsos/core/plugins/storage/ceph/cluster.py
pponnuvel: done

hotsos/core/plugins/storage/ceph/cluster.py
pponnuvel: done

pponnuvel:
> Overall this is a good set of checks, but as written it will flag clusters that are perfectly fine.
> 
> CRUSH imbalance matters when it creates unusable capacity. These warnings are accurate when the number of failure domains equals the replica count (e.g. 3 AZs with 3-way replication): every AZ holds one replica of everything, so the cluster fills to the smallest AZ and anything above that is unusable. When AZs > replicas, that problem mostly goes away. CRUSH places proportionally to weight, and capacity is only stranded once a single AZ exceeds sum(weights)/replicas. The imbalance has to be far more extreme before anything is lost.
> 
> In practice our deployments always use balanced buckets, so this works well as a handover check: any imbalance it flags is a deviation from our reference architecture and worth investigating. For other supported deployments, where uneven failure domains can be intentional and harmless, the check will need be improved in the future.

This is now addressed too. Skipping the crush tree when AZs (or whatever the failure domain) is greater than the size.

dosaboy (approved):

PR #1093, opened 2026-06-23
[ceph] warn on mixed size OSDs

### LINKED ISSUE #214: Warn on mixed size OSD disks
Mixed size OSDs often cause problems:
http://lists.ceph.com/pipermail/ceph-users-ceph.com/2018-April/026010.html

While this is relatively infrequent in practice, it can happen (e.g. when replacing broken OSDs).

*If* it's necessary to have mixed size OSDs (e.g. during transition from lower sized OSDs to higher sized ones), it's necessary to have correct crush weight to account for mixed size OSDs.

pponnuvel:
Fixes #214.

Fixes #SET-2225.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -20,6 +20,10 @@ from hotsos.core.utils import sorted_dict
 
 log = logging.getLogger()
 
+# Percentage by which OSD sizes within a device class may differ
+# before we consider them "mixed".
+OSD_MIXED_SIZE_THRESHOLD_PERCENT = 10
+
 
 class CephCluster():  # pylint: disable=too-many-public-methods
     """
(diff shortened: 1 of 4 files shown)


hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -838,3 +838,54 @@ def osds_missing_device_class(self):
                 _bad_osds.append(osd['name'])
 
         return sorted(_bad_osds)
+
+    # Percentage by which OSD sizes within a device class may differ
+    # before we consider them "mixed".
+    OSD_MIXED_SIZE_THRESHOLD_PERCENT = 10
```
dosaboy: this needs to live at the top of the file.

pponnuvel: ack, done

hotsos/defs/scenarios/storage/ceph/ceph-mon/osd_mixed_size_disks.yaml
```diff
@@ -0,0 +1,21 @@
...
+        Mixed size OSD disks detected within the same device class:
+        {affected_classes}. Mixed size OSDs can cause data distribution
+        imbalance if the CRUSH weights are not set proportionally to
+        disk size. Verify that crush weights reflect the actual disk sizes
+        using 'ceph osd crush reweight-all' or by manually setting weights
+        with 'ceph osd crush reweight <osd> <weight>'. If mixed sizes are
+        intentional (e.g. during a migration), ensure CRUSH weights are
+        correct to avoid overloading smaller OSDs.
+      format-dict:
+        affected_classes: hotsos.core.plugins.storage.ceph.CephCluster.mixed_size_osd_device_classes_str
```
dosaboy: you should be able to use :comma_join to do this rather than have python code to do it. can you try?

dosaboy (changes requested):

hotsos/defs/scenarios/storage/ceph/ceph-mon/osd_mixed_size_disks.yaml
pponnuvel: ack, done

dosaboy (approved):

Issue #291, opened 2022-03-03
Incorporate CheckPGs script?


pponnuvel:
O7K helm project has a script: https://opendev.org/openstack/openstack-helm-infra/src/branch/master/ceph-mon/templates/bin/utils/_checkPGs.py.tpl

If licensing is OK (MIT?), we can incorporate in hotsos.

pponnuvel:
https://github.com/beskar-cloud/openstack-helm-infra/blob/main/ceph-mon/templates/bin/utils/_checkPGs.py.tpl is the current location.

pponnuvel:
This is addressed by https://github.com/canonical/hotsos/pull/1094

pponnuvel:
#1094 fixed this.

PR #1094, opened 2026-06-24
Detect manual pg-upmap items that break the crush failure domain

taodd:
## Problem

Operators can run commands such as `ceph osd pg-upmap 3.326 6 18 74 100` or `ceph osd pg-upmap-items ...` to fully override a PG's placement. Because these overrides bypass the crush rule, they can place two or more OSDs of a PG into the **same failure domain** (e.g. two OSDs on the same host when the rule requires host-level separation). This silently defeats the redundancy the crush rule is meant to provide and can lead to data loss if that domain fails.

## Change

Add a `CephCluster.manual_upmaps_disrespecting_failure_domain` property that, for every PG carrying a manual upmap (`pg_upmap` full override or `pg_upmap_items` pairwise):

1. Resolves the PG's pool → crush rule → failure domain type (the `type` of the rule's `choose`/`chooseleaf` step) and root tree (the `take` step).
2. Maps each OSD to its failure-domain bucket by walking only the rule's subtree, so shadow trees (e.g. `default~ssd`) that share OSDs do not interfere.
3. Reads the PG's actual placement — the override `osds` list from `ceph osd dump` for full overrides, and the authoritative resulting `up` set from `ceph pg dump` for pairwise items.
4. Flags the PG if any failure-domain bucket holds more than one of its OSDs.

A new `ceph-mon` scenario raises a `CephCrushWarning` listing the offending PGs and buckets, with guidance to remove the mappings via `ceph osd rm-pg-upmap[-items]`.

## Tests

- Unit tests covering a full-override violation, a `pg-upmap-items` violation (via the resulting `up` set), a valid placement across distinct hosts, and the no-upmaps case.
- Scenario test stub validating the raised warning.

All new tests pass; `flake8`, `yamllint` and `pylint` (10.00/10) are clean on the changed files.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -705,3 +705,163 @@ class CephCluster():  # pylint: disable=too-many-public-methods
                     f"'{device_class}' ({min_tb} TiB - {max_tb} TiB)")
 
         return results
+
+    @staticmethod
+    def _rule_root_and_failure_domain(rule):
+        """Extract the (root_item_id, failure_domain_type) from a crush rule.
+
+        A rule selects a starting bucket with a 'take' step then chooses
+        leaves at a bucket type (the failure domain) with a 'choose' step.
+        Returns None if the rule has no such pair of steps.
+        """
+        root = None
+        for step in rule.get('steps', []):
+            op = step.get('op', '')
+            if op == 'take':
+                root = step.get('item')
+            elif 'choose' in op and 'type' in step and root is not None:
+                return (root, step['type'])
+
+        return None
+
+    @staticmethod
+    def _map_osds_to_failure_domain(root_id, fd_type, buckets_by_id):
+        """Map each OSD under a crush tree to its failure domain bucket.
+
+        Walks the subtree rooted at root_id (the tree selected by a rule's
+        'take' step) returning {osd_id: fd_bucket_id} where fd_bucket_id is
+        the ancestor bucket of type fd_type. Restricting the walk to the
+        rule's root means shadow trees (e.g. 'default~ssd') that share the
+        same OSDs do not interfere.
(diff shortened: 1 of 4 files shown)


pponnuvel:
Hi @taodd 

This looks fine.

But I wonder if an alternative approach would be much simpler...what I am thinking of is to simply check if each PG's OSDs are respect failure domains.

Roughly, the steps would be:
- Precompute osd to failure_domain_bucket map (as done here)
- For each PG in pg_stats, look up pool -> rule -> (root, fd_type)
- Check if any two OSDs in the PG's up set share a failure domain bucket

I think that would do the equivalent check and there would be no need to parse pg_upmap and pg_upmap_items from osd_dump separately. What do you think?

taodd:
@pponnuvel, thanks for the review. Won't parsing pg_stats require iterating over all PGs for all pools? For large clusters with over 100k PGs could be expensive. This patch intends to only cover the manual pg-upmap settings by humans, usually caused by a misoperation.
But I agree this approach may be simplified. I'll mark it as a draft for now.

pponnuvel:
Hi @taodd , 

Incidentally I was looking to fix https://github.com/canonical/hotsos/issues/291. And the only check it has this... (based on https://github.com/beskar-cloud/openstack-helm-infra/blob/main/ceph-mon/templates/bin/utils/_checkPGs.py.tpl).

I proposed #1096. Please take a look.

pponnuvel (approved):
LGTM

pponnuvel:
@taodd There's a merge conflict to solve - we can then proceed with this.

taodd:
> @taodd There's a merge conflict to solve - we can then proceed with this.
@pponnuvel Thanks for taking a look, done.

pponnuvel (approved):
LGTM

dosaboy (approved):

PR #1096, opened 2026-06-27
[ceph] detect PG failure domain violations.

### LINKED ISSUE #291: Incorporate CheckPGs script?

pponnuvel:
Fixes #291.

Fixes #SET-2218.

### DIFF
--- a/hotsos/defs/scenarios/storage/ceph/ceph-mon/pg_failure_domain_violation.yaml
+++ b/hotsos/defs/scenarios/storage/ceph/ceph-mon/pg_failure_domain_violation.yaml
@@ -1,0 +1,20 @@
+checks:
+  pgs_violate_failure_domain:
+    property:
+      path: hotsos.core.plugins.storage.ceph.CephCluster.pg_failure_domain_violations
+      ops: [[length_hint], [gt, 0]]
+conclusions:
+  pgs-violating-failure-domain:
+    decision: pgs_violate_failure_domain
+    raises:
+      type: CephCrushWarning
+      message: >-
+        Found placement groups with OSDs that violate failure domain
+        separation. This means some PGs have multiple OSDs in the same
+        failure domain bucket (e.g. same host), reducing data redundancy.
+        Affected pools: {pools}. This is typically caused by manual
+        pg_upmap entries that override CRUSH placement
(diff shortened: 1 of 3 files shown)


pponnuvel:
This is an alternative approach to #1094. So only one should merged.

pponnuvel:
I am converting this to draft. I want @taodd to have a look before deciding this vs. #1094. 

Also @hillpd

pponnuvel:
C0302 (too-many-lines) is fixed in https://github.com/canonical/hotsos/pull/1091

taodd:
@pponnuvel Your approach to check all PGs looks good to me. 
As far as I can see, the only case in which a PG's up set would violate the failure domain is due to a manual pg_upmap setting. 
Is there any other reason that would cause PG not to honor the crush rule?

hotsos/defs/scenarios/storage/ceph/ceph-mon/pg_failure_domain_violation.yaml
```diff
@@ -0,0 +1,18 @@
...
+  pgs-violating-failure-domain:
+    decision: pgs_violate_failure_domain
+    raises:
+      type: CephCrushWarning
+      message: >-
+        Found placement groups with OSDs that violate failure domain
+        separation. This means some PGs have multiple OSDs in the same
+        failure domain bucket (e.g. same host), reducing data redundancy.
+        Affected pools: {pools}. Check the CRUSH map and consider running
+        'ceph osd crush reweight-subtree' or adjusting CRUSH rules.
```
taodd: Perhaps here we can list some possible reasons that would cause the PG to violate the failure domain, so that the user can immediately get what they might have done to cause the problematic PG.

pponnuvel:
> @pponnuvel Your approach to check all PGs looks good to me. As far as I can see, the only case in which a PG's up set would violate the failure domain is due to a manual pg_upmap setting. Is there any other reason that would cause PG not to honor the crush rule?

yeah, I can't think anything else other than manual upmap.

After some thoughts, I think #1094 is fine. So closing this.

PR #1097, opened 2026-06-28
[ceph] report high latencies observed

### LINKED ISSUE #1007: Analyse and report abnormal OSD latencies

pponnuvel:
Ceph logs if the latency of ops is >=5s. This is almost always an issue, typically caused by high load on OSDs, network issue, dying disk, etc.

Fixes #1007.

Fixes #SET-2412.

### DIFF
--- a/tests/unit/storage/test_ceph_osd.py
+++ b/tests/unit/storage/test_ceph_osd.py
@@ -263,6 +263,33 @@ class TestCephOSDEvents(StorageCephOSDTestsBase):
             for key, value in result.items():
                 self.assertEqual(actual[key], value)
 
+    @utils.create_data_root(
+        {'var/log/ceph/ceph-osd.0.log': (
+            '2022-02-10T16:20:23.226+0000 7fc33ca06700  0 '
+            'bluestore(/var/lib/ceph/osd/ceph-0) log_latency slow operation '
+            'observed for submit_transact, latency = 6.402924154s\n'
+            '2022-02-10T16:20:23.310+0000 7fc34e229700  0 '
+            'bluestore(/var/lib/ceph/osd/ceph-0) log_latency_fn slow '
+            'operation observed for _txc_committed_kv, latency = '
+            '6.485089964s, txc = 0x55d96303af00\n'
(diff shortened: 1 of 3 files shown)


tests/unit/fake_data_root/openstack/var/log/ceph/ceph-osd.0.log
```diff
@@ -3688,3 +3688,7 @@ Stalls(count): 0 level0_slowdown, 0 level0_slowdown_with_compaction, 0 level0_nu
 
 ** File Read Latency Histogram By Level [default] **
 
+
```
dosaboy: we need to avoid modifying these files. one day they may be replaced with files from a newer sosreport and these changes will be lost. Please put this into the unit test file itself.

tests/unit/storage/test_ceph_osd.py
```diff
@@ -255,7 +255,8 @@ def test_ceph_daemon_log_checker(self, mock_cli):
         # ensure log file contents are within allowed timeframe ("since")
         mock_cli.return_value.date.return_value = "2021-01-01 00:00:00"
         result = {'crc-err-bluestore': {'2021-02-12': 5, '2021-02-13': 1},
```
dosaboy, line 257: following on from prev comment. to do this will need a new test uing the data provided in  test data root (utils.create_data_root()).

dosaboy, line 257: following on from prev comment. to do this will need a new test using the data provided in  test data root (utils.create_data_root()).

tests/unit/fake_data_root/openstack/var/log/ceph/ceph-osd.0.log
pponnuvel: done

tests/unit/storage/test_ceph_osd.py
pponnuvel: done

pponnuvel: done

dosaboy (approved):

PR #1098, opened 2026-06-28
[ceph] detect and use running versions of OSDs

### LINKED ISSUE #283: Improve OSD version checks in bcache yaml
OSD version checking in [bcache yaml checking](https://github.com/canonical/hotsos/blob/main/defs/bugs/storage/bcache.yaml#L22-L24) can be improved with the command:

```
ceph daemon osd.<num> version
```

which is now available in OSD sosreports too.

CC: @dosaboy

pponnuvel:
We currently used apt to check whether the OSDs are affected by lp1936136. Running versions are a more accurate method to check instead of the installed versions.

Fixes #283.

Fixes #SET-2220.

### DIFF
--- a/hotsos/core/host_helpers/cli/catalog.py
+++ b/hotsos/core/host_helpers/cli/catalog.py
@@ -46,6 +46,17 @@ class CommandCatalog(UserDict):
             'apparmor_status':
                 [BinCmd('apparmor_status'),
                  FileCmd('sos_commands/apparmor/apparmor_status')],
+            'ceph_daemon_osd_version':
+                [BinCmd('ceph daemon osd.{osd_id} version',
+                        json_decode=True),
+                 FileCmd('sos_commands/ceph_osd/'
+                         'ceph_daemon_.var.run.ceph.'
+                         'ceph-osd.{osd_id}.asok_version',
+                         json_decode=True),
+                 FileCmd('sos_commands/ceph_osd/'
+                         'ceph_daemon_.var.snap.microceph.current.run.'
+                         'ceph-osd.{osd_id}.asok_version',
+                         json_decode=True)],
             'ceph_daemon_osd_config_show':
                 [BinCmd('ceph daemon osd.{osd_id} config show',
                         json_decode=True),
--- a/hotsos/core/plugins/storage/ceph/common.py
+++ b/hotsos/core/plugins/storage/ceph/common.py
@@ -532,6 +532,74 @@ class CephDaemonConfigShow():
         return getattr(self.cmd, name)
 
 
+class CephDaemonVersion():
+    """ Interface to ceph daemon osd version command. """
+    def __init__(self, osd_id):
+        self.cmd = CephDaemonCommand('ceph_daemon_osd_version',
+                                     osd_id=osd_id)
+
+    @property
+    def version(self):
+        """
+        Parses the version string from ceph daemon osd version output.
+        The output format is:
+        {"version":"ceph version <ver> (<hash>) <release> (<stability>)"}
+        Returns the version part e.g. "15.2.14".
+        """
+        try:
+            raw = getattr(self.cmd, 'version')
+        except AttributeError:
+            return None
+
+        if raw:
+            match = re.match(r'ceph version (\S+)', raw)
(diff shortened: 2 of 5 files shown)


hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -345,6 +345,55 @@ def local_osds_use_bcache(self):
...
+                 given ranges.
+        """
+        for ver in self.local_osd_daemon_versions:
+            if DPKGVersion.is_version_within_ranges(ver, version_ranges):
+                return True
+
+        return False
+
+    @cached_property
+    def osd_daemon_version_in_lp1936136_ranges(self):
```
dosaboy: lets do this in yaml using operators. this definitely should not live in library code.

dosaboy (changes requested):

hotsos/defs/scenarios/storage/ceph/ceph-osd/bugs/lp1936136.yaml
```diff
@@ -1,20 +1,21 @@
 vars:
   cache_available_percent: '@hotsos.core.plugins.storage.bcache.CachesetsInfo.cache_available_percent'
 checks:
-  node_is_ceph_osd_and_has_version:
-    # Get version of osd based on package installed. This is prone to
-    # inaccuracy since the daemon many not have been restarted after
-    # package update.
-    apt:
+  osd_daemon_has_vulnerable_version:
```
dosaboy, line 4: What about if this is running against a sosreport, dont we want that to work as well? Sosreports wont have the binary. If you really want you could use this first then fallback to apt but I think we need to support apt also.

pponnuvel: I've now added the FileCmd that are collected in sosreports for version checks. That should cover both cases.

But also put the apt check in as a fallback - this isn't strictly necessary but doesn't hurt (in rare cases where admin sockets aren't available).

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -463,6 +463,72 @@ def __getattr__(self, name):
...
+        if raw:
+            match = re.match(r'ceph version (\S+)', raw)
+            if match:
+                return match.group(1)
+
+        return None
+
+
+class CephOSDDaemonVersionInterface():
+    """ Handler for the binary check type to verify OSD daemon versions. """
```
dosaboy, line 563: can we please get a unit test for this method?

hotsos/core/host_helpers/cli/catalog.py
```diff
@@ -46,6 +46,17 @@ def __init__(self):
             'apparmor_status':
                 [BinCmd('apparmor_status'),
                  FileCmd('sos_commands/apparmor/apparmor_status')],
+            'ceph_daemon_osd_version':
```
taodd, line 49: For a monitor sosreport, ceph_versions can tell the running version for all OSD processes. 
But I guess this pr is focusing on the OSD node, or you also intend to add the check to the monitor node ?

pponnuvel: Yes, `ceph versions` does give running versions but it doesn't tell which OSDs are on which versions. There's another check in the monitor that does use it and warn on mixed versions. 

This is focused on OSD nodes and replacing an existing check that uses `apt` for version checking.

taodd: Right.
In that case, ceph_report has which OSD on which version, would that a convenient place to look from ceph mon node?

pponnuvel: `ceph report` would be better but that's not available from OSD sosreports.

taodd: Agreed, Mon's check would suit a separate PR

dosaboy:
lgtm but has merge conflicts, please resolve

pponnuvel:
> lgtm but has merge conflicts, please resolve

done

dosaboy (approved):

PR #1099, opened 2026-06-28
[ceph] warn when size == min_size for replicated pools.

### LINKED ISSUE #998: Check for sane `size` and `min_size` in `ceph report` output
We recently spent some time on a cluster where some pools had `size: 3, min_size: 3`. It would be nice if hotsos could check for sane values of `size` and `min_size` in all the pools given by `ceph report`. If not to recommend exactly `size: 3, min_size: 2`, at least `size > min_size`.

pponnuvel:
This is never recommended except in rare situations where as a temporary recovery mechanism.

Fixes #998.

Fixes #SET-2193.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -635,6 +635,27 @@ class CephCluster():  # pylint: disable=too-many-public-methods
                 return True
         return False
 
+    @cached_property
+    def pools_with_size_equal_min_size(self):
+        """
+        Returns a list of pool names where size == min_size. This is a risky
+        configuration because it means the pool cannot tolerate any OSD
+        failures without becoming unavailable.
+        """
+        bad_pools = []
+        ceph_report = self.crush_map.ceph_report
+        if not ceph_report:
+            return bad_pools
+
+        pools = ceph_report.get('osdmap', {}).get('pools', [])
+        for pool in pools:
+            size = pool.get('size', 0)
+            min_size = pool.get('min_size', 0)
+            if 0 < size <= min_size:
+                bad_pools.append(pool['pool_name'])
+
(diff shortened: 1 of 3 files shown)


hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -635,6 +635,39 @@ def cluster_has_non_empty_pools(self):
...
+
+        pools = ceph_report.get('osdmap', {}).get('pools', [])
+        for pool in pools:
+            # Only check replicated pools
+            if CEPH_POOL_TYPE.get(pool.get('type')) != 'replicated':
+                continue
+
+            size = pool.get('size', 0)
+            min_size = pool.get('min_size', 0)
+            if size > 0 and size == min_size:
```
xtrusia: minor: would size <= min_size be better here? although it is unlikely in practice.

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -635,6 +635,39 @@ def cluster_has_non_empty_pools(self):
...
+                bad_pools.append(pool['pool_name'])
+
+        return bad_pools
+
+    @cached_property
+    def pools_with_size_equal_min_size_str(self):
```
xtrusia: seems it is not used anywhere

pponnuvel: removed

hotsos/core/plugins/storage/ceph/cluster.py
pponnuvel: yeah, ceph won't even allow setting min_size > size. But I've switched condition as it wouldn't hurt in any case.

dosaboy (approved):

dosaboy (approved):

PR #1102, opened 2026-06-29
Catch kubectl error when kube config is missing

xtrusia:
When sunbeam is a controller, ~/.kube/config is expected to exist.

If kubectl fails because the config is missing or invalid, the output is empty.

Handle empty/None output safely, log a warning, and raise an issue in the report so an incomplete sosreport is not missed instead of failing silently.


dosaboy (approved):
lgtm thanks seyeong!

PR #1103, opened 2026-06-30
Detect MDS balancer issue

### LINKED ISSUE #1079: MDS balancer stability and performance issues on Quincy
[issue#61378](https://tracker.ceph.com/issues/61378) – MDS balancer stability and performance issues on Quincy
The mds balancer has been found to introduce performance and stability problems, which is why it is disabled by default in reef and squid. However, this change was not backported to quincy (due to it being EoL upstream)

We recommend disabling the balancer by setting:
```
sudo ceph config set mds mds_bal_interval 0
```

This configuration change takes effect immediately and does not require any service restarts.

xtrusia:
Fixes #1079.

Detect Ceph clusters with CephFS configured where the MDS balancer does not appear to be disabled, and raise Ceph tracker bug 61378 with the recommended mitigation.

### DIFF
--- a/hotsos/core/plugins/storage/ceph/cluster.py
+++ b/hotsos/core/plugins/storage/ceph/cluster.py
@@ -297,6 +297,10 @@ class CephCluster():  # pylint: disable=too-many-public-methods
         self.crush_map = CephCrushMap()
 
     @cached_property
+    def ceph_config_dump(self):
+        return CLIHelper().ceph_config_dump_json_decoded() or []
+
+    @cached_property
     def _osd_daemon_config(self):
         """
         Try to get daemon config from any local OSD. Returns a dict of config
--- a/hotsos/core/plugins/storage/ceph/common.py
+++ b/hotsos/core/plugins/storage/ceph/common.py
@@ -345,6 +345,72 @@ class CephChecks(StorageBase):
     def local_osds_devtypes(self):
         return [osd.devtype for osd in self.local_osds]
 
+    @staticmethod
+    def _is_zero_config_value(value):
+        try:
+            return int(value) == 0
+        except (TypeError, ValueError):
+            return False
+
+    @cached_property
+    def _mds_bal_interval_is_zero(self):
+        """
+        Returns True if mds_bal_interval is explicitly set to 0, which
+        disables the balancer on all releases.
+        """
+        mds_values = []
+        global_values = []
+        for item in self.cluster.ceph_config_dump:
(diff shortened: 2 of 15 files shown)


hotsos/defs/scenarios/storage/ceph/ceph-mon/mds_balancer_quincy.yaml
```diff
@@ -0,0 +1,33 @@
+vars:
+  msg_main: >-
+    This Ceph Quincy cluster has CephFS configured and the MDS balancer does
```
pponnuvel: This isn't specific to Quincy, so I think we just need to make it generic. Also, it's controlled by `balance_automate` in newer versions.

pponnuvel (approved):
LGTM

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -366,6 +370,15 @@ def pg_dump(self):
     def _ceph_mgr_module_ls(self):
         return CLIHelper().ceph_mgr_module_ls() or {}
 
+    @cached_property
+    def filesystems(self):
+        report = CLIHelper().ceph_report_json_decoded() or {}
+        return report.get('fsmap', {}).get('filesystems') or []
+
+    @cached_property
```
dosaboy: this isnt needed. when you call filesystems from a scenario check it will be treated as bool() by default.

hotsos/core/plugins/storage/ceph/cluster.py
```diff
@@ -366,6 +370,15 @@ def pg_dump(self):
     def _ceph_mgr_module_ls(self):
         return CLIHelper().ceph_mgr_module_ls() or {}
 
+    @cached_property
+    def filesystems(self):
```
dosaboy: All of this could be done directly in yaml using operators. That would also make it clearer what it is you are accessing.

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -345,6 +345,72 @@ def local_osds_use_bcache(self):
...
 
+    @staticmethod
+    def _is_zero_config_value(value):
+        try:
+            return int(value) == 0
+        except (TypeError, ValueError):
+            return False
+
+    @cached_property
+    def _mds_bal_interval_is_zero(self):
```
dosaboy, line 356: please add unit tests for these methods

dosaboy (approved):
lgtm

PR #1109, opened 2026-08-04
Fix missing docstrings in a couple of methods

wilkmar:
After merging #1100 there are a couple of functions that don't pass pylint. Added missing docstrings to make pylint pass for the entire project.

PR #1110, opened 2026-08-07
[SET-2465] Scaffold COS CLI package and test integration

wilkmar:
Depends on #1109 

Add a new cos package scaffold with module entry points (__main__, cli.main), basic argparse wiring, and placeholder CLI output.

Include initial COS docs and unit tests for CLI success and help behavior, and wire the package into project packaging, script entry points, coverage settings, and tox test paths.


wilkmar:
As discussed, this will be moved to a separate project

PR #1111, opened 2026-08-07
[SET-2593] Refactor yaml file discover/traversing logic

wilkmar:
Depends on #1109 

This commit moves the YAML file discover/traversing logic to YDefsLoader. It removes the duplicated logic, adds support for optional per-caller file filtering and adds public helper methods for use in other modules.

hotsos/core/ycheck/engine/common.py:
  - Introduce SCENARIOS_SUBDIR and TESTS_SUBDIR constants together with get_defs_dir() / get_defs_tests_dir() helpers that resolve HotSOSConfig.plugin_yaml_defs on every call, so the returned paths always reflect the current config (including values set during test setup).
  - Add YDefsLoader._walk(path): a generic recursive file walker that yields every regular file under path with no filtering.
  - Add YDefsLoader.find_files_recursively(path, file_filter=None): public helper that walks a directory and yields only *.yaml files. An optional callable(abs_path) -> bool file_filter is applied on top to further narrow the result.
  - Add YDefsLoader._default_scenario_file_filter that keeps only *.yaml files. This is what makes discovery yaml-only for both scenario definitions and scenario test definitions: non-yaml files, including editor/disabled artefacts (*.swp / *.disabled) and extensionless files, are all excluded since they do not end in .yaml, so no explicit *.swp / *.disabled handling is needed.
  - Add YDefsLoader.get_scenario_files(subpath=None, file_filter=None) and YDefsLoader.get_scenario_test_files( subpath=None, file_filter=None). Both accept an optional subpath, defaulting to <defs>/scenarios and <defs>/tests respectively (get_scenario_test_files' subpath is relative to <defs>/tests to preserve the existing 'scenarios/<plugin>' calling convention). Both apply the yaml-only filter and any caller-provided file_filter on top.

tests/unit/utils.py:
  - Remove the local find_all_templated_tests() helper.
  - load_templated_tests() now iterates YDefsLoader.find_files_recursively() rooted at os.path.join(DEFS_TESTS_DIR, path). No local filter is needed since discovery is yaml-only.

tests/unit/test_ut_scenario_loader.py:
  - test_find_all_templated_tests now calls YDefsLoader.find_files_recursively(dtmp) directly.
  - create_tests() and the test_check_test_names / test_find_all_templated_tests expectations no longer include extensionless test-definition fixtures, matching the yaml-only discovery.

tools/validation/hotyvalidate.py:
  - Drop the manual glob/os.listdir walking of the scenario and scenario-test trees and delegate discovery to YDefsLoader.get_scenario_files() and YDefsLoader.get_scenario_test_files('scenarios').
  - Recover the plugin sub-root (e.g. 'kernel') from each yielded absolute test path via os.path.relpath, and pass the correct 'scenarios/<subdir>' value as TemplatedTestGenerator's test_defs_root argument.

@dosaboy bringing this explicitly to your attention:
This commit removes the historical support for the extensionless test-def files. Currently, no extensionless test-def files exist in the hotsos/defs/tests tree, so runtime impact today is nil — but the feature/contract (and its tests) is broken.

### DIFF
--- a/hotsos/core/ycheck/engine/common.py
+++ b/hotsos/core/ycheck/engine/common.py
@@ -5,18 +5,57 @@ import yaml
 from hotsos.core.config import HotSOSConfig
 from hotsos.core.log import log
 
+SCENARIOS_SUBDIR = 'scenarios'
+TESTS_SUBDIR = 'tests'
+
+
+def get_defs_dir():
+    """
+    Return the currently configured yaml defs root as an absolute,
+    normalised path (or None if the option has not been set yet).
+
+    Resolved on every call so that changes to HotSOSConfig made after this
+    module has been imported (e.g. during test setup) are honoured.
+    """
+    defs = HotSOSConfig.plugin_yaml_defs
+    if not defs:
+        return None
+    return os.path.abspath(os.path.normpath(defs))
+
+
+def get_defs_tests_dir():
+    """
+    Return <defs>/tests as an absolute path (or None if defs is not set).
+    """
+    defs = get_defs_dir()
+    return os.path.join(defs, TESTS_SUBDIR) if defs else None
+
 
 class YDefsLoader():
-    """ Load yaml definitions. """
+    """ Load yaml definitions with support for file discovery and
+    filtering.
+    """
 
     def __init__(self, ytype, filter_path=None):
         """
         @param ytype: the type of defs we are loading i.e. defs/<ytype>
+        @param filter_path: optional path filter for YAML content
         """
         self.ytype = ytype
         self._loaded_defs = None
         self.stats_num_files_loaded = 0
         self.filter_path = filter_path
+
+    @staticmethod
+    def _walk(path):
+        """Yield absolute paths of every file under path recursively.
+
+        No filtering is applied by the walker itself; callers are
+        responsible for filtering yielded paths as needed.
+        """
+        for root, _dirs, files in os.walk(path):
+            for name in files:
+                yield os.path.join(root, name)
 
     @staticmethod
     def _is_def(abs_path):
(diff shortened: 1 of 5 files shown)


hotsos/core/ycheck/engine/common.py
```diff
@@ -5,19 +5,58 @@
...
+SCENARIOS_SUBDIR = 'scenarios'
+TESTS_SUBDIR = 'tests'
+
+
+def get_defs_dir():
+    """
+    Return the currently configured yaml defs root as an absolute,
+    normalised path (or None if the option has not been set yet).
+
+    Resolved on every call so that changes to HotSOSConfig made after this
```
dosaboy, line 17: I was initially not clear in which situation a config reload might be needed since we currently load config when the client is run/lib is imported and it isnt imported from file so I could not think of a situation when it would change but then I remembered that this can happen in unit tests so I assume that is what you are targeting here.

wilkmar: Yes, this is targeting unit tests specifically.`

tools/validation/hotyvalidate.py
```diff
@@ -26,38 +25,43 @@ def _discover_tests():
         # with the scenario.
         test_scenario_mappings = {}
 
-        # Iterate over all subdirectories of `hotsos/defs/tests` and try to
-        # discover all the available test cases.
-        for subdir in os.listdir(tests_root_path):
```
dosaboy, line 31: there is something not right here. previously the output was:

INFO:root:processed [193] scenarios and [354] tests, all OK!

now it is:

INFO:root:processed [0] scenarios and [0] tests, all OK!

tests/unit/test_ut_scenario_loader.py
```diff
@@ -144,15 +142,12 @@ def test_find_all_templated_tests(self):
...
-                '1/myscenario_noextension1',
                 '1/2/myscenario2.yaml',
                 '1/2/myscenario2alt.yaml',
                 '1/2/myscenario.with.many.dots.2.yaml',
-                '1/2/myscenario_noextension2',
                 '1/2/3/myscenario3.yaml',
                 '1/2/3/myscenario3alt.yaml',
-                '1/2/3/myscenario.with.many.dots.3.yaml',
-                '1/2/3/myscenario_noextension3'
+                '1/2/3/myscenario.with.many.dots.3.yaml'
```
dosaboy, line 150: while removing these works, we could leave them in and update the test accordingly to keep coverage for files with no extension.

dosaboy (changes requested):

tools/validation/hotyvalidate.py
wilkmar: done

tests/unit/test_ut_scenario_loader.py
wilkmar: As discussed, this PR removes support for extensionless scenario/test files.

dosaboy (approved):

PR #1115, opened 2026-08-19
Force fork multiprocessing start method for searchkit on Python 3.14

wilkmar:
searchkit's parallel search relies on the 'fork' start method: worker processes inherit global state (e.g. HotSOSConfig) from the parent and are identified as ForkProcess. Python 3.14 changed the default start method on Linux from 'fork' to 'forkserver', which broke this - workers no longer inherited HotSOSConfig so search constraints (e.g. result age limiting) were silently skipped and search results could be corrupted.

This caused the juju functest to report 72 tracebacks instead of 24 for nova-compute/0. Ensure 'fork' is used where available.

Assisted-By: Claude Opus 4.8

### DIFF
--- a/hotsos/core/search.py
+++ b/hotsos/core/search.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import os
 from datetime import timedelta
 from functools import cached_property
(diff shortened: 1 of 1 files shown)


dosaboy:
Thanks for the patch. So a couple of things; hotsos is not yet tested for 3.14 and once that is done we should add CI to test it. The same applies for dependencies like https://github.com/dosaboy/searchkit and https://github.com/dosaboy/propertree and i feel that the fix you have proposed here belongs in searchkit itself.

dosaboy:
discussed with marcin and we are going to close this and have added an item to the roadmap to add 3.14 support

Issue #974, opened 2024-09-26
Ceph: Detect small DB partition sizes


lathiat:
A common fault in Ceph deployments is that the DB devices are incorrectly configured (missed or allocate from the wrong device), or not big enough. The majority of the time these would be picked up by looking for:
- DB partitions which are obviously far too small, e.g. the default 1GB. Ideally we'd report the DB-to-OSD size ratio informationally

(The following tasks were moved to Issue #1122 instead):
- Empty partitions that have not been used
- Empty space on a disk that is not partition
- Volume groups which are not mostly used (basically the same as empty space on a disk)

pponnuvel:
Shouldn't this be handled by the deployer i.e. the Ceph charms and/or the FE team? A hotsos check might be useful but probably not the most effective place for this.

lathiat:
Yes ideally, but in practice it keeps getting missed. So we need to catch it. Both for analysing new deployments but also detecting the issue on old deployments.

lathiat:
It can also happen because the charm will create OSDs with no DB if it can't find any space, so if you add new OSDs, and the old DB devices were full, a customer could silently have this happen. Or that can happen in a field deployment due to a weird issue even if they designed it right.

dosaboy:
@pponnuvel i agree that the charm should be doing this as a first port of call and we should open a bug on the charm to get this done. In the interim, if it is a small enough addition to the checks we could add this to cover the cases where the charm does not yet support it since it has been cropping up repeatedly in deployments and this will help reduce time of analysis by flagging the issue at the start.

lathiat:
> [@lathiat](https://github.com/lathiat) this almost looks like several checks and it might make sense to break it into smaller chunks to make it easier to implement

Good call, I have updated this ticket for DB partition size only, and split out the empty space on disks to https://github.com/canonical/hotsos/issues/1122

PR #1120, opened 2026-08-28
PR 1: Detect undersized BlueStore DB devices

zhhuabj:
Implement the first item of #974 only: warn when a local OSD has a positive BlueStore DB capacity of 5 GiB or less.

Discover local OSDs from ceph-volume lvm list and obtain each OSD's DB capacity from bluefs.db_total_bytes in its admin-socket perf dump. Support the current standard Ceph sosreport perf-dump filename, which avoids relying on ceph report: that cluster-wide command can fail on OSD-only hosts without an admin keyring.

Add unit and scenario coverage for the warning.

Other checks proposed by #974, including BlueStore DB utilisation and spillover detection, are intentionally not included in this change.

SET-2190


hotsos/core/host_helpers/cli/catalog.py
```diff
@@ -74,6 +74,10 @@ def __init__(self):
             'ceph_daemon_osd_perf_dump':
                 [BinCmd('ceph daemon osd.{osd_id} perf dump',
                         json_decode=True),
+                 FileCmd('sos_commands/ceph_osd/'
+                         'ceph_daemon_.var.run.ceph.'
+                         'ceph-osd.{osd_id}.asok_perf_dump',
+                         json_decode=True),
                  # requires sosreport 4.3 or above
```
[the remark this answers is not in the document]
zhhuabj: done

zhhuabj:
hua@y9000p:/bak/work/hotsos$ .tox/py3/bin/python -m hotsos.cli --storage --scenario storage.ceph.ceph-osd.small_bluestore_db --short data/hotsos-small-db-test.tar.xz
INFO: extracting sosreport data/hotsos-small-db-test.tar.xz to /tmp/tmp73zvhep8/sosreport-juju-b45c2d-ceph-22-hotsos-small-db-minimal-2026-08-27-xthswlm
INFO: analysing sosreport /tmp/tmp73zvhep8/sosreport-juju-b45c2d-ceph-22-hotsos-small-db-minimal-2026-08-27-xthswlm
INFO: analysing -
potential-issues:
  storage:
    CephOSDWarnings:
      - Found OSD(s) osd.0 with a BlueStore DB device of 5 GiB or smaller. A small
        DB device can cause metadata to spill over to the main OSD device and degrade
        performance. Review the OSD device layout and provision a larger fast DB device.

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -444,6 +447,29 @@ def local_osds_with_oversized_bluefs_log(self):
 
         return sorted(bad)
 
+    @cached_property
+    def local_osds_with_small_bluestore_db(self):
+        """Return local OSDs with a BlueStore DB device at most 5 GiB."""
```
pponnuvel, line 452: Use `BLUESTORE_DB_SIZE_MIN` instead of 5 GiB

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -179,6 +179,9 @@ class CephInstallInfo(InstallInfoBase):
 
 class CephChecks(StorageBase):
     """ Ceph Checks. """
+    # A 1 GiB BlueStore DB is the ceph-volume default and is generally too
+    # small for production OSDs.
+    BLUESTORE_DB_SIZE_MIN = 5 * 1024 * 1024 * 1024
```
pponnuvel, line 184: wy 5 GiB? The typical recommendation is 4% of the block dev size. So it's possible that a DB dev is > 5GiB but still be "small" for that OSD.

pponnuvel (commented):
Looks good overall but I do want to understand how 5GiB was chosen and whether it's good for different OSD sizes.

hotsos/core/plugins/storage/ceph/common.py
zhhuabj: Hi pon, thanks for the review. I addressed your comments to check the BlueStore DB capacity against 4% of the actual OSD block-device capacity.

The check now:
- reads `bluefs.db_total_bytes` from each local OSD admin-socket perf dump;
- uses the primary ceph-volume `[block]` entry, with `block device` retained as a compatibility fallback;
- maps `lsblk -O -P` device paths to exact `blockdev --report` byte sizes;
- supports standard Ceph LVM mapper aliases and MicroCeph `/dev/disk/by-id/...` paths through `ID-LINK`;
- skips an OSD when required inventory data is unavailable, preserving compatibility with incomplete/older sosreports.

I tested the implementation with real sosreport data from both Charmed Ceph and MicroCeph. The unmodified archives return no result for this scenario, which is expected for their healthy layouts. I then copied each archive, changed only one OSD's `bluefs.db_total_bytes` in the copied perf dump to 100 MiB, and ran hotsos end-to-end - https://pastebin.com/2HtuWpfg

pponnuvel (approved):
Looks good. Thanks @zhhuabj !

One suggestion:

If these commands provide json (or yaml) formatted output, we could use that instead of "grepping". I guess that depends on if those commands support it and whether sosreport collects those formatted output. If that's not available or sosreport doesn't collect now, it's fine to merge this as is.

zhhuabj:
> Looks good. Thanks @zhhuabj !
> 
> One suggestion:
> 
> If these commands provide json (or yaml) formatted output, we could use that instead of "grepping". I guess that depends on if those commands support it and whether sosreport collects those formatted output. If that's not available or sosreport doesn't collect now, it's fine to merge this as is.

@pponnuvel  Pon, thanks for the suggestion. `ceph-volume lvm list` supports
`--format=json`, but sosreport currently only collects the default
pretty output:

https://github.com/sosreport/sos/blob/main/sos/report/plugins/ceph_osd.py#L105

It does not currently collect `ceph-volume lvm list --format=json`.
The sosreport used for validation also only contains the pretty output.

I therefore kept the existing parsing approach for compatibility with
current sosreports. We can switch to structured output if sosreport
starts collecting the JSON variant in the future.

hotsos/core/plugins/storage/ceph/common.py
```diff
@@ -532,6 +541,127 @@ def local_osds_with_oversized_bluefs_log(self):
 
         return sorted(bad)
 
+    @staticmethod
```
dosaboy, line 544: Can we please consider putting lsblk parsing code into a utility class of its own so that it can be easily reused by other code and possibly plugins? For example we could introduce a new hotsos.core.host_helpers.storage module that has a LsblkHelper class with this code in it (sticking to the code that is purely for parsing lsblk and nothing ceph specific)

zhhuabj: Hi Ed @dosaboy  , I've put lsblk parsing code into hotsos/core/host_helpers/storage.py , and here is my latest test result  -  https://pastebin.com/P3MSuzCW

dosaboy: thank you joshua for fixing this

hotsos/core/plugins/storage/ceph/daemon.py
```diff
@@ -125,11 +125,14 @@ def __init__(self):
 
 class CephOSD(CephDaemonBase):
     """ Representation of a Ceph OSD """
-    def __init__(self, ceph_id, fsid=None, device=None, dump=None):
+    # pylint: disable-next=too-many-arguments,too-many-positional-arguments
```
dosaboy: rather than do this we should create a settings object and pass that in instead

dosaboy (approved):
approving but would like to followup with change for commented code

PR #1126, opened 2026-09-09
Fix OVS scenario warning when other_config invalid

brianphaley:
If the ovs-vsctl "other_config" file is invalid and does not contain a dictionary, the openvswitch plugin will show various warnings for scenarios since it
throws an exception under the hood.

For example, if the file contains a string like:

  "ovs-vsctl: unix:/var/snap/openstack-hypervisor/
   common/run/openvswitch/db.sock: database connection
   failed (No such file or directory)"

You will see:

  HotSOSScenariosWarnings:
      - One or more scenarios failed to run (dpdk_config) - run hotsos in debug mode (--debug) to get more detail

Check it is a dict before trying to use it as such.

### DIFF
--- a/hotsos/core/plugins/openvswitch/ovs.py
+++ b/hotsos/core/plugins/openvswitch/ovs.py
@@ -315,7 +315,14 @@ class OVSDPDK(OpenvSwitchBase):
     @cached_property
     def config(self):
         """ Return the OVSDB Open_vSwitch other_config dict. """
-        return self.ovsdb.Open_vSwitch.other_config or {}
+        config = self.ovsdb.Open_vSwitch.other_config
+        if not config:
+            return {}
+
+        if not isinstance(config, dict):
+            return {}
+
+        return config
 
     @property
     def enabled(self):


hotsos/core/plugins/openvswitch/ovs.py
```diff
@@ -315,7 +315,14 @@ class OVSDPDK(OpenvSwitchBase):
     @cached_property
     def config(self):
         """ Return the OVSDB Open_vSwitch other_config dict. """
-        return self.ovsdb.Open_vSwitch.other_config or {}
+        config = self.ovsdb.Open_vSwitch.other_config
+        if not config:
+            return {}
+
+        if not isinstance(config, dict):
```
dosaboy, line 322: since this is not considered expected it would be good to have a log here to say that the return type was not expected dict type.

dosaboy:
I am also hitting this and have found another case of the problem so going to merge and will followup with another patch to address points here also the other issue.

PR #1133, opened 2026-09-15
[ceph] fix false-positive overlapping-roots CRUSH check

### LINKED ISSUE #1132: Ceph: 'overlapping roots' check produces false positives
This check flags an issue whenever a real (non-shadow) CRUSH root contains OSDs from more than one device class. But this is not correct. 

Every cluster that uses CRUSH device classes (e.g. mixed hdd/ssd clusters) will have its real default root span multiple classes by design; the per-class ~ssd/~hdd shadow trees are subsets of it, not separate misconfigurations. 

As a result, the check fires on essentially any normal mixed-device-class cluster, even when ceph health detail reports HEALTH_OK and no pool is actually affected by the real ceph-mgr PG autoscaler "overlapping roots" condition.

pponnuvel:
CephCrushMap.crush_tree_has_overlapping_roots flagged any real root containing OSDs from more than one device class, which is normal for every cluster using CRUSH device classes and produced guaranteed false positives.

Replace it with crushmap_overlapping_roots, which mirrors what the ceph-mgr PG autoscaler actually flags: two or more pools whose CRUSH rules take different but overlapping roots (e.g. an unrestricted root and a device-class-restricted shadow root sharing OSDs). Only in-use rules/roots are considered, and overlap is determined by comparing each root's reachable OSD set rather than its device-class makeup.

Update the autoscaler_overlap_roots scenario and tests accordingly, replacing the false-positive fixture with regression tests for both the non-overlapping (mixed device classes, single rule) and genuinely overlapping (multiple rules, shared OSDs) cases.

Fixes SET-2744.

Fixes #1132.


hotsos/defs/scenarios/storage/ceph/ceph-mgr/autoscaler_overlap_roots.yaml
```diff
@@ -13,18 +13,20 @@ checks:
...
+  overlapping-roots-from-crush-dump:
     priority: 1
-    decision: overlapping_roots_crush_tree
+    decision: overlapping_roots_crush_dump
     raises:
       type: CephMgrError
       message: >-
-        Overlapping CRUSH roots detected via ceph osd crush tree
-        --show-shadow. {common}
+        Overlapping CRUSH roots detected: {roots}. {common}
```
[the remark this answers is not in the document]
pponnuvel: Modified as suggested.

hotsos/defs/scenarios/storage/ceph/ceph-mgr/autoscaler_overlap_roots.yaml
```diff
@@ -13,19 +13,22 @@ checks:
...
-    decision: overlapping_roots_crush_tree
+    decision: overlapping_roots_crush_dump
     raises:
       type: CephMgrError
       message: >-
-        Overlapping CRUSH roots detected via ceph osd crush tree
-        --show-shadow. {common}
+        Overlapping CRUSH roots detected: {roots}. Please inspect the
+        affected pool CRUSH rules and use disjoint roots so the PG
+        autoscaler can scale those pools.
```
dosaboy: It would be great to elaborate here on what steps need to be taken or point to some documenation or upstream info on how to resolve the problem.

hillpd (approved):
Thanks for this. This is the right direction: comparing the OSDs under each in-use root is exactly what ceph-mgr does, and it gets rid of the false positive on every cluster that uses device classes. Approving as is.

While checking it against ceph-mgr I found another false positive on multi-take crush rules, and that from Tentacle the autoscaler no longer skips pools on overlapping roots, so the check only applies to Pacific through Squid. I also took a stab at making the message more actionable. The suggested changes grew into a 13-file branch, so rather than block this PR I'll open it as a follow-up once this lands. The details are in #1139 and the branch is at https://github.com/pponnuvel/hotsos/compare/set-2744...hillpd:hotsos:pr1133-review

```
[ceph] make overlapping-roots check match mgr and name the fix

Use the first take step of each in-use rule as its root, as the mgr does,
so multi-take rules (ssd-primary) are not flagged against themselves.
Report each overlapping pair with the rules and pools on each side and
the corrective action. Gate on Pacific through Squid; from Tentacle the
autoscaler budgets PGs across shared OSDs instead (https://github.com/ceph/ceph/pull/68532).
Reword the log-based conclusion to match and let the crush dump one take
precedence. Add tentacle to the release tables, drop the unused crush
tree command, and give the crush dump test data full bucket fields.
```

hotsos/defs/scenarios/storage/ceph/ceph-mgr/autoscaler_overlap_roots.yaml
pponnuvel: Elaborated a bit more. Please take a look again.

PR #1134, opened 2026-09-17
[SET-2758] Fix MAAS plugin support for >= 3.5

dosaboy:
MAAS 3.5 moved to using pebble to manage subcomponents and as a results changed how logs are written. We now get logs from systemd journal.


pponnuvel (approved):
LGTM

goldberl:
lgtm as well

goldberl (approved):

PR #1135, opened 2026-09-18
[SET-2764] Add skill to verify scenario search constraints

dosaboy:
Adds a skill that looks at all scenarios, identifies those that have search constraints and ensures that the search expressions contain the required patterns to match dates used by the constraints filter.

Also ensures we have test coverage for the negative version of a scenario i.e. testing what happens when use logs that have old timestamps. These tests all have an old_logs suffix.


wilkmar (commented):
Do we need to involve LLM to identify incorrect search patterns? If implemented in code, would be 100% deterministic. This is very important, because this check should now be part of tox testing, or PR automated review process to detect these type of problems for every new PR (if the PR adds/modifies check). However once you have done the cleanup for the entire repository, now the PR check should rather test each new PR/check rather than testing the entire repo over and over again. This should be implemented as a separate PR.

dosaboy:
Your question is the right one to ask.

My experience with this particular skill is that given the complexity of what we need to check (and the fact that what we are checking could evolve over time e.g. if we add support for more log formats) a skill seems the most effective way to do it. The part of writing both positive and negative tests for each scenario seems like a good fit for a skill.

We could potentially encapsulate the check that this skill performs into code and perhaps that would be a good way to enable this kind of check in the CI. We could use this skill to periodically check that script is up-to-date and has the necessary coverage.

wilkmar:
> Your question is the right one to ask.
> 
> My experience with this particular skill is that given the complexity of what we need to check (and the fact that what we are checking could evolve over time e.g. if we add support for more log formats) a skill seems the most effective way to do it. The part of writing both positive and negative tests for each scenario seems like a good fit for a skill.
> 
> We could potentially encapsulate the check that this skill performs into code and perhaps that would be a good way to enable this kind of check in the CI. We could use this skill to periodically check that script is up-to-date and has the necessary coverage.

Sounds good to me. 
In my previous comment I forgot to mention that you should consider using the evals.json for your skill as a means of documenting the prompt you use to execute the skill.

wilkmar (approved):
LGTM

PR #1136, opened 2026-09-18
[SET-2762] Add checks for openstack name resolution errors

dosaboy:
When common openstack services have name resolution errors in their logs we now raise an OpenstackWarning.


brianphaley (approved):

hotsos/defs/scenarios/openstack/dns_resolution_failures.yaml
```diff
@@ -0,0 +1,191 @@
...
+      format-dict:
+        names: '@checks.has_dns_failures.search.results_group_3:unique_comma_join'
+  sunbeam_keystone_has_dns_failures:
+    decision:
+      - is_sunbeam_controller
+      - has_sunbeam_keystone_dns_failures
+    raises:
+      type: OpenstackWarning
+      message: >-
+        The Sunbeam keystone pod on this host has name resolution failures
```
wilkmar: this and the following pod/services have mostly the same conclusion message. I believe it would be easy to make it a variable and avoid duplicating the text for each service

wilkmar (changes requested):
Technically any service can be affected. This patch only adds checks for keystone, neutron, and nova.

